PR #17, opened 2019-08-18
Create Sigma sysdeps

thomtl:
Also add the possibility to have sysdeps link in extra libs


abis/sigma/abi.h
```diff
@@ -0,0 +1,23 @@
+
```
avdgrinten: Please check if managarm/qword ABIs are identical here. If yes, move the file to abis/mlibc and change managarm's and qword's symlinks.

sysdeps/sigma/generic/generic.cpp
```diff
@@ -0,0 +1,110 @@
...
+#include <libsigma/thread.h>
+
+
+#define STUB_ONLY { __ensure(!"STUB_ONLY function was called"); __builtin_unreachable(); }
+
+namespace mlibc {
+    // Unknown Functions
+    int sys_tcb_set(void *pointer){
+	    libsigma_set_fsbase(reinterpret_cast<uint64_t>(pointer));
+        return 0;
```
avdgrinten: Indentation inconsistent. In general, mlibc uses tabs.

sysdeps/sigma/generic/generic.cpp
```diff
@@ -0,0 +1,110 @@
...
+    int sys_getrusage(int who, struct rusage *usage) STUB_ONLY
+    int sys_sigprocmask(int how, const sigset_t *__restrict set, sigset_t *__restrict retrieve) STUB_ONLY
+    void sys_yield() STUB_ONLY
+    gid_t sys_getgid() STUB_ONLY
+    gid_t sys_getegid() STUB_ONLY
+    uid_t sys_getuid() STUB_ONLY
+    uid_t sys_geteuid() STUB_ONLY
+    pid_t sys_getpid() STUB_ONLY
+    pid_t sys_getppid() STUB_ONLY
+    #endif
```
avdgrinten: Note that implementing all sysdeps functions will disable some graceful fallbacks in mlibc. For example, may functions will log an error and fail with ENOSYS if the sysdeps function is unavailable. This makes mlibc panic instead.

sysdeps/sigma/generic/log.cpp
```diff
@@ -0,0 +1,23 @@
...
+#include <assert.h>
+
+#include <bits/ensure.h>
+#include <mlibc/debug.hpp>
+#include <mlibc/sysdeps.hpp>
+
+#include <libsigma/klog.h>
+#include <libsigma/thread.h>
+
+#define UNIMPLEMENTED() { __ensure(!"Unimplemented function was called"); __builtin_unreachable(); }
```
avdgrinten: This is unused.

sysdeps/sigma/generic/memory.cpp
```diff
@@ -0,0 +1,25 @@
...
+#include <errno.h>
+
+#define STUB_ONLY { __ensure(!"STUB_ONLY function was called"); __builtin_unreachable(); }
+
+namespace mlibc
+{
+    int sys_anon_allocate(size_t size, void **pointer){
+        __ensure((size & 0xFFF) == 0);
+        uint64_t ret = libsigma_valloc(LIBSIGMA_VALLOC_TYPE_SBRK_LIKE, 0, size / 0x1000);
+        if(ret == 1) return ENOMEM;
```
avdgrinten: This coding style is inconsistent with mlibc. Put the "then case" onto the next line.

sysdeps/sigma/meson.build
```diff
@@ -0,0 +1,40 @@
+sigma_incl = include_directories('../../subprojects/libsigma/include')
+#		'../../subprojects/managarm/hel/include')
```
avdgrinten: Copy & paste mistake.

sysdeps/sigma/meson.build
```diff
@@ -0,0 +1,40 @@
+sigma_incl = include_directories('../../subprojects/libsigma/include')
```
avdgrinten: Is this needed? After all, libsigma is already added as a dependency.

sysdeps/sigma/rtdl-generic/support.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+        private:
+        uint64_t id;
+    };
+
+    auto file_descriptor_allocator = id_allocator();
+
+    struct file_descriptor {
+        file_descriptor(): fd(0), offset(0), name(frg::string<MemoryAllocator>(getAllocator())) {}
+        file_descriptor(const char* name): fd(file_descriptor_allocator.get_id()), offset(0), name(frg::string<MemoryAllocator>(getAllocator(), name) + '\0') {}
```
avdgrinten: This coding style is inconsistent with mlibc. Put `: <initialization list>` onto the next line, at the same level of indentation as the constructor declaration.

sysdeps/sigma/rtdl-generic/support.cpp
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    return EBADF;
+}
+int sys_read(int fd, void *data, size_t length, ssize_t *bytes_read){
+    for(auto& file : get_file_descriptor_list()){
+        if(file.fd == fd){
+            if(libsigma_read_initrd_file(file.name.data(), (uint8_t*)data, file.offset, length)){
+                *bytes_read = 0;
+                return 1;
```
avdgrinten: Is this EOF or an error? EOF should return success but set bytes_read to zero. Errors need to return error codes.

sysdeps/sigma/rtdl-generic/support.cpp
```diff
@@ -0,0 +1,105 @@
...
+}
+int sys_close(int fd){
+    return 0;
+}
+
+int sys_vm_map(void *hint, size_t size, int prot, int flags, int fd, off_t offset, void **window){
+    if(!(flags & MAP_ANONYMOUS)) __ensure(!"Anything else than MAP_ANONYMOUS is unimplemented");
+
+    int ret = libsigma_vm_map(size, hint, prot, flags);
+    *window = hint;
```
avdgrinten: Note that hint is really only a hint unless MAP_FIXED is set (e.g., it might be NULL). You may want to assert that MAP_FIXED is set.

sysdeps/sigma/generic/generic.cpp
thomtl: Do all function do this?

qookei: Not all iirc, some assume the sysdep function is present (`mkstemp` assumes `sys_open` exists, `mkdtemp` assumes `sys_mkdir` exists, for example)

thomtl: Is there a list somewhere, because going through all calls to all sysdep functions seems very tedious.

avdgrinten: Not all functions have graceful fallbacks but all functions should either panic or fallback to an error return. If some functions use `[[gnu::weak]]` sysdeps without checking for them, that's a bug.

abis/sigma/abi.h
thomtl: Yep they are thesame

sysdeps/sigma/include/abi-bits/abi.h
```diff
@@ -0,0 +1 @@
+../../../../abis/sigma/abi.h
```
avdgrinten: This still links abis/sigma instead of abis/mlibc, even though both files are identical.

sysdeps/sigma/rtdl-generic/support.cpp
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    return EBADF;
+}
+int sys_close(int fd){
+    return 0;
+}
+
+int sys_vm_map(void *hint, size_t size, int prot, int flags, int fd, off_t offset, void **window){
+    if(!(flags & (MAP_ANONYMOUS | MAP_FIXED)))
```
avdgrinten: This check is likely not what you want. It checks that whether none of the bits are set. Do you want to check for `!(flags & MAP_ANONYMOUS) || !(flags & MAP_FIXED))`?

sysdeps/sigma/meson.build
avdgrinten: I think this should be solved differently (both in the managarm sysdeps and here), as the current solution requires libsigma as a subproject but also requires it to be installed. We can merge it for now and change this later on.

sysdeps/sigma/rtdl-generic/support.cpp
thomtl: No, It should fail if MAP_ANON or MAP_PRIVATE are not set so i think its correct

sysdeps/sigma/include/abi-bits/abi.h
thomtl: my bad

sysdeps/sigma/rtdl-generic/support.cpp
avdgrinten: Your check does not trigger the panic if e.g., MAP_FIXED is set but MAP_ANONYMOUS isn't.

thomtl: Oh yeah, i missed that

PR #22, opened 2019-12-19
sysdeps/managarm: Implement flock

Matt8898:


subprojects/managarm
```diff
@@ -0,0 +1 @@
+/home/matt/mnt/managarm/build/../src/managarm
```
avdgrinten: This should not be added.

PR #25, opened 2020-02-13
Added Lemon compatibility and 'static' build option

fido2020:
I added headers and sysdeps for Lemon OS, I also added option to build a static library.


meson.build
```diff
@@ -126,11 +131,17 @@ internal_sources = [
 	'options/internal/gcc/guard-abi.cpp',
 	'options/internal/gcc/initfini.cpp',
 	'options/internal/gcc-extra/cxxabi.cpp',
-	'options/internal/gcc-extra/mlibc_crtbegin.S',
-	'options/internal/gcc-extra/mlibc_crtend.S',
```
avdgrinten: Why does lemon not use these?

sysdeps/lemon/generic/lemon.cpp
```diff
@@ -0,0 +1,50 @@
...
+	}
+} 
+
+extern "C" int __cxa_atexit(){
+	return 0;
+}
+
+void __mlibc_do_finalize(){
+
+}
```
avdgrinten: Why are those functions missing and not provided by mlibc (outside of sysdeps)?

sysdeps/lemon/include/lemon/filesystem.h
```diff
@@ -0,0 +1,36 @@
...
+	char name[128]; // Filename
+    uint32_t type;
+} lemon_dirent_t;
+
+int lemon_open(const char* filename, int flags);
+void lemon_close(int fd);
+int lemon_read(int fd, void* buffer, size_t count);
+int lemon_write(int fd, const void* buffer, size_t count);
+off_t lemon_seek(int fd, off_t offset, int whence);
+int lemon_readdir(int fd, uint64_t count, lemon_dirent_t* dirent);
```
avdgrinten: I think it would make sense to extract these functions into some `liblemon` (for example). You could link such a library into mlibc at compile time. This is what the `sigma` target also does. Let me know if you have any questions about such an approach.

fido2020: I moved a bunch of code to my liblemon library, however I kept those filesystem functions in the libc as they are used by `sys_open`, `sys_read`, etc.

sysdeps/lemon/generic/lemon.cpp
fido2020: I don't have support for dynamic linking, and lsb uses __dlapi_get_tls

sysdeps/lemon/include/lemon/filesystem.h
qookei: You probably want to inline the code from the `lemon_{read,write,open,close}` etc into lemon's `sys_xxx` functions directly, since you're just passing the arguments directly anyway.

meson.build
```diff
@@ -126,11 +132,16 @@ internal_sources = [
...
 	'options/internal/x86_64/setjmp.S',
 ]
 
+if host_machine.system() != 'lemon'
+	internal_sources += [
+		'options/internal/gcc-extra/mlibc_crtbegin.S',
+		'options/internal/gcc-extra/mlibc_crtend.S',
```
avdgrinten: Do not test for `lemon` here but for dynamic linking.

meson.build
```diff
@@ -169,30 +180,58 @@ subdir('options/glibc')
...
+		static_library('c',
+				[
+					libc_sources,
+					internal_sources,
+					ansi_sources,
+				],
+				cpp_args: ['-DFRIGG_HAVE_LIBC'],
+				include_directories: libc_include_dirs,
+				dependencies: libc_deps,
+				link_with: [ldso_lib],
```
avdgrinten: The static library still links against the dynamic linker. Is that intended?

sysdeps/lemon/generic/filesystem.cpp
```diff
@@ -0,0 +1,79 @@
+#include <lemon/syscall.h>
+#include <sys/types.h>
+
+namespace mlibc{
+
+	int sys_write(int fd, const void* buffer, size_t count, ssize_t* written){
+
+
```
avdgrinten: Minor: strange whitespace.

sysdeps/lemon/generic/filesystem.cpp
```diff
@@ -0,0 +1,79 @@
...
+		return 0;
+	}
+
+	int sys_read(int fd, void *buf, size_t count, ssize_t *bytes_read) {
+		int ret;
+		int sys_errno;
+
+    	syscall(SYS_READ, fd, (uintptr_t)buf, count, (uintptr_t)&ret, 0);
+
+		if (ret == -1)
```
avdgrinten: Minor: indentation is broken.

abis/lemon/auxv.h
```diff
@@ -0,0 +1,9 @@
+#ifndef _ABIBITS_AUXV_H
+#define _ABIBITS_AUXV_H
+
+#define AT_PHDR 3
+#define AT_PHENT 4
+#define AT_PHNUM 5
+#define AT_ENTRY 9
```
avdgrinten: These definitions match the Linux ABI. You could symlink the Linux ABI's `auxv.h` instead.

sysdeps/lemon/generic/filesystem.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+		if(ret) {
+		    sys_close(ret);
+		    return 0;
+		} else return 1;
+	}
+
+	void sys_libc_log(const char*){
+
+	}
```
avdgrinten: Consider implementing this to get debugging information.

sysdeps/lemon/crt-x86_64/crt0.S
```diff
@@ -0,0 +1,23 @@
+.section .text
+
+.global _start
+_start:
+    movq $0, %rbp
+	pushq %rbp
+	pushq %rbp
```
avdgrinten: Why do you push `%rbp` twice? (Indentation is broken here.)

sysdeps/lemon/generic/lemon.cpp
```diff
@@ -0,0 +1,50 @@
...
+namespace mlibc{
+
+	void sys_exit(int status){
+		syscall(SYS_EXIT, status, 0, 0, 0, 0);
+	}
+
+	int sys_anon_allocate(size_t size, void **pointer) {
+		// Make sure to only allocate whole pages
+		__ensure(!(size & 0xFFF));
+		syscall(SYS_ALLOC, (((size / 0x1000)*0x1000 < size) ? (size / 0x1000) + 1 : (size / 0x1000)), (uintptr_t)pointer, 0, 0, 0);
```
avdgrinten: Rounding up to page size can be simplified: `(size + 0xFFF) & ~static_cast<size_t>(0xFFF)`. This is also faster because it doesn't generate a `div` instruction.

sysdeps/lemon/generic/lemon.cpp
```diff
@@ -0,0 +1,50 @@
...
+		    return -1;
+		return 0;
+	}
+
+	int sys_anon_free(void *pointer, size_t size) {
+		return 0; // Not implemented
+	}
+
+	void sys_libc_panic(){
+		*((int*)0xffffDEADDEADDEAD) = 1; // Force Page Fault
```
avdgrinten: You could use `__builtin_trap()` instead (forces an illegal instruction / `ud2`).

sysdeps/lemon/include/lemon/spawn.h
```diff
@@ -0,0 +1,18 @@
...
+
+#include <stddef.h>
+#include <sys/types.h>
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C"{
+#endif
+
+void lemon_spawn(const char* path);
```
avdgrinten: This header is not inlined yet.

sysdeps/lemon/generic/filesystem.cpp
avdgrinten: (See also the other code below.)

sysdeps/lemon/generic/lemon.cpp
avdgrinten: This needs to be fixed in the long term. Maybe add a comment that this should be removed once LSB is also compiled into the static lib.

fido2020: will do

abis/lemon/auxv.h
fido2020: Probably better off being separate in case I need to add anything to the header in the future

avdgrinten:
Alright, from my side, this can be merged. Do you also consider this PR to be done?

fido2020:
Yep I see nothing further to be done

PR #26, opened 2020-02-13
Changes to compile wget and curl

Geertiebear:
More to follow obviously, maybe check if I didn't mess up the abi-bits stuff


options/ansi/generic/string-stubs.cpp
```diff
@@ -174,8 +174,17 @@ char *strstr(const char *s, const char *pattern) {
 	return nullptr;
 }
 char *strtok(char *__restrict s, const char *__restrict delimiter) {
```
avdgrinten: Can we Implement strtok() based on strtok_r()?

options/ansi/generic/file-io.cpp
```diff
@@ -331,6 +335,23 @@ int abstract_file::_write_back() {
 	return 0;
 }
 
+int abstract_file::_discard_buffer() {
+	if (_type == stream_type::file_like) {
+		// save our current location
+		off_t new_offset;
+		auto seek_offset = (off_t(__offset) - off_t(__io_offset));
+		if (int e = io_seek(seek_offset, SEEK_CUR, &new_offset); e) {
```
avdgrinten: Why do we need to seek in the underlying FD?

options/ansi/generic/file-io.cpp
```diff
@@ -237,9 +237,13 @@ void abstract_file::purge() {
 	__dirty_end = __dirty_begin;
 }
 
-// TODO: For input files, discard the buffer.
 int abstract_file::flush() {
-	return _write_back();
+	if (__dirty_end == __dirty_begin) {
+		return _discard_buffer;
```
avdgrinten: Why don't we discard the buffer unconditionally?

options/ansi/generic/file-io.cpp
```diff
@@ -331,6 +335,23 @@ int abstract_file::_write_back() {
 	return 0;
 }
 
+int abstract_file::_discard_buffer() {
+	if (_type == stream_type::file_like) {
```
avdgrinten: Is _type guaranteed to be initialized here?

abis/linux/x86_64/signal.h
```diff
@@ -0,0 +1,108 @@
+#ifndef _ABIBITS_SIGNAL_H
+#define _ABIBITS_SIGNAL_H
+
+#ifdef __MLIBC_POSIX_OPTION
+#include <bits/posix/siginfo_t.h>
```
avdgrinten: siginfo_t is also a OS-specific ABI (?).

avdgrinten:
One further comment: for single line `if`s, please put the then-statement on its own line.

options/ansi/generic/file-io.cpp
Geertiebear: Because as far as I saw there was now way to save the current position in the file properly, without rewriting some of the other code. I could probably rewrite it to avoid seeking but I was reluctant to do that.

options/ansi/generic/file-io.cpp
Geertiebear: Spec states that we should only discard the buffer for input streams, and output streams should write back to the FD

options/ansi/generic/file-io.cpp
Geertiebear: no, good point

abis/linux/x86_64/signal.h
Geertiebear: siginfo is needed for sigval

options/ansi/generic/file-io.cpp
avdgrinten: As discussed over Discord, this seek makes sense. Maybe add a comment that we seek to reset the position of the FD to the position of the FILE.

options/ansi/generic/file-io.cpp
avdgrinten: I think here we should really make sure that we do the correct thing for r/w files. Can we maybe write a test case to validate that our behavior matches glibc/musl?

PR #28, opened 2020-02-17
ci: add files needed to build with jenkins

ArsenArsen:
adding tests would be trivial: adding another stage inside Docker that runs the tests


avdgrinten:
LGTM. Thank you!

PR #34, opened 2020-04-07
options/linux: add sys/reboot.h

ArsenArsen:


options/linux/generic/sys-reboot.cpp
```diff
@@ -0,0 +1,10 @@
+#include <errno.h>
+#include <sys/reboot.h>
+#include <mlibc/sysdeps.hpp>
+
+int reboot(int what) {
+	if (!mlibc::sys_reboot) {
+		return ENOSYS;
```
avdgrinten, line 7: This needs to call `MLIBC_MISSING_SYSDEP();` (grep for it and compare with other stubs. Also, is the error handling (errno vs. return value) correct?

options/linux/include/utmp.h
```diff
@@ -2,6 +2,8 @@
 #ifndef  _UTMP_H
 #define  _UTMP_H
 
+#include <stdint.h>
```
avdgrinten, line 5: Why do we need this?

options/posix/include/sys/types.h
```diff
@@ -22,6 +22,7 @@
 
 #include <bits/posix/fsblkcnt_t.h>
 #include <bits/posix/fsfilcnt_t.h>
+#include <sys/select.h>
```
avdgrinten, line 25: This looks strange. Which types do we actually want? Those types should be extracted into a `bits/` header (see the other includes above this one); that header should be included here.

sysdeps/qword/meson.build
```diff
@@ -21,6 +21,7 @@ if not no_headers
 		'include/abi-bits/in.h',
 		'include/abi-bits/stat.h',
 		'include/abi-bits/signal.h',
+		'include/abi-bits/reboot.h',
```
avdgrinten, line 24: qword doesn't have this header.

sysdeps/lemon/meson.build
```diff
@@ -19,6 +19,7 @@ if not no_headers
 		'include/abi-bits/in.h',
 		'include/abi-bits/signal.h',
 		'include/abi-bits/stat.h',
+		'include/abi-bits/reboot.h',
```
avdgrinten, line 22: lemon doesn't have this header.

options/linux/include/utmp.h
avdgrinten: We should rather replace the uses by `__mlibc_uint32` (etc.) from `bits/types.h`.

avdgrinten:

PR #35, opened 2020-04-07
Added required functions for GNU Tar

Dennisbonke:
GNU Tar wants this function when making tar files.
However, it also fails on openat, getpwnam and maybe others which I haven't found yet.
I tried looking into that but I am not quite sure what to do with those functions, so I leave those up to the more experienced.
According to creat(2) the creat call is nothing more then a call to open with these 3 flags set.


Dennisbonke:
I have added the getpwnam and openat functions, openat works good enough for tar. It seems we now need fchown, although as far as i know file permissions are not implemented yet

Dennisbonke:
I deem tar sufficiently working. You can create an archive and extract it, however overwriting files is broken on unlinkat. Other than that I found no issues in the quick testing that I did

avdgrinten:
Great. Thank for the PR!

PR #39, opened 2020-04-13
posix: fdopendir should respect flags that are already set

Dennisbonke:


avdgrinten:
LGTM. Thank you for the PR!

PR #63, opened 2020-04-20
Various fixes for man-db

Dennisbonke:
This PR implements the following:

- `towlower()` and `towupper()` char converters.
-  the flushing of all files when `fflush(NULL)` is called and on `exit()`

### DIFF
--- a/options/ansi/generic/file-io.cpp
+++ b/options/ansi/generic/file-io.cpp
@@ -578,6 +578,15 @@ long ftell(FILE *file_base) {
 }
 
 int fflush_unlocked(FILE *file_base) {
+	if(file_base == NULL) {
+		// Only flush the files but do not close them.
+    	for(auto it : mlibc::global_file_list) {
+			if(int e = it->flush(); e)
+				mlibc::infoLogger() << "mlibc warning: Failed to flush file"
+						<< frg::endlog;
+		}
+		return 0;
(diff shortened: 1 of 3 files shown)


options/ansi/generic/stdlib-stubs.cpp
```diff
@@ -205,6 +206,7 @@ int at_quick_exit(void (*func)(void)) {
 }
 
 void exit(int status) {
+	fflush_unlocked(NULL);
```
avdgrinten, line 209: We do not need to call this here (that is already done by a global destructor).

options/ansi/generic/file-io.cpp
```diff
@@ -578,6 +578,15 @@ long ftell(FILE *file_base) {
 }
 
 int fflush_unlocked(FILE *file_base) {
+	if(file_base == NULL) {
+		// Only flush the files but do not close them.
+    	for(auto it : mlibc::global_file_list) {
+			if(int e = it->flush(); e)
+				mlibc::infoLogger() << "mlibc warning: Failed to flush file before exit()"
+						<< frg::endlog;
```
avdgrinten, line 586: This message is confusing (we are not necessarily exiting).

avdgrinten:
LGTM now, thank you.

PR #64, opened 2020-04-22
Implementing various functions and a bugfix

Dennisbonke:
This PR adds the following:

- An implementation of the `setbuf()` function,
- An implementation of the `pread()` function,
- And a small bug fix to `strtok()`, related to strings which end on the requested delimiter.


avdgrinten:
LGTM. Thank you for the PR!

PR #71, opened 2020-04-26
various net-related fixes

ArsenArsen:


options/posix/generic/arpa-inet-stubs.cpp
```diff
@@ -65,9 +65,13 @@ in_addr_t inet_addr(const char *) {
 	__ensure(!"Not implemented");
 	__builtin_unreachable();
 }
-char *inet_ntoa(struct in_addr) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+char *inet_ntoa(struct in_addr addr) {
+	thread_local static char buffer[16];
```
avdgrinten, line 69: Is this buffer big enough? 4*4+1 = 17.

options/posix/generic/arpa-inet-stubs.cpp
```diff
@@ -65,9 +65,13 @@ in_addr_t inet_addr(const char *) {
...
+	int proper = htonl(addr.s_addr);
```
avdgrinten, line 70: This should be a `uint32_t`.

options/posix/generic/arpa-inet-stubs.cpp
ArsenArsen: four groups of up to three numbers, three periods, and a null: `4 * 3 + 3 + 1 = 16`

avdgrinten: Add a comment (it's not so hard to see but also not totally trivial).

avdgrinten:
LGTM. Blocked on the corresponding Managarm changes.

PR #73, opened 2020-04-30
Various fixes to port python

Dennisbonke:
These fixes also allow `gettext` to compile.
Python is functioning in script form, the interpreter doesn't respond to input, but the basic functionality is there

### DIFF
--- a/options/ansi/include/setjmp.h
+++ b/options/ansi/include/setjmp.h
@@ -3,6 +3,7 @@
 #define _SETJMP_H
 
 #include <bits/machine.h>
+#include <abi-bits/signal.h>
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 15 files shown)


options/internal/x86_64/setjmp.S
```diff
@@ -31,3 +31,13 @@ longjmp:
 	mov 0x30(%rdi), %rsp
 	jmp *0x38(%rdi)
 
+.global sigsetjmp
+.type sigsetjmp, "function"
+sigsetjmp:
+	ret
```
avdgrinten, line 37: Here we either need a proper implementation, or we need to warn that it is not implemented.

options/internal/x86_64/setjmp.S
```diff
@@ -31,3 +31,13 @@ longjmp:
...
+
+.global siglongjmp
+.type siglongjmp, "function"
+siglongjmp:
+	ret
```
avdgrinten, line 42: Same.

options/ansi/include/setjmp.h
```diff
@@ -13,8 +14,16 @@ extern "C" {
 // FIXME: use intptr_t equivalent
 typedef long jmp_buf[__MLIBC_JMPBUF_SIZE];
 
+typedef struct sigjmp_buf {
+	jmp_buf __jmpbuf;
+	int __mask_was_saved;
+	sigset_t __saved_mask;
+} sigjmp_buf;
```
avdgrinten, line 21: The standard defines `jmp_buf` to be an array. Is `sigjump_buf` supposed to be a struct?

avdgrinten: POSIX says:
> The <setjmp.h> header shall define the array types jmp_buf and sigjmp_buf.

PR #82, opened 2020-05-10
managarm: Implement various uid/gid functions

Dennisbonke:
This PR aims to implement get/set(e)uid and friends for managarm.
It also holds a small bugfix in `strerror` where EPERM was listed as EFAULT


Dennisbonke:
I think this part of users/groups/permissions in managarm is done, so should be good for review

PR #84, opened 2020-05-13
Various functions to allow xbps and perl to compile

Dennisbonke:
This adds the following functions and definitions

For `xbps`:
- `getentropy()`, stubbed
- `vsyslog()`, stubbed

For `perl`:
- `h_errno`
- `gethostbyaddr()`, stubbed
- `drand48()`, stubbed
- `srand48()`, stubbed
- `struct ipc_perm`
- `struct shmid_ds`

### DIFF
--- a/options/posix/generic/netdb-stubs.cpp
+++ b/options/posix/generic/netdb-stubs.cpp
@@ -7,6 +7,14 @@
 #include <stdlib.h>
 #include <stddef.h>
 
+#undef h_errno
+__thread int __mlibc_h_errno;
+
+// This function is from musl
+int *__h_errno_location(void) {
+	return &__mlibc_h_errno;
+}
+
 void endhostent(void) {
 	__ensure(!"Not implemented");
 	__builtin_unreachable();
@@ -120,6 +128,11 @@ struct netent *getnetent(void) {
 
 struct hostent *gethostbyname(const char *) {
 	__ensure(!"gethostbyname() not implemented");
+	__builtin_unreachable();
+}
+
+struct hostent *gethostbyaddr(const void *, socklen_t, int) {
+	__ensure(!"gethostbyaddr() not implemented");
 	__builtin_unreachable();
 }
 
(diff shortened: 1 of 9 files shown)


options/posix/generic/netdb-stubs.cpp
```diff
@@ -7,6 +7,13 @@
 #include <stdlib.h>
 #include <stddef.h>
 
+#undef h_errno
+int h_errno;
+
+int *__h_errno_location(void) {
+	return &h_errno;
+}
```
avdgrinten, line 15: Add a comment that this function is from glibc, it looks *really* strage (but the LSB standard contains it, so we should keep it).

options/posix/generic/netdb-stubs.cpp
```diff
@@ -7,6 +7,13 @@
 #include <stdlib.h>
 #include <stddef.h>
 
+#undef h_errno
+int h_errno;
```
avdgrinten, line 11: Shouldn't this be `__thread_local` like the normal errno?

avdgrinten: It should probably be called `__mlibc_h_errno` since it's a "anonymous" global variable.

options/posix/include/sys/ipc.h
```diff
@@ -14,4 +18,16 @@
...
+struct ipc_perm {
+	key_t __ipc_perm_key;
+	uid_t uid;
+	gid_t gid;
+	uid_t cuid;
+	gid_t cgid;
+	mode_t mode;
+	int __ipc_perm_seq;
+	long __pad1;
+	long __pad2;
```
avdgrinten, line 30: I think we can drop the padding if Linux doesn't use it. Are the double underscore fields used in Linux?

options/posix/include/sys/shm.h
```diff
@@ -46,6 +46,19 @@ extern "C" {
...
+	struct ipc_perm shm_perm;
+	size_t shm_segsz;
+	time_t shm_atime;
+	time_t shm_dtime;
+	time_t shm_ctime;
+	pid_t shm_cpid;
+	pid_t shm_lpid;
+	unsigned long shm_nattch;
+	unsigned long __pad1;
+	unsigned long __pad2;
```
avdgrinten, line 59: Dito.

avdgrinten:
Thank you for the PR!

PR #86, opened 2020-05-17
Various fixes for socat

Dennisbonke:
This PR aims to implement the required functionality for `socat` to compile.
So far, only the build time errors have bee resolved, the link time errors are still there.

### DIFF
--- a/abis/linux/x86_64/termios.h
+++ b/abis/linux/x86_64/termios.h
@@ -1,6 +1,10 @@
 #ifndef _ABIBITS_TERMIOS_H
 #define _ABIBITS_TERMIOS_H
 
+typedef unsigned int cc_t;
+typedef unsigned int speed_t;
+typedef unsigned int tcflag_t;
+
 // indices for the c_cc array in struct termios
 #define NCCS     32
 #define VINTR    0
--- a/abis/linux/x86_64/termios.h
+++ b/abis/linux/x86_64/termios.h
@@ -1,0 +1,35 @@
+#ifndef _ABIBITS_TERMIOS_H
+#define _ABIBITS_TERMIOS_H
+
+// indices for the c_cc array in struct termios
+#define NCCS     32
+#define VINTR    0
+#define VQUIT    1
+#define VERASE   2
+#define VKILL    3
+#define VEOF     4
+#define VTIME    5
+#define VMIN     6
+#define VSWTC    7
+#define VSTART   8
+#define VSTOP    9
+#define VSUSP    10
+#define VEOL     11
+#define VREPRINT 12
+#define VDISCARD 13
+#define VWERASE  14
+#define VLNEXT   15
+#define VEOL2    16
+
+struct termios {
+	tcflag_t c_iflag;
+	tcflag_t c_oflag;
+	tcflag_t c_cflag;
+	tcflag_t c_lflag;
+	cc_t c_line;
+	cc_t c_cc[NCCS];
+	speed_t ibaud;
+	speed_t obaud;
+};
+
+#endif
--- a/abis/mlibc/termios.h
+++ b/abis/mlibc/termios.h
@@ -1,0 +1,30 @@
+#ifndef _ABIBITS_TERMIOS_H
+#define _ABIBITS_TERMIOS_H
+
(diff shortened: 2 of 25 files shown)


Dennisbonke:
The link errors have been fixed, `socat` compiles fine. For reasons unrelated to this PR, it doesn't run yet.

options/posix/include/sys/param.h
```diff
@@ -3,6 +3,11 @@
 #define _SYS_PARAM_H
 
 #include <endian.h>
+#include <limits.h>
+
+#if !defined NGROUPS && defined NGROUPS_MAX
+# define NGROUPS NGROUPS_MAX
+#endif
```
avdgrinten, line 10: This is quite ugly. Whether the value is defined or not depends on the order of the headers?

options/posix/include/termios.h
```diff
@@ -25,6 +25,8 @@ typedef unsigned int tcflag_t;
 #define VSTOP    8
 #define VSUSP    9
 #define VTIME   10
+#define VLNEXT  15
+#define VEOL2   16
```
avdgrinten, line 29: This would need an adjustment of NCCS which is an ABI break (because it changes the size of `struct termios`). It is okay to break managarm's ABI if we need this but we should move these definitions and the struct into an `abi-bits/` header and keep the old definitions for all other ports.

options/glibc/include/resolv.h
```diff
@@ -12,6 +21,16 @@ int dn_expand(const unsigned char *, const unsigned char *,
...
 
+int res_init(void);
+
+/* From musl: Unused; purely for broken apps
+ * To avoid an massive struct, only add the items requested. */
+typedef struct __res_state {
+	unsigned long options;
+} *res_state;
+struct __res_state *__res_state(void);
+#define _res (*__res_state())
```
avdgrinten, line 32: Why is this needed?

options/posix/generic/syslog-stubs.cpp
```diff
@@ -1,15 +1,47 @@
...
 #include <bits/ensure.h>
 
+static char log_ident[32];
+static int log_options;
+static int log_facility = LOG_USER;
+static int log_fd = -1;
+
+static const struct {
+	short sun_family;
+	char sun_path[9];
```
avdgrinten, line 15: Why don't we use a proper `struct sockaddr_un`?

options/posix/include/sys/param.h
Dennisbonke: Yeah I don’t like it either, but this is what Linux does. I think we can unconditionally define them

options/glibc/include/resolv.h
Dennisbonke: This is what musl does. As it states, this entire thing is purely for broken applications. It shouldn’t be used in new applications

options/posix/include/sys/param.h
avdgrinten: I guess that would be preferable.

abis/linux/x86_64/termios.h
```diff
@@ -0,0 +1,35 @@
...
+#define VSWTC    7
+#define VSTART   8
+#define VSTOP    9
+#define VSUSP    10
+#define VEOL     11
+#define VREPRINT 12
+#define VDISCARD 13
+#define VWERASE  14
+#define VLNEXT   15
+#define VEOL2    16
```
avdgrinten, line 22: These are all the V... constants defined for linux?

abis/mlibc/termios.h
```diff
@@ -0,0 +1,30 @@
...
+#define VINTR    3
+#define VKILL    4
+#define VMIN     5
+#define VQUIT    6
+#define VSTART   7
+#define VSTOP    8
+#define VSUSP    9
+#define VTIME   10
+#define VLNEXT  15
+#define VEOL2   16
```
avdgrinten, line 18: The old ABI should not declare as programs detecting them could cause buffer overflows (by `c_cc[VLNEXT]` etc.).

avdgrinten:
There are some CI failures (missing includes in the abi headers).

avdgrinten:

Issue #27, opened 2020-02-13
mlibc ABI has overlapping signal.h constants


Geertiebear:
For example, ``SIGIO`` and ``SIGBUS`` have the same value

PR #100, opened 2020-05-31
Various additions for tcl, gdb and perl

Dennisbonke:
This PR will hold the necessary changes to mlibc for `gdb` and `tcl` to build.
For now, these are:

- `pthread_attr_getscope()`, stubbed in, required by `tcl`
- `pthread_attr_setscope()`, stubbed in, required by `tcl`
- Defined `FIONBIO` for `gdb`
- Fixed overlapping signal constants (issue #27)
- Updated `sysconf()` with hardcoded values for `_SC_GETPW_R_SIZE_MAX` and `_SC_GETGR_R_SIZE_MAX`
- Moved the declarations of `htonl()` and friends to `netinet/in.h`, where they reside on Linux

Furthermore, some general improvements have been made:
- Implemented `utime()`, `utimes()`, `futimens()` and `utimensat()`
- Implemented `remove()`
- Defined `UTIME_NOW` and `UTIME_OMIT`

This PR is blocked on inclusion of managarm/managarm#181

### DIFF
--- a/options/ansi/generic/time-stubs.cpp
+++ b/options/ansi/generic/time-stubs.cpp
@@ -334,9 +334,31 @@ int clock_settime(clockid_t, const struct timespec *) {
 	__builtin_unreachable();
 }
 
-int utimes(const char *, const struct timeval[2]) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+int utimes(const char *filename, const struct timeval times[2]) {
+	if (!mlibc::sys_utimensat) {
+		MLIBC_MISSING_SYSDEP();
+		errno = ENOSYS;
+		return -1;
+	}
+	struct timespec time[2];
+	if(times == nullptr) {
+		time[0].tv_sec = UTIME_NOW;
+		time[0].tv_nsec = UTIME_NOW;
+		time[1].tv_sec = UTIME_NOW;
+		time[1].tv_nsec = UTIME_NOW;
+	} else {
+		time[0].tv_sec = times[0].tv_sec;
+		time[0].tv_nsec = times[0].tv_usec * 1000;
+		time[1].tv_sec = times[1].tv_sec;
+		time[1].tv_nsec = times[1].tv_usec * 1000;
+	}
+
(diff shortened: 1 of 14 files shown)


Dennisbonke:
commit 8f3e4f9 fixes #27

options/ansi/generic/time-stubs.cpp
```diff
@@ -334,9 +334,25 @@ int clock_settime(clockid_t, const struct timespec *) {
...
-int utimes(const char *, const struct timeval[2]) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+int utimes(const char *filename, const struct timeval[2]) {
+	if (!mlibc::sys_utimensat) {
+		MLIBC_MISSING_SYSDEP();
+		errno = ENOSYS;
+		return -1;
+	}
+	// TODO: We really should respect timeval here
```
Geertiebear, line 343: We can just check if ``times != NULL`` here and convert it to a timespec right? Seems simple enough and it's a good idea to include it else we might produce unexpected results silently

options/glibc/include/sys/ioctl.h
```diff
@@ -15,6 +15,7 @@ extern "C" {
 int ioctl(int fd, unsigned long request, ...);
 
 #define FIONREAD 0x541B
+#define FIONBIO	 0x5421
```
Geertiebear, line 18: minor: style issue with the tab here

options/posix/include/bits/posix/stat.h
```diff
@@ -3,6 +3,10 @@
 
 #include <abi-bits/stat.h>
 
+// Used by utimensat and friends
+#define UTIME_NOW      ((1l << 30) - 1l)
```
Geertiebear, line 7: minor: the tab again

options/posix/include/arpa/inet.h
```diff
@@ -9,11 +9,6 @@
 extern "C" {
 #endif
 
-uint32_t htonl(uint32_t);
```
Geertiebear, line 12: we shouldn't remove these from this header, musl has these functions defined in both so I think we should match that behaviour.

PR #106, opened 2020-06-07
Small fix in sys_utimensat

Dennisbonke:
This fix allows sys_utimensat to compile when managarm/managarm#181 is merged and adds support for separate atime and mtime specifications.

As such, it is blocked on managarm/managarm#181


avdgrinten:
Nice, thank you!

PR #107, opened 2020-06-09
ABI: Fix overlapping signal constants and struct sockaddr

### LINKED ISSUE #27: mlibc ABI has overlapping signal.h constants

Dennisbonke:
This ABI breaking change is needed for `tcl` to compile. It also fixes #27 
This PR also includes an ABI breaking change regarding the `sockaddr` struct, we now match glibc's behavior regarding sa_data's array size. This satisfies `python` enough to build the `_socket` module

### DIFF
--- a/abis/mlibc/signal.h
+++ b/abis/mlibc/signal.h
@@ -34,42 +34,41 @@ extern "C" {
 #define SIG_DFL ((__sighandler)(void *)(-2))
 #define SIG_IGN ((__sighandler)(void *)(-3))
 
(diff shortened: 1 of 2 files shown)


avdgrinten:
Signal constants should be changed to match Linux before we merge.

Dennisbonke:
They match linux as of a493aa2

avdgrinten:
LGTM, thank you!

PR #115, opened 2020-06-16
Fix regressions in select() and fflush()

### LINKED ISSUE #111: options/internal: SIGSEGV in locking code
While testing `git` on managarm, the following bug came to my attention.
By doing the following operations in order, the program can be killed with an `SIGSEGV`
To reproduce:

- Have `git` installed and a git repo available, in this test case managarm
- Do the required git setup, for managarm, this is `touch /dev/null`, and setting the global username and email (git will complain about this)
- In the git repo, touch a file, and run `git add .`
- Run `git commit -m <message>`, where message is the commit message (irrelevant for triggering this bug) and watch it crash.

Running `addr2line` on `libc.so` gives the following line
```
dennis@DENNIS-MAIN-LAPTOP-SIDUCTION:~/managarm-clone/build$ addr2line --e system-root/usr/lib/libc.so 
0x987d0
/home/dennis/managarm-clone/build/pkg-builds/mlibc/../../../src/mlibc/options/internal/include/mlibc/lock.hpp:24
```

which translates to

```
if(__atomic_compare_exchange_n(&_futex, &expected, 1,
							false, __ATOMIC_ACQUIRE, __ATOMIC_ACQUIRE)) {
```

where the first part of the if statement is line 24 in lock.hpp.

Geertiebear:
Fixes #111 and a ``select()`` regression

### DIFF
--- a/options/ansi/generic/file-io.cpp
+++ b/options/ansi/generic/file-io.cpp
@@ -586,10 +586,10 @@ long ftell(FILE *file_base) {
 int fflush_unlocked(FILE *file_base) {
 	if(file_base == NULL) {
 		// Only flush the files but do not close them.
-    	for(auto it : mlibc::global_file_list) {
+		for(auto it : mlibc::global_file_list) {
 			if(int e = it->flush(); e)
 				mlibc::infoLogger() << "mlibc warning: Failed to flush file"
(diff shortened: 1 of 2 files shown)


avdgrinten:
LGTM.

PR #122, opened 2020-06-25
rtdl static linking fix and debugging support

Geertiebear:


options/rtdl/generic/linker.cpp
```diff
@@ -16,6 +16,8 @@ bool stillSlightlyVerbose = false;
 bool logBaseAddresses = false;
 bool eagerBinding = true;
 
+extern DebugInterface debug;
```
avdgrinten, line 19: Can we give that variable a more specific name? Something like `globalDebugInterface`?

options/rtdl/generic/linker.cpp
```diff
@@ -279,6 +281,11 @@ void ObjectRepository::_fetchFromPhdrs(SharedObject *object, void *phdr_pointer,
 			object->tlsAlignment = phdr->p_align;
 			object->tlsImageSize = phdr->p_filesz;
 			tls_offset = phdr->p_vaddr;
+		case PT_INTERP:
+			object->interpreterPath = frg::string<MemoryAllocator>{
+				(char*)(phdr->p_vaddr),
+					getAllocator()
+			};
```
avdgrinten, line 288: Do we need to respect `->p_filesz` or is this string always null-terminated according to the ELF spec?

options/rtdl/generic/linker.cpp
```diff
@@ -851,6 +860,11 @@ Scope::Scope()
 : _objects(getAllocator()) { }
 
 void Scope::appendObject(SharedObject *object) {
+	if (_objects.size()) {
+		auto prevObject = _objects[_objects.size() - 1];
+		object->linkMap.prev = &prevObject->linkMap;
+		prevObject->linkMap.next = &object->linkMap;
+	}
```
avdgrinten, line 867: Hm, maybe this should be moved to `Loader::submitObject`? Semantically, this happens during loading and the scope is not really involved here since the linkMap is global anyway.

options/rtdl/generic/linker.cpp
```diff
@@ -885,6 +899,12 @@ void Loader::submitObject(SharedObject *object) {
 	if(_linkSet.get(object))
 		return;
 
+	// at this point the object is loaded and we can fill in its debug struct
+	// the linked list fields will be filled later
```
avdgrinten, line 903: Comments should be proper sentences (capitalization + punctuation, also elsewhere).

options/rtdl/generic/linker.hpp
```diff
@@ -80,6 +80,21 @@ enum class HashStyle {
 
 using InitFuncPtr = void (*)();
 
+struct DebugInterface {
+	int ver;
+	void *head;
+	void (*brk)(void);
+	int state;
+	void *base;
+};
```
avdgrinten, line 89: Add a comment that the ABI of this struct is fixed by GDB.

options/rtdl/generic/linker.hpp
```diff
@@ -80,6 +80,21 @@ enum class HashStyle {
...
+
+struct LinkMap {
+	uintptr_t base = 0;
+	const char *name = nullptr;
+	Elf64_Dyn *dynv = nullptr;
+	LinkMap *next = nullptr, *prev = nullptr;
+};
```
avdgrinten, line 96: Dito.

options/rtdl/generic/main.cpp
```diff
@@ -95,10 +95,17 @@ extern "C" [[ gnu::visibility("default") ]] void *__rtdl_allocateTcb() {
...
+DebugInterface debug;
+DebugInterface *_dl_debug_addr = &debug;
+
+extern "C" {
+	static void dl_debug_state() {
+
+	}
 }
 
+extern "C" [[gnu::weak, gnu::alias("dl_debug_state"), gnu::visibility("default")]] void _dl_debug_state();
```
avdgrinten, line 107: Why `gnu::weak`?

options/rtdl/generic/main.cpp
```diff
@@ -95,10 +95,17 @@ extern "C" [[ gnu::visibility("default") ]] void *__rtdl_allocateTcb() {
...
 
-extern "C" [[ gnu::visibility("default") ]] void _dl_debug_state()
-{
+DebugInterface debug;
+DebugInterface *_dl_debug_addr = &debug;
+
+extern "C" {
+	static void dl_debug_state() {
+
+	}
```
avdgrinten, line 104: Add a comment that GDB sets a breakpoint here but that we don't need to do anything.

options/rtdl/generic/main.cpp
```diff
@@ -95,10 +95,17 @@ extern "C" [[ gnu::visibility("default") ]] void *__rtdl_allocateTcb() {
 	return allocateTcb();
 }
 
-extern "C" [[ gnu::visibility("default") ]] void _dl_debug_state()
-{
+DebugInterface debug;
+DebugInterface *_dl_debug_addr = &debug;
```
avdgrinten, line 99: Is one of these fields external ABI?

options/rtdl/generic/main.cpp
```diff
@@ -199,20 +207,24 @@ extern "C" void *interpreterMain(uintptr_t *entry_stack) {
 	// Add the dynamic linker, as well as the exectuable to the repository.
 #ifndef MLIBC_STATIC_BUILD
 	auto ldso_soname = reinterpret_cast<const char *>(ldso_base + strtab_offset + soname_str);
-	initialRepository->injectObjectFromDts(ldso_soname,
+	auto ldso = initialRepository->injectObjectFromDts(ldso_soname,
 		frg::string<MemoryAllocator> { execfn, getAllocator() },
```
avdgrinten, line 211: Can we initialize the path to an empty string? `execfn` is certainly wrong here.

options/rtdl/generic/linker.cpp
Geertiebear: It's null terminated.

options/rtdl/generic/linker.cpp
```diff
@@ -942,8 +957,19 @@ void Loader::linkObjects() {
...
 
-	for(auto it = _linkBfs.begin(); it != _linkBfs.end(); ++it)
+	for(auto it = _linkBfs.begin(); it != _linkBfs.end(); ++it) {
 		(*it)->wasLinked = true;
+
+		if((*it)->inLinkMap)
+			continue;
+
+		auto linkMap = reinterpret_cast<LinkMap*>(globalDebugInterface.head);
+		for (; linkMap->next != nullptr; linkMap = linkMap->next);
```
avdgrinten, line 967: Why is this needed? Can't we just store the tail somewhere?

avdgrinten: Or, alternatively, append at the head?

PR #123, opened 2020-06-26
Small fixes for xorg-server

### LINKED ISSUE #117: getentropy() unimplemented
During testing on managarm, the following program required `getentropy()`

- `wget`, when fetching `https://` links

Dennisbonke:
This PR fixes some small bugs that the compilation of xorg-server on managarm found.
Furthermore, it implements some functions that are required at runtime for `Xwayland`. It also adds `getentropy()`, and the corresponding syscall for managarm.

Blocked on managarm/managarm#195
Closes #117

### DIFF
--- a/sysdeps/managarm/generic/file.cpp
+++ b/sysdeps/managarm/generic/file.cpp
@@ -1406,7 +1406,6 @@ int sys_epoll_ctl(int epfd, int mode, int fd, struct epoll_event *ev) {
 		req.set_flags(ev->events);
 		req.set_cookie(ev->data.u64);
 	}else if(mode == EPOLL_CTL_DEL) {
-		__ensure(!ev);
 		req.set_request_type(managarm::posix::CntReqType::EPOLL_DELETE);
 	}else{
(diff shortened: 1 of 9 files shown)


sysdeps/managarm/generic/file.cpp
```diff
@@ -4088,5 +4089,20 @@ int sys_utimensat(int dirfd, const char *pathname, const struct timespec times[2
...
+	SignalGuard sguard;
+	uint8_t buf[length];
+	size_t size;
+
+	if(length > 256)
+		return EIO;
+
+	HEL_CHECK(helGetRandomBytes(buf, length, &size));
+
+    memcpy(buffer, buf, size);
```
avdgrinten, line 4102: Here we have an issue since `helGetRandomBytes` might return fewer than `length` bytes, while `getentropy` is supposed to return exactly `length` bytes. We also do not need `buf`. Suggestion:

```suggestion
	auto p = reinterpret_cast<char *>(buffer);
	size_t n = 0;
	while(n < length) {
		size_t chunk;
		HEL_CHECK(helGetRandomBytes(p + n, length - n, &chunk));
		n += chunk;
	}
```

PR #129, opened 2020-07-18
Various fixes required for Xwayland support in managarm

Dennisbonke:
As discussed in the Discord, this PR holds all the changes required in mlibc

Blocked on managarm/managarm#201

### DIFF
--- a/options/ansi/generic/signal-stubs.cpp
+++ b/options/ansi/generic/signal-stubs.cpp
@@ -7,22 +7,19 @@
 #include <mlibc/sysdeps.hpp>
 
 __sighandler signal(int sn, __sighandler handler) {
-	mlibc::infoLogger() << "\e[31mmlibc: signal() always returns SIG_DFL\e[39m" << frg::endlog;
 	if(!mlibc::sys_sigaction) {
 		MLIBC_MISSING_SYSDEP();
 		errno = ENOSYS;
 		return SIG_ERR;
 	}
-	if(handler == SIG_DFL || handler == SIG_IGN) {
-	}else{
-		struct sigaction sa;
-		sa.sa_handler = handler;
-		sa.sa_flags = 0;
-		sa.sa_mask = 0;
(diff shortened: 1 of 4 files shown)


options/ansi/generic/signal-stubs.cpp
```diff
@@ -7,13 +7,20 @@
 #include <mlibc/sysdeps.hpp>
 
 __sighandler signal(int sn, __sighandler handler) {
-	mlibc::infoLogger() << "\e[31mmlibc: signal() always returns SIG_DFL\e[39m" << frg::endlog;
 	if(!mlibc::sys_sigaction) {
 		MLIBC_MISSING_SYSDEP();
 		errno = ENOSYS;
 		return SIG_ERR;
 	}
 	if(handler == SIG_DFL || handler == SIG_IGN) {
```
avdgrinten, line 15: Do we need this special case? Doesn't it to the same as the case below?

options/posix/generic/sys-uio.cpp
```diff
@@ -13,14 +18,32 @@ ssize_t writev(int fd, const struct iovec *iovs, int iovc) {
...
-		// This emulation here is not correct if there is an error!
-		__ensure(chunk > 0);
-		written += chunk;
+		if(SSIZE_MAX - bytes < iovs[i].iov_len) {
+			errno = EINVAL;
+			return -1;
+		}
+		bytes += iovs[i].iov_len;
 	}
+	char *buffer = (char *)malloc(bytes);
```
avdgrinten, line 29: We can use a `frg::vector<char, MemoryAllocator>` instead of `malloc`. (Use `.resize()` to resize the buffer and `.data()` to access the `char *`.)

options/ansi/generic/signal-stubs.cpp
Dennisbonke: It does something different, in this case we return the old handler, in the other case we don't have to do that

avdgrinten:
LGTM, thank you for the PR.

PR #133, opened 2020-07-22
struct rusage fix

BrainStackOverFlow:
i used this site as a source: https://www.gnu.org/software/libc/manual/html_node/Resource-Usage.html


Geertiebear:
Please submit a PR to the ``abi-break`` branch and not to ``master``

BrainStackOverFlow:
target branch is incorrect

PR #134, opened 2020-07-22
abis/mlibc: add missing struct rusage members

BrainStackOverFlow:
i used this site as a source: https://www.gnu.org/software/libc/manual/html_node/Resource-Usage.html

### DIFF
--- a/abis/mlibc/resource.h
+++ b/abis/mlibc/resource.h
@@ -6,6 +6,20 @@
 struct rusage {
 	struct timeval ru_utime;
 	struct timeval ru_stime;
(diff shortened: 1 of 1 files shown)


avdgrinten:
Does the ABI now match Linux?

Geertiebear:
musl also has a reserved field, but for the rest it looks the same ([source](https://github.com/bminor/musl/blob/91e662d1d941215eb024787db5e910dbfb5b169f/include/sys/resource.h#L27)).

avdgrinten:
Alright, I guess this can be merged.

Issue #99, opened 2020-05-31
getaddrinfo() improvements


Geertiebear:
As discussed on discord there are numerous improvements to be made to our current implementation of ``getaddrinfo()``
- [ ] Handle DNS ipv6 records
- [ ] Parse ``/etc/resolv.conf`` for the nameserver
- [ ] Parse ``/etc/hosts`` and use it in name resolution
- [ ] Parse ``/etc/services`` to implement the services part of the function

PR #135, opened 2020-07-22
various name resolving related improvements

Geertiebear:
Crosses off 4/5 points currently listed on issue #99, mainly:

- add ``/etc/services`` parsing
- add ``/etc/resolv.conf`` parsing
- add ``/etc/hosts`` parsing
- deal with CNAME dns cleaning up the alias related code

### DIFF
--- a/options/posix/generic/netdb-stubs.cpp
+++ b/options/posix/generic/netdb-stubs.cpp
@@ -66,12 +66,12 @@ int getaddrinfo(const char *__restrict node, const char *__restrict service,
 	if (serv_count < 0)
 		return -serv_count;
 
-	frg::vector<struct mlibc::dns_addr_buf, MemoryAllocator> addr_buf(getAllocator());
+	struct mlibc::lookup_result addr_buf;
 	int addr_count = 1;
 	frg::string<MemoryAllocator> canon{getAllocator()};
 	if (node) {
 		if ((addr_count = mlibc::lookup_name_hosts(addr_buf, node, canon)) <= 0)
-			addr_count = mlibc::lookup_name_dns(addr_buf, node);
+			addr_count = mlibc::lookup_name_dns(addr_buf, node, canon);
 		else
 			addr_count = 1;
 		if (addr_count < 0)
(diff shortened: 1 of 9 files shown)


options/posix/generic/netdb-stubs.cpp
```diff
@@ -53,52 +55,63 @@ int getaddrinfo(const char *__restrict node, const char *__restrict service,
...
-	if (ret < 0)
-		return ret;
-
-	auto out = (struct mlibc::ai_buf *) calloc(ret, sizeof(struct addrinfo));
-
-	auto canon_len = strlen(node);
-	auto canon = (char *) malloc(canon_len);
-	strncpy(canon, node, canon_len);
+	struct mlibc::service_buf serv_buf[SERV_MAX] = {};
+	int serv_count = mlibc::lookup_serv(serv_buf, service, protocol, socktype);
```
avdgrinten, line 65: It's guaranteed that this function only ever writes SERV_MAX = 2 entries? (If not, we could use a frg::small_vector or so.)

options/posix/generic/services.cpp
```diff
@@ -0,0 +1,136 @@
...
+			break;
+		}
+		if (!pos)
+			continue;
+
+		// now that we know the name is in this line we can parse
+		// the rest
+
+		// skip the name at the beginning of the line
+		for(pos = line; *pos && !isspace(*pos); pos++);
```
avdgrinten, line 54: Put the `;` onto the next line and indent it.

options/posix/include/mlibc/lookup.hpp
```diff
@@ -18,6 +18,13 @@ struct dns_addr_buf {
 	uint8_t addr[16];
 };
 
+struct lookup_result {
+	lookup_result() : buf(getAllocator()), aliases(getAllocator())
+	{}
```
avdgrinten, line 23: Coding style wise:
```suggestion
	lookup_result()
	: buf(getAllocator()), aliases(getAllocator()) {}
```
Also, `{}` initialization might be preferable to `()` initialization in general.

options/posix/generic/netdb-stubs.cpp
Geertiebear: yeah it is

PR #136, opened 2020-07-23
Update ar.h

aurelian2:
The extern "C" block is never terminated


avdgrinten:
LGTM,

ArsenArsen:
this block can be removed entirely, this section emits no symbols, hence is unaffected by extern

PR #145, opened 2020-08-28
Small additions

Dennisbonke:
This PR adds the following:

- 3 new `NETLINK_` constants, required by `python`'s socket module.

Furthermore, the following has been updated:

- `rewinddir()` has been implemented by `lseek()`.

### DIFF
--- a/options/posix/generic/dirent-stubs.cpp
+++ b/options/posix/generic/dirent-stubs.cpp
@@ -114,10 +114,12 @@ int readdir_r(DIR *dir, struct dirent *entry, struct dirent **result) {
 	*result = entry;
 	return 0;
 }
-void rewinddir(DIR *) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+
+void rewinddir(DIR *dir) {
+	lseek(dir->__handle, 0, SEEK_SET);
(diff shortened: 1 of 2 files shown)


options/posix/include/sys/socket.h
```diff
@@ -17,7 +17,7 @@ extern "C" {
 
 struct sockaddr {
 	sa_family_t sa_family;
-	char sa_data[];
+	char sa_data[14];
 };
```
avdgrinten, line 21: This is an ABI break.

options/posix/generic/dirent-stubs.cpp
```diff
@@ -114,10 +114,11 @@ int readdir_r(DIR *dir, struct dirent *entry, struct dirent **result) {
 	*result = entry;
 	return 0;
 }
-void rewinddir(DIR *) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+
+void rewinddir(DIR *dir) {
+	lseek(dir->__handle, 0, SEEK_SET);
```
Geertiebear, line 119: An ``lseek`` is not enough, we should also reset ``dir->__ent_next``

Issue #138, opened 2020-07-29
when compiling mlibc configured to ansi-c only, i get errors about missing posix headers


BrainStackOverFlow:
error message: https://pastebin.com/FiASwneX
mlibc meson.build: https://pastebin.com/G70T1S84

Geertiebear:
This issue should have been fixed in #147. @BrainStackOverFlow can you verify it works now?

Geertiebear:
I'll close this issue, we test against ansi-only builds now.

PR #147, opened 2020-08-28
Sysdeps refactor

Geertiebear:
Should be a fix for #138


options/ansi/generic/environment.cpp
```diff
@@ -106,7 +106,7 @@ char *getenv(const char *name) {
 	return const_cast<char *>(view.data() + s + 1);
 }
 
-int putenv(const char *string) {
+int __mlibc_putenv(char *string) {
```
avdgrinten, line 109: It would be nice to put this into the `mlibc::` namespace?

Geertiebear: Where would I put the declaration?

avdgrinten: Make some new ´environment.hpp` if nothing suitable exists?

avdgrinten:
LGTM (but there are conflicts). I did not check that the sysdeps files are minimal but from a quick glance, they look fine. After this is merged, the logic behind the sysdeps headers should be documented in the README (and also explain why functions can appear in multiple sysdeps headers).

PR #148, opened 2020-08-28
mlibc-gcc fixed and inet_pton implementation

### LINKED ISSUE #112: inet_pton() unimplemented
The following programs, tested on managarm, require `inet_pton()` to properly function

- `curl`, running `curl <ip>` and `curl <url>` crashes due to the `inet_pton()` function being unimplemented

Geertiebear:
Fixes #112 and makes ``mlibc-gcc`` more portable across distro's

### DIFF
--- a/options/posix/generic/arpa-inet-stubs.cpp
+++ b/options/posix/generic/arpa-inet-stubs.cpp
@@ -5,6 +5,7 @@
 #include <ctype.h>
 #include <stdio.h>
 #include <errno.h>
+#include <string.h>
 
 namespace {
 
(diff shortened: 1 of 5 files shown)


options/posix/generic/arpa-inet-stubs.cpp
```diff
@@ -146,8 +146,32 @@ const char *inet_ntop(int af, const void *__restrict src, char *__restrict dst,
 	errno = ENOSPC;
 	return NULL;
 }
-int inet_pton(int, const char *__restrict, void *__restrict) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+int inet_pton(int af, const char *__restrict src, void *__restrict dst) {
```
avdgrinten, line 149: Should `dst` be zero'd before writing to it?

Geertiebear: Not that I know of, and musl doesn't seem to do it either unless I'm misreading

avdgrinten: Alright.

options/posix/generic/arpa-inet-stubs.cpp
```diff
@@ -146,8 +146,32 @@ const char *inet_ntop(int af, const void *__restrict src, char *__restrict dst,
...
+	switch (af) {
+		case AF_INET: {
+			auto addr = reinterpret_cast<struct in_addr*>(dst);
+			auto dest = reinterpret_cast<uint8_t*>(&addr->s_addr);
```
avdgrinten, line 153: Why is this cast needed?

Geertiebear: ``s_addr`` is of type ``in_addr_t`` (aka. ``uint32_t``) so i cast it to a byte array instead.

avdgrinten: That's quite ugly. We should assemble the uint32_t locally or declare a local `uint8_t` array that we memcpy to `s_addr`.

avdgrinten (approved):
LGTM now.

PR #152, opened 2020-08-30
Fix static linking and add it to the ci

Geertiebear:


avdgrinten:
Hm, instead of add the static build to the same job as the non-static one, we should create an own job for it. We can use a build matrix as we do for the sysdeps.

Issue #58, opened 2020-04-18
printf formatting issues


Dennisbonke:
While testing the coreutils programs on managarm the `stat` program crashed with the following error:

`../../../src/mlibc/subprojects/frigg/include/frg/formatting.hpp:379: Assertion '!opts.left_justify' failed!`

This can be reproduced by `stat`ing any regular file.

Running the following command gives the same error, on a different location, this time at line 459
`readelf -I <executable>`

Dennisbonke:
Running `python3 --help` on Managarm produces the following error:

```
../../../src/mlibc/subprojects/frigg/include/frg/formatting.hpp:381: Assertion 'szmod == printf_siz

_mod::default_size' failed!
```

Geertiebear:
I think this issue can be closed right?

64:
I agree - can't reproduce anymore

Dennisbonke:
I'll test it in a bit and close it if I can't reproduce as well

Dennisbonke:
Can confirm, issue is fixed. Closing

PR #161, opened 2020-09-06
vsnprintf fixes and printf testing

### LINKED ISSUE #118: vsnprintf() behavior mismatch
`gdb` uses `vsnprintf()` to determine the size needed for a buffer like so

```
va_copy (vp, args);
size = vsnprintf (NULL, 0, fmt, vp);
va_end (vp);
```

However, while testing in managarm, this behaviour isn't matched, and size is 0, which crashes gdb. We should match the behaviour of glibc and the likes in this case

Geertiebear:
Fixes #118 and adds tests for #58.


avdgrinten (approved):
LGTM. I assume that this behavior matches the C standard?

Geertiebear:
Yes, it does.

PR #163, opened 2020-09-08
elf: Various Elf32 fixes required by cmake

Dennisbonke:
This PR adds various new structs required by cmake, it also fixes the declaration of Elf64_Dyn to match glibc and adjusts rtdl to reflect this.


avdgrinten (approved):
LGTM.

PR #164, opened 2020-09-09
Various libuv and cmake related changes

Dennisbonke:
This PR aims to stub and provide functions and structs as requested by `libuv` and `cmake`. To accomplish this, the following struct has been added:

- `struct mmsghdr`.

Furthermore, the following functions have been added and stubbed:

- `cfmakeraw()`.
- `dup3()`,
- `recvmmsg()`,
- `sendmmsg()`,
- `preadv()`,
- `pwritev()`.
- All the functions in the net-if-stubs.cpp file located in options/posix are now declared with `extern "C"` linkage to satisfy `libuv`.

### DIFF
--- a/options/posix/generic/net-if-stubs.cpp
+++ b/options/posix/generic/net-if-stubs.cpp
@@ -1,5 +1,6 @@
 
 #include <dirent.h>
+#include <net/if.h>
 #include <bits/ensure.h>
 
 void if_freenameindex(struct if_nameindex *) {
@@ -21,4 +22,3 @@ unsigned int if_nametoindex(const char *) {
 	__ensure(!"Not implemented");
 	__builtin_unreachable();
 }
-
--- a/options/posix/include/unistd.h
+++ b/options/posix/include/unistd.h
@@ -2,6 +2,7 @@
 #ifndef _UNISTD_H
 #define _UNISTD_H
 
+#include <bits/feature.h>
 #include <bits/types.h>
 #include <bits/size_t.h>
 #include <bits/ssize_t.h>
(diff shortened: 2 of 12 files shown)


options/posix/generic/unistd-stubs.cpp
```diff
@@ -791,6 +791,11 @@ int dup2(int fd, int newfd) {
 	return newfd;
 }
 
+int dup3(int fd, int newfd, int flags) {
```
Geertiebear, line 794: hmm ``dup3`` is linux specific... maybe this should go into the linux option?

options/posix/generic/net-if-stubs.cpp
```diff
@@ -2,6 +2,10 @@
 #include <dirent.h>
 #include <bits/ensure.h>
 
+#ifdef __cplusplus
```
Geertiebear, line 5: Don't do this. The reason why they are getting exported with C++ linkage is because the file is missing a ``#include <net/if.h>``

options/posix/include/unistd.h
```diff
@@ -123,6 +123,7 @@ ssize_t confstr(int, char *, size_t);
 char *crypt(const char *, const char *);
 int dup(int fd);
 int dup2(int src_fd, int dest_fd);
+int dup3(int fd, int newfd, int flags);
```
Geertiebear, line 126: Again, should maybe be in a linux only header

Geertiebear (changes requested):

options/posix/include/unistd.h
```diff
@@ -222,5 +222,9 @@ int pipe2(int *pipefd, int flags);
 }
 #endif
 
+#if __MLIBC_LINUX_OPTION
```
Geertiebear, line 225: Make sure to include ``bits/feature.h``

Geertiebear (changes requested):
One last minor issue, and also the commit history shouldn't include the previous stubbing of dup3 and the ``extern "C"`` hack for ``net-if-stubs``.

Geertiebear (approved):
LGTM

PR #173, opened 2020-10-01
Formatting

Geertiebear:
this PR won't build it's waiting on [frigg/#5](https://github.com/managarm/frigg/pull/5). Also the test for ``%lc`` is disabled because it currently crashes due to a missing locale implementation, but leaving it there for future use.

### DIFF
--- a/tests/ansi/locale.c
+++ b/tests/ansi/locale.c
@@ -1,0 +1,17 @@
+#include <stdio.h>
+#include <wchar.h>
+#include <locale.h>
+#include <assert.h>
+
+int main() {
+	wchar_t c = 0xC9;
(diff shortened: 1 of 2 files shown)


tests/ansi/locale.c
```diff
@@ -0,0 +1,17 @@
...
+#include <assert.h>
+
+int main() {
+	wchar_t c = 0xC9;
+	char buf[sizeof(wchar_t)] = { 0 };
+	setlocale(LC_ALL, "");
+	if (sprintf(buf, "%lc", c) < 0)
+		return -1;
+
+	assert(buf[0] == 0xffffffc3 && buf[1] == 0xffffff89
```
avdgrinten, line 13: That's a bit ugly, why not cast to unsigned char?

PR #174, opened 2020-10-12
Improve error handling for curl

Dennisbonke:
This PR changes some hard failures that are just assertions about not implemented features into soft failures, thus allowing `curl` to proceed further in Managarm.


Geertiebear:
LGTM

PR #175, opened 2020-10-13
ansi: Enable compilation of nearbyint and friends

Dennisbonke:
`GTK+3` requires `nearbyint` and friends, as we do have the declaration in `math.h`, but we don't compile it, it gets confused and fails to compile. This fixes that.


avdgrinten (approved):
LGTM.

PR #176, opened 2020-10-18
Implement lockf and fixed a bug in environment handling

### LINKED ISSUE #169: lockf() unimplemented
On managarm, `xbps` requires `lockf()`, which is unimplemented.

Dennisbonke:
This PR fixes a bug in the handling of environment variables found while working with `gtk3` on managarm. Furthermore, it implements `lockf()` like musl.

This PR fixes #169

### DIFF
--- a/options/ansi/generic/environment.cpp
+++ b/options/ansi/generic/environment.cpp
@@ -80,7 +80,8 @@ void unassign_variable(frg::string_view name) {
 	__ensure(environ == vector.data());
 
 	auto k = find_environ_index(name);
-	FRG_ASSERT(k != size_t(-1));
+	if(k == size_t(-1))
+		return;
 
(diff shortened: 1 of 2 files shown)


avdgrinten (commented):
LGTM. If the code is taken literally from musl, add a comment (the license allows this but we need to clarify the copyright status).

Dennisbonke:
Yeah that makes sense, I've added a comment that the code is from musl.

Geertiebear:
Hmm on linux ``lockf`` boils down to a ``fcntl`` but maybe we should have it as a sysdep instead? Other platforms might not implement it as a ``fcntl``

avdgrinten:
That could be done but OTOH it would probably lead to duplication of this code in multiple sysdeps. I think for now it's better to go with the Linux-style implementation.

PR #179, opened 2020-11-13
fixes for clang and libcxx

fido2020:


options/ansi/generic/stdio-stubs.cpp
```diff
@@ -898,6 +898,10 @@ int getchar(void) {
 	return fgetc(stdin);
 }
 
+char* gets(char* s){
```
ArsenArsen, line 901: maybe we should get not expose `gets` by default (and name its symbol something else, or this a static inline function, to prevent an undeclared gets linking properly)? it is actively being removed from standards:
```
       LSB  deprecates  gets().   POSIX.1-2008 marks gets() obsolescent.  ISO C11 removes the specification of gets() from the C language, and since
       version 2.16, glibc header files don't expose the function declaration if the _ISOC11_SOURCE feature test macro is defined.
```
from glibc `gets(3)`

options/rtdl/generic/main.cpp
```diff
@@ -108,10 +108,12 @@ extern "C" {
...
 }
 
-extern "C" [[gnu::alias("dl_debug_state"), gnu::visibility("default")]] void _dl_debug_state();
+#ifndef __clang__
+	extern "C" [[gnu::alias("dl_debug_state"), gnu::visibility("default")]] void _dl_debug_state();
+#endif
 
 // This symbol can be used by GDB to find the global interface structure
-[[gnu::visibility("default")]] DebugInterface *_dl_debug_addr = &globalDebugInterface;
+[[ gnu::visibility("default") ]] DebugInterface *_dl_debug_addr = &globalDebugInterface;
```
ArsenArsen, line 116: I think the spaces here collide with current convention, but I'm not sure

meson.build
```diff
@@ -39,26 +39,36 @@ if not headers_only
...
 	rtdl_deps += frigg_dep
 
-
 	add_languages('c', 'cpp')
 	c_compiler = meson.get_compiler('c')
 
 	add_project_arguments('-nostdinc', '-fno-builtin', language: ['c', 'cpp'])
 	add_project_arguments('-std=c++17', language: 'cpp')
 	add_project_arguments('-fno-rtti', '-fno-exceptions', language: 'cpp')
+	add_project_arguments('-isysroot=/home/computerfido/.local/share/lemon/sysroot/system', language: ['c', 'cpp'])
```
ArsenArsen: this is a problem

ArsenArsen (commented):
a few more comments:
- check the rest of your files for missing newlines at the end of files, and general code style errors
- I think it would be preferable to define a `weak_alias` macro than to repeat ourselves many times
- maybe this could be split up into a few more commits? plus adhering to our commit name practices would be nice

this is not a full review, just a few thoughts I had while reading the patch, so I will not mark it as requesting changes

options/rtdl/generic/main.cpp
```diff
@@ -108,10 +108,12 @@ extern "C" {
 	}
 }
 
-extern "C" [[gnu::alias("dl_debug_state"), gnu::visibility("default")]] void _dl_debug_state();
+#ifndef __clang__
```
Geertiebear, line 111: Why the ifdef for clang?

options/posix/generic/posix_locale.cpp
```diff
@@ -2,12 +2,15 @@
 #include <bits/ensure.h>
 
 locale_t newlocale(int category_mask, const char *locale, locale_t base) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+	// Due to all of the locale functions being stubs, the locale will not be used
```
Geertiebear, line 5: We should probably do some logging to say it's a stub.

options/posix/generic/posix_locale.cpp
```diff
@@ -2,12 +2,15 @@
...
+	return nullptr;
 }
 
 void freelocale(locale_t locobj) {
 	__ensure(!"Not implemented");
 	__builtin_unreachable();
 }
 
+locale_t uselocale(locale_t locobj) {
+	return nullptr;
```
Geertiebear, line 15: Same here.

options/glibc/include/features.h
```diff
@@ -0,0 +1,4 @@
+#ifndef FEATURES_H
```
Geertiebear, line 1: What's the point of this header?

options/ansi/musl-generic-math/signgam.c
```diff
@@ -1,6 +1,4 @@
 #include <math.h>
 //#include "libc.h"
 
-int __signgam = 0;
-
-weak_alias(__signgam, signgam);
```
Geertiebear, line 6: This weak alias shouldn't be removed.

options/ansi/generic/stdlib-stubs.cpp
```diff
@@ -342,11 +342,12 @@ int mblen(const char *mbs, size_t mb_limit) {
 }
 int mbtowc(wchar_t *__restrict wc, const char *__restrict mbs, size_t max_size) {
 	mlibc::infoLogger() << "mlibc: Broken mbtowc() called" << frg::endlog;
-	__ensure(wc);
-	__ensure(mbs);
+	//__ensure(wc);
```
Geertiebear, line 345: Commenting out these ensures is probably not such a good idea.

Geertiebear (changes requested):
Some comments + please fix up the commit history (e.g. remove the merge commit)

meson.build
fido2020: Ahh that I removed after I noticed I left that in

options/rtdl/generic/main.cpp
fido2020: Hmm, I noticed a mix of spaces and not spaces within existing code, that was changed because I had to add some gnu::visibility lines back

options/posix/generic/posix_locale.cpp
fido2020: Sure

options/glibc/include/features.h
fido2020: This header was just a stub libcxx did not attempt to use anything from it, just expected it to be there

meson.build
```diff
@@ -244,7 +253,7 @@ if not headers_only
 					libc_sources,
 					internal_sources,
 					ansi_sources,
-                                        lsb_sources,
+                    lsb_sources,
```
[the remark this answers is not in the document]
fido2020: My text editor had replaced that with spaces, I'll make sure to fix that

options/glibc/include/features.h
Geertiebear: Hmm... Maybe leave a comment stating that then.

options/rtdl/generic/main.cpp
fido2020: clang does not like the static alias, does dl_debug_state need to be static?

avdgrinten: What error does Clang print?

avdgrinten (approved):
LGTM.

options/ansi/generic/stdio-stubs.cpp
Geertiebear: Maybe this is more something for when we actually start adhering to feature test macros?

options/ansi/include/stdio.h
```diff
@@ -121,6 +121,7 @@ int fgetc(FILE *stream);
 char *fgets(char *__restrict buffer, size_t max_size, FILE *__restrict stream);
 int fputc(int c, FILE *stream);
 int fputs(const char *__restrict string, FILE *__restrict stream);
+char* gets(char* s);
```
Geertiebear, line 124: We use right side pointers everywhere.

options/rtdl/generic/main.cpp
Geertiebear: I think for the most part we use no spaces

PR #186, opened 2020-11-27
setjmp: Fix non spec-compliant longjmp for x86_64

streaksu:
The definition of `longjmp` specifies that it must make `setjmp` return 1 if the passed value is 0, else it must make it return said value. The current `x86_64` implementation does not take that into account, now it does.

Source: https://pubs.opengroup.org/onlinepubs/009695399/functions/longjmp.html, more specifically
```
After longjmp() is completed, program execution continues as if the corresponding invocation of setjmp() had just returned the value specified by val. The longjmp() function shall not cause setjmp() to return 0; if val is 0, setjmp() shall return 1.
```

### DIFF
--- a/options/internal/x86_64/setjmp.S
+++ b/options/internal/x86_64/setjmp.S
@@ -43,7 +43,9 @@ longjmp:
 	mov 0x20(%rdi), %r14
 	mov 0x28(%rdi), %r15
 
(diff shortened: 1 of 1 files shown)


avdgrinten:
Good catch. LGTM.

PR #188, opened 2020-12-02
options/posix: fix getaddrinfo() service resolution

Geertiebear:


tests/posix/getaddrinfo.c
```diff
@@ -21,5 +21,12 @@ int main() {
 	assert(res[0].ai_protocol == IPPROTO_TCP);
 
 	freeaddrinfo(res);
+
+	/* check we can resolve any domain */
+	ret = getaddrinfo("example.net", NULL, &hints, &res);
+	assert(ret == 0);
+
+	freeaddrinfo(res);
```
ArsenArsen, line 29: there is a double-free here

options/posix/generic/services.cpp
```diff
@@ -83,39 +84,43 @@ static int lookup_serv_file(struct service_buf *buf, const char *name,
...
 int lookup_serv(struct service_buf *buf, const char *name, int proto,
-		int socktype) {
+		int socktype, int flags) {
 	switch(socktype) {
 		case SOCK_STREAM:
 			if (!proto)
 				proto = IPPROTO_TCP;
 			else if (proto != IPPROTO_TCP)
-				return EAI_SERVICE;
+				return -EAI_SERVICE;
```
avdgrinten, line 93: Why the minus?

Geertiebear: A positive return value indicates success, so I return a negative one, and then make it positive again in ``getaddrinfo()``

avdgrinten: Okay, maybe add a comment to state that it works like this (?).

Other than that, the PR LGTM.

PR #189, opened 2020-12-06
Add preliminary aarch64 support

qookei:


meson.build
```diff
@@ -248,9 +261,12 @@ if not headers_only
 				],
 				cpp_args: ['-DFRIGG_HAVE_LIBC'],
 				include_directories: libc_include_dirs,
-				dependencies: libc_deps,
+				# Link with libgcc to get math symbols ...
+				dependencies: [libc_deps, libgcc],
 				link_with: [ldso_lib],
 				link_whole: [libc_sublibs, ldso_lib],
+				# ... but don't export them ourselves
+				link_args: ['-Wl,--exclude-libs=gcc'],
```
avdgrinten, line 269: AFAIU `--exclude-libs` does not work in the static case (ld is not even called here). However, that should not be an issue since the final executable will link against libgcc anyway.

options/internal/aarch64/gcc-extra/cxxabi.cpp
```diff
@@ -0,0 +1,11 @@
+
+#include <bits/ensure.h>
+
+// The cxxabi needs operator delete for *deleting* destructors, i.e., destructors that
+// are called by delete expressions. We never use such expressions in mlibc.
+// Note that G++ complains if we make the operator hidden,
+// thus we use it's mangled name as a workaround.
+extern "C" [[gnu::visibility("hidden")]] void _ZdlPvm() { // operator delete (void *, size_t)
```
avdgrinten, line 8: Is this different from the x86_64 case?

options/internal/aarch64/setjmp.S
```diff
@@ -0,0 +1,21 @@
+// TODO: implement this
+
+.type __setjmp, "function"
+__setjmp:
+	brk #1
```
avdgrinten, line 5: Unimplemented functions should log a message.

options/rtdl/generic/main.cpp
```diff
@@ -136,9 +140,12 @@ extern "C" void *interpreterMain(uintptr_t *entry_stack) {
 		mlibc::infoLogger() << "ldso: Own dynamic section is at: " << _DYNAMIC << frg::endlog;
 	}
 
+// FIXME: on aarch64 these lines corrupt unrelated GOT entries (entries for ld.so functions)
```
avdgrinten, line 143: Remove the FIXME (the ABI just seems to differ here and that is fine).

sysdeps/managarm/aarch64/signals.S
```diff
@@ -0,0 +1,8 @@
+
+.section .text
+.global __mlibc_signal_restore
+__mlibc_signal_restore:
+	ldr x0, =0x80000006
+	svc 0
+	brk #1
```
avdgrinten, line 7: Is this tested?

avdgrinten (commented):
In addition to the other comments: why are some files in `.../x86/...` dirs and others in `.../x86_64/...`?

sysdeps/managarm/aarch64/signals.S
qookei: This can't really be tested until we get posix-subsystem running, but it's just equivalent to the x86_64 code (supercall 6 and `__builtin_trap()`)

options/internal/aarch64/gcc-extra/cxxabi.cpp
qookei: All the files in that subdir were the same so I just use the top-level copy instead now.

qookei: Update: I didn't notice `mlibc_crtend.S` differs because it contains actual asm.

meson.build
```diff
@@ -8,12 +8,19 @@ rtdl_include_dirs = [
...
-
 ]
+
+if host_machine.cpu_family() == 'x86_64'
+	libc_include_dirs += include_directories('options/internal/x86_64-include')
+elif host_machine.cpu_family() == 'aarch64'
+	libc_include_dirs += include_directories('options/internal/aarch64-include')
+else
+	error('Unknown architecture')
+endif
```
notYuriy, line 22: If statements can be replaced with
```
include_directories('options/internal' / host_machine.cpu_family() + '-include')
```
so adding more architectures won't require additional changes

Issue #168, opened 2020-09-25
fsync() unimplemented


Dennisbonke:
On managarm, `xbps` requires `fsync()`, which is unimplemented

Issue #55, opened 2020-04-17
sync() unimplemented


Dennisbonke:
While testing the coreutils programs the following utilities required `sync()`

- `sync`

no92:
Fixed by #196.

Issue #51, opened 2020-04-17
fdatasync() unimplemented


Dennisbonke:
While testing the coreutils programs the following utilities required `fdatasync()`

- `shred`

The following programs also requires `fdatasync()`

- `update-mime-info` from the package `shared-mime-info`
- `xbps-pkgdb` from the package `xbps`

no92:
Implemented by #196.

PR #196, opened 2020-12-29
options/posix: Wire up some functions to syscalls

Dennisbonke:
This PR wires the following functions up to a syscall (syscall is unimplemented for now, as this requires work in managarm first).

- `sync()`, wired up to a `sys_sync()` function,
- `fdatasync(int fd)`, wired up to a `sys_fdatasync(int fd)` function.

This PR technically fulfills #51 and #55, but seeing that they are not implemented in any sysdep (and certainly not in managarm, as the issue states), they, just like #168, can **not** be closed by this issue.


options/posix/generic/unistd-stubs.cpp
```diff
@@ -600,9 +612,16 @@ int symlinkat(const char *target_path, int dirfd, const char *link_path) {
 	}
 	return 0;
 }
+
 void sync(void) {
-	__ensure(!"Not implemented");
+	if(!mlibc::sys_sync) {
+		MLIBC_MISSING_SYSDEP();
+		errno = ENOSYS;
```
avdgrinten, line 619: `sync()` should not set errno.

avdgrinten (commented):
Small comment below. I think we can actually close these mlibc issues even though no sysdeps implement the hooks. From mlibc's side, the work is done here; now the sysdeps need to deliver implementatios.

Dennisbonke:
Considering that the issues explicitly state that on managarm, that function is unimplemented, I kinda see that as please implement the sysdep, then close the issue, but that is my opinion

avdgrinten (approved):
LGTM.

PR #197, opened 2020-12-30
options/linux: Implement getrandom

Dennisbonke:
This PR implements `getrandom` via `sys_getentropy()`, this should remove the `mlibc: getrandom() is a no-op` message shown on managarm boot.


avdgrinten (approved):
LGTM.

PR #201, opened 2021-01-11
options/posix: Implement posix_spawn

Dennisbonke:
This PR implements `posix_spawn` in the same way that musl does. Furthermore, it fixes some off-by-one bugs in some signal related functions, adds a test for `posix_spawn` and adds several linux syscalls to fulfill the test.

This PR is not really blocked on managarm/managarm#243 but should be merged together with it, to avoid breaking managarm

### DIFF
--- a/options/posix/generic/spawn-stubs.cpp
+++ b/options/posix/generic/spawn-stubs.cpp
@@ -1,14 +1,251 @@
 
 #include <spawn.h>
+#include <errno.h>
+#include <pthread.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <limits.h>
+#include <sched.h>
+#include <signal.h>
+#include <sys/wait.h>
 
 #include <bits/ensure.h>
+#include <mlibc/debug.hpp>
 
-int posix_spawn(pid_t *__restrict pid, const char *__restrict path,
+/*
+ * Musl places this in a seperate header called fdop.h
+ * This header isn't present in glibc, or on my host, so I
+ * include it's contents here
+ */
+
+#define FDOP_CLOSE 1
+#define FDOP_DUP2 2
+#define FDOP_OPEN 3
+#define FDOP_CHDIR 4
+#define FDOP_FCHDIR 5
+
+struct fdop {
+	struct fdop *next, *prev;
+	int cmd, fd, srcfd, oflag;
+	mode_t mode;
+	char path[];
+};
+
+/*
+ * This posix_spawn implementation is taken from musl
+ */
+
+static unsigned long handler_set[NSIG / (8 * sizeof(long))];
+
+static void __get_handler_set(sigset_t *set) {
+	memcpy(set, handler_set, sizeof handler_set);
+}
+
+struct args {
+	int p[2];
+	sigset_t oldmask;
+	const char *path;
+	const posix_spawn_file_actions_t *fa;
+	const posix_spawnattr_t *__restrict attr;
+	char *const *argv, *const *envp;
+};
+
+static int child(void *args_vp) {
+	int i, ret;
+	struct sigaction sa = {0};
+	struct args *args = (struct args *)args_vp;
+	int p = args->p[1];
+	const posix_spawn_file_actions_t *fa = args->fa;
+	const posix_spawnattr_t *__restrict attr = args->attr;
+	sigset_t hset;
+	bool use_execvpe = false;
+
+	if(attr->__fn)
+		use_execvpe = true;
+
+	close(args->p[0]);
+
+	/* All signal dispositions must be either SIG_DFL or SIG_IGN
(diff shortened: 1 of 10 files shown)


meson.build
```diff
@@ -45,7 +45,7 @@ if not headers_only
 
 	add_project_arguments('-nostdinc', '-fno-builtin', language: ['c', 'cpp'])
 	add_project_arguments('-std=c++17', language: 'cpp')
-	add_project_arguments('-fno-rtti', '-fno-exceptions', language: 'cpp')
+	add_project_arguments('-fno-rtti', '-fno-exceptions', '-fpermissive', language: 'cpp')
```
Geertiebear, line 48: Don't add the flag, just fix the compiler warnings/errors

options/posix/generic/spawn-stubs.cpp
```diff
@@ -1,14 +1,226 @@
...
+#include <mlibc/debug.hpp>
 
-int posix_spawn(pid_t *__restrict pid, const char *__restrict path,
+/*
+ * This posix_spawn implementation is taken from musl
+ */
+
+static unsigned long handler_set[NSIG / (8 * sizeof(long))];
+
+void __get_handler_set(sigset_t *set) {
```
Geertiebear, line 21: Can this be static?

options/posix/generic/spawn-stubs.cpp
```diff
@@ -1,14 +1,226 @@
...
+	}
+
+	if(attr->__flags & POSIX_SPAWN_SETSID) {
+		if((ret = setsid()) < 0)
+			goto fail;
+	}
+
+	if(attr->__flags & POSIX_SPAWN_SETPGROUP) {
+		mlibc::infoLogger() << "mlibc: posix_spawn: ignoring SETPGROUP" << frg::endlog;
+		//if((ret = setpgid(0, attr->__pgrp)))
```
Geertiebear, line 80: ?

options/posix/generic/spawn-stubs.cpp
```diff
@@ -1,14 +1,226 @@
...
+	args.path = path;
+	args.fa = file_actions;
+	args.attr = attrs ? attrs : &(const posix_spawnattr_t){0};
+	args.argv = argv;
+	args.envp = envp;
+	pthread_sigmask(SIG_BLOCK, SIGALL_SET, &args.oldmask);
+
+	/* The lock guards both against seeing a SIGABRT disposition change
+	 * by abort and against leaking the pipe fd to fork-without-exec. */
+	//LOCK(__abort_lock);
```
Geertiebear, line 187: The comment above makes it seem like this isn't entirely safe to ignore

options/posix/generic/spawn-stubs.cpp
```diff
@@ -1,14 +1,226 @@
...
+	if(pipe2(args.p, O_CLOEXEC)) {
+		//UNLOCK(__abort_lock);
+		ec = errno;
+		goto fail;
+	}
+
+	/* Mlibc change: We use fork + execve, as clone is not implemented.
+	 * This yields the same result in the end. */
+	//pid = clone(child, stack + sizeof stack, CLONE_VM | CLONE_VFORK | SIGCHLD, &args);
+	pid = fork();
```
Geertiebear, line 198: I still think this should be implemented with a ``sys_vfork`` sysdep. On linux this would result in a ``clone`` syscall, but on other platforms it could be something else.

options/posix/include/pthread.h
```diff
@@ -50,6 +50,8 @@ extern "C" {
 #define PTHREAD_MUTEX_INITIALIZER {0, 0, 0}
 #define PTHREAD_RWLOCK_INITIALIZER {0, 0, 0}
 
+#define SIGALL_SET ((sigset_t *)(const unsigned long long [2]){ (unsigned long)-1, (unsigned long)-1 })
```
Geertiebear, line 53: This macro is broken, mlibc's ``sigset_t`` is of type ``long`` not ``uint64_t[2]``

Geertiebear (changes requested):

options/posix/generic/spawn-stubs.cpp
Geertiebear: Ah I see, you are ignoring SETPGROUP, my bad. Is there a reason not to implement it right now?

Dennisbonke: Managarm doesn't have the sysdep / the kernel logic to handle it, that is why I ignored it

Geertiebear: Hmmm, would still be nice to have for other platforms that might support it

Geertiebear (approved):
LGTM now, although I think the commit history could be cleaned up preferrably.

PR #208, opened 2021-01-15
sysdeps/managarm: Fix bug in sys_connect and convert to helix_ng

Dennisbonke:
We also handle ENOENT correctly now in `sys_connect()`, the managarm side is yet to be changed (soon), but this pr is not blocked on that change.


avdgrinten (approved):
LGTM.

PR #213, opened 2021-01-19
Stub out duplocale

no92:
This is needed for `json-c`.


Geertiebear (approved):
LGTM but maybe for the future our commit style is usually ``options/<option>: change`` so e.g. it would be ``options/posix: stub out duplocale`` in this case.

PR #218, opened 2021-01-24
options/linux: Add changes needed for libinput 1.16.4 and wlroots

no92:
This moves some #defines out of `<linux/input.h>` into `<linux/input-event-codes.h>`, as both packages expect them to be. The former header now includes the new one.

Also, `<linux/major.h>` gets a new #define as needed for wlroots.

For libinput to work, this depends on managarm/bootstrap-managarm#80


avdgrinten (approved):
LGTM.

PR #214, opened 2021-01-21
Various new stubs and defines

Dennisbonke:
This PR adds several stubs and defines that are required by the bare minimum `util-linux` build configuration. I expect to add more to this pr when we enable more functionality of `util-linux`.

This PR adds the following new functions, definitions and structs:
- `SCNd64` define for scanf,
- `tsearch()`,
- `tfind()`,
- `tdelete()`,
- `msgget()`,
- `semget()`,
- Several defines defining tty defaults, matching musl's,
- `timercmp()`,
- `TIMEVAL_TO_TIMESPEC()`,
- `TIMESPEC_TO_TIMEVAL()`,
- `ENOMEDIUM` errno constant,
- 2 CDROM related defines from the Linux kernel,
- `SIOCGIFCONF` ioctl command,
- `struct ifconf`,
- `wcswidth()`,
- `versionsort()`,
- 5 blockdev related defines,
- `umount()`,
- `umount2()`,
- `ENOTBLK` errno constant,
- `O_PATH` constant,
- `strverscmp()`.

PR #220, opened 2021-01-28
stuff for clang/LLVM

fido2020:


avdgrinten:
#214 added wait4 concurrently. Are the off64_t functions exported by other libcs (like musl or glibc)?

fido2020:
I noticed that the wait4 added just throws an assertion, would it not be better to just call waitpid? It would be better I replace lseek64 with a define as that appears to be the way musl defines it. glibc's lseek64 is system dependent.

avdgrinten:
I would actually prefer proper type defs and a proper lseek64 function over macros. However, it is not clear to me that they should be defined at all, since they seem to be compatibility hacks on 32-bit systems. Does llvm call them directly?

sysdeps/lemon/generic/lemon.cpp
```diff
@@ -72,6 +72,8 @@ namespace mlibc{
 
 	void sys_exit(int status){
 		syscall(SYS_EXIT, status);
+
+		for(;;); // Prevent compiler warning
```
Geertiebear, line 76: This could also be done with ``__builtin_unreachable`` no?

fido2020: Yes, fixeed

fido2020:
LLVM calls it once, I could just patch my LLVM if you would prefer that?

avdgrinten (approved):
LGTM.

PR #221, opened 2021-01-29
sysdeps/managarm: Change the startup code to use the generic version

### LINKED ISSUE #41: program_invocation_name is NULL
A lot of programs get their own name by looking at program_invocation_name or program_invocation_short_name. Right now, those are NULL. Setting these correctly would fix this.

Dennisbonke:
This PR makes managarm use the generic entry code instead of some custom thing. This fixes #41


avdgrinten (commented):
LGTM.

PR #223, opened 2021-01-30
meta: Split libintl functions to its own option

Dennisbonke:
This PR splits off the functions found in `libintl-stubs.cpp` to it's own option. This is in preparation for managarm porting the `libintl` library provided by `gettext`.

After this is merged, managarm/bootstrap-managarm#83 should be merged


meson.build
```diff
@@ -141,6 +142,11 @@ if not disable_iconv_option
 	libc_include_dirs += include_directories('options/iconv/include')
 endif
 
+if not disable_intl_option
+	rtdl_include_dirs += include_directories('options/intl/include')
+	libc_include_dirs += include_directories('options/intl/include')
+endif
```
avdgrinten, line 148: Does rtdl require these headers at all?

options/intl/generic/libintl-stubs.cpp
```diff
@@ -2,49 +2,59 @@
 #include <bits/ensure.h>
 
 char *gettext(const char *msgid) {
+	return (char *)msgid;
+	return nullptr;
 	__ensure(!"Not implemented");
 	__builtin_unreachable();
 }
```
avdgrinten, line 9: Huh? Why do we return nullptr here (and in the other functions)? Why are there two returns?

avdgrinten (changes requested):

avdgrinten (approved):
LGTM.

PR #226, opened 2021-02-02
options/linux: Add TIOCSTI

Dennisbonke:
This PR adds a ioctl define needed for `cmatrix`

### DIFF
--- a/options/linux/include/asm/ioctls.h
+++ b/options/linux/include/asm/ioctls.h
@@ -6,8 +6,9 @@
 #define TCGETS 0x5401
 #define TCSETS 0x5402
 #define TIOCSCTTY 0x540E
-#define TIOCNOTTY 0x5422
+#define TIOCSTI 0x5412
 #define TIOCGWINSZ 0x5413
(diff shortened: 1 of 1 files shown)


options/linux/include/asm/ioctls.h
```diff
@@ -6,8 +6,9 @@
 #define TCGETS 0x5401
 #define TCSETS 0x5402
 #define TIOCSCTTY 0x540E
-#define TIOCNOTTY 0x5422
+#define TIOCSTI 0x5412
```
Geertiebear, line 9: Was this one wrong before?

Dennisbonke: No, I moved the `TIOCNOTTY` define to the bottom of the list, as it has the highest number. The value is still the same, the only change is that the list is now sorted.

Geertiebear: Ah okay I just can't read I see hahahahah

Geertiebear (approved):
LGTM

PR #228, opened 2021-02-03
sysdeps/managarm: Handle EINVAL in sigaction

Dennisbonke:
This is the mlibc counterpart of the signal changes in managarm/managarm#246


avdgrinten (approved):
LGTM.

PR #236, opened 2021-03-04
Fix aarch64 TLS-related issues

qookei:


options/rtdl/generic/linker.cpp
```diff
@@ -1017,13 +1045,22 @@ void Loader::_buildTlsMaps() {
 			runtimeTlsMap->indices.push_back(object);
 
 			__ensure(16 % object->tlsAlignment == 0);
-			runtimeTlsMap->initialPtr += object->tlsSegmentSize;
-			size_t misalign = runtimeTlsMap->initialPtr % object->tlsAlignment;
+			auto ptr = runtimeTlsMap->initialPtr + object->tlsSegmentSize;
+			size_t misalign = ptr % object->tlsAlignment;
```
avdgrinten, line 1049: If tlsAlignment is guaranteed to be a power of 2, this can be done with a bit shift.

avdgrinten: Why do we nee to store `ptr`? IMHO the old code that modified `runtimeTlsMap->initialPtr` was easier to read. Why don't we keep modifying that var directly? (inside the `if constexpr`).

avdgrinten (approved):
LGTM.

PR #239, opened 2021-03-15
options/posix: implement getnameinfo()

Geertiebear:


options/posix/include/mlibc/lookup.hpp
```diff
@@ -44,8 +46,10 @@ struct ai_buf {
 
 int lookup_name_dns(struct lookup_result &buf, const char *name,
 		frg::string<MemoryAllocator> &canon_name);
+int lookup_addr_dns(frg::span<char> name, frg::array<uint8_t, 16> &addr, int family);
```
avdgrinten, line 49: `string_view` probably makes more sense than `span<char>`.

Geertiebear: A ``string_view`` is immutable, but I need to be able to overwrite the buffer

avdgrinten: Alright. So the result is fixed-length? If it isn't, it might make more sense to return a `string`.

Geertiebear: Yeah, it's always limited by the size of the buffer the user passes.

avdgrinten:
From my side, we can merge this PR. Is it properly tested and does it work?

Geertiebear:
Yeah, Dennis said it worked, so it should be good to go.

avdgrinten:
Alright, let's merge this. Thanks for the PR!

PR #240, opened 2021-03-19
options/ansi: avoid uneeded lseek()'ing in file writes

Geertiebear:
When doing large (causing a buffer flush) consecutive writes we call ``lseek(fd, 0 SEEK_CUR)`` for every flush operation, essentially doubling the needed syscall overhead. So we should only lseek when ``__io_offset != __dirty_begin``.


avdgrinten (approved):
LGTM.

PR #241, opened 2021-03-20
For now adding a stub, more research is required

sappigeninja:
For now im adding a stub for the FIONREAD operation.
Im doing some more research on how to implement it (if its actually needed that is).

### DIFF
--- a/sysdeps/managarm/generic/file.cpp
+++ b/sysdeps/managarm/generic/file.cpp
@@ -2859,6 +2859,12 @@ int sys_ioctl(int fd, unsigned long request, void *arg, int *result) {
 		*result = resp.result();
 		return 0;
 	}
+	case FIONREAD: {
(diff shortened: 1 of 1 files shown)


sysdeps/managarm/generic/file.cpp
```diff
@@ -1986,7 +1986,7 @@ int sys_ioctl(int fd, unsigned long request, void *arg, int *result) {
 		auto send_req = parseSimple(element);
 		auto recv_resp = parseInline(element);
 		auto recv_list = parseLength(element);
-
+nn
```
Dennisbonke, line 1989: I assume this is committed by accident?

sappigeninja: jup
fixed it just now

avdgrinten:
We silently return success here. That is probably not a good idea. What do we need this ioctl for? We should probably implement it properly.

qookei:
Superseded by #358

PR #245, opened 2021-04-13
Add missing defines for Rust libc

64:
Adds the following constants:

- `IPV6_ADD_MEMBERSHIP` (defined in `glibc` [here](https://sourceware.org/git/?p=glibc.git;a=blob;f=bits/in.h;h=35a95f69fb8e9d98a4853e5a68742355157a4609;hb=HEAD#l108))
- `IPV6_DROP_MEMBERSHIP`
-  `PTHREAD_STACK_MIN`, as per POSIX. We set it to the same value as glibc - 16kb.


Geertiebear (approved):
LGTM

Issue #18, opened 2019-09-22
Missing getline/getdelim


alula:
Prevents many programs from compiling, standarized by POSIX.

http://man7.org/linux/man-pages/man3/getline.3.html

avdgrinten:
Thank you for the report. This is indeed something we should fix.

Out of curiosity: on which OS do you use mlibc?

gabixdev:
I've tried to build some stuff on qword and had to add a stub for these functions to make it build

PR #249, opened 2021-04-15
posix: Implement getline and getdelim

64:
Fixes managarm/mlibc#18.

### DIFF
--- a/options/ansi/generic/stdio-stubs.cpp
+++ b/options/ansi/generic/stdio-stubs.cpp
@@ -1016,8 +1016,6 @@ ssize_t getdelim(char **line, size_t *n, int delim, FILE *stream) {
 		return -1;
 	}
 
-	constexpr size_t chunk = 1024;
-
 	char *buffer = *line;
 	size_t capacity = *n, nwritten = 0;
 
@@ -1034,7 +1032,7 @@ ssize_t getdelim(char **line, size_t *n, int delim, FILE *stream) {
 
 	while (true) {
 		// Fill the buffer
-		while (capacity > 0 && nwritten < capacity - 1) {
+		while (buffer && capacity > 0 && nwritten < capacity - 1) {
 			auto c = fgetc_unlocked(stream);
 			if (ferror(stream)) {
 				return -1;
@@ -1053,7 +1051,7 @@ ssize_t getdelim(char **line, size_t *n, int delim, FILE *stream) {
 
 		// Double the size of the buffer (but make sure it's at least 1024)
 		capacity = (capacity >= 1024) ? capacity * 2 : 1024;
-		buffer = reinterpret_cast<char *>(getAllocator().reallocate(*line, capacity));
+		buffer = reinterpret_cast<char *>(realloc(*line, capacity));
 		if (!buffer) {
 			errno = ENOMEM;
 			return -1;
--- a/options/ansi/generic/stdio-stubs.cpp
+++ b/options/ansi/generic/stdio-stubs.cpp
@@ -1051,7 +1051,7 @@ ssize_t getdelim(char **line, size_t *n, int delim, FILE *stream) {
 
 		// Double the size of the buffer (but make sure it's at least 1024)
 		capacity = (capacity >= 1024) ? capacity * 2 : 1024;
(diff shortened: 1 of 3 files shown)


options/ansi/generic/stdio-stubs.cpp
```diff
@@ -1003,46 +1003,69 @@ void perror(const char *string) {
...
+				return nwritten;
+			}
+		}
 
-ssize_t getdelim(char **line, size_t *n, int delim, FILE *file_base) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+		// Double the size of the buffer (but make sure it's at least 1024)
+		capacity = (capacity >= 1024) ? capacity * 2 : 1024;
+		buffer = reinterpret_cast<char *>(getAllocator().reallocate(*line, capacity));
```
Geertiebear, line 1056: In general use ``realloc`` directly, and not frigg's allocator. Only use the latter when actually dealing with something from frigg.

options/ansi/generic/stdio-stubs.cpp
```diff
@@ -1003,46 +1003,69 @@ void perror(const char *string) {
...
 		return -1;
 	}
 
-	char *buffer = nullptr;
-	size_t capacity = 0;
-	if(*line) {
-		buffer = *line;
-		capacity = *n;
-	}
+	constexpr size_t chunk = 1024;
```
Geertiebear, line 1019: This variable seems unused.

options/ansi/generic/stdio-stubs.cpp
```diff
@@ -1003,46 +1003,69 @@ void perror(const char *string) {
...
+	auto c = fgetc_unlocked(stream);
+	if (c == EOF || ferror(stream)) {
 		return -1;
+	} else {
+		file->unget(c);
+	}
+
+	while (true) {
+		// Fill the buffer
+		while (capacity > 0 && nwritten < capacity - 1) {
```
Geertiebear, line 1037: What happens when capacity is not 0, but the buffer is still NULL? Maybe that wouldn't happen but might still be a good case to catch.

tests/posix/getdelim.c
```diff
@@ -0,0 +1,59 @@
...
+	}
+	assert(!ferror(fp));
+	free(line);
+	fclose(fp);
+
+	size_t nchars = 10000;
+	fp = fopen(TEST_FILE, "w");
+	assert(fp);
+	for (int i = 0; i < nchars; i++)
+		fputc('a', fp);
```
Geertiebear, line 34: We could use`` fwrite()`` here

Geertiebear (changes requested):

tests/posix/getdelim.c
Geertiebear: Actually, no we can't

avdgrinten:
I think we should actually use the allocator directly for mlibc-internal uses. That way, we know that if something faults in realloc, it is called by user code. (realloc is implemented on top of getAllocator(), not the other way around).

options/ansi/generic/stdio-stubs.cpp
avdgrinten: We could also just use `frg::small_vector` instead of open coding a small vector.

avdgrinten: Never mind, that doesn't work because we need to return the pointer.

Geertiebear:
> I think we should actually use the allocator directly for mlibc-internal uses. That way, we know that if something faults in realloc, it is called by user code. (realloc is implemented on top of getAllocator(), not the other way around).

How useful is this really? Considering we have gdb support and we can usually addr2line. Also is it safe to assume that no code will break if you allocate something with ``getAllocator()`` and free it using ``free()``?

Geertiebear (approved):
Aside from the ``realloc()`` discussion, LGTM now

avdgrinten:
`free()` is implemented as `getAllocator()->free()`, so that is always safe.

Geertiebear:
Ah alright, well in that case I guess using ``getAllocator()`` is a good idea.

64:
I think this is ready now

avdgrinten:
Thanks for the PR. I'm merging this now since Geert already approved it.

PR #250, opened 2021-04-15
Python fixes

Dennisbonke:
This PR adds more error handling code to several of Managarms sysdeps and adds proper handling to `wcsrtombs()` and `wcsrntombs()` when the first argument is NULL. This also allows the python dynamic interpreter to work on Managarm. Furthermore, a bug in the pthread cancel implementation was found and fixed,  `_CS_GNU_LIBC_VERSION` is now handled in `confstr()` and the buffer mode of `stderr` is now correctly set.

Blocked on managarm/managarm#351, but should be merged at the same time to avoid build failures.

### DIFF
--- a/options/ansi/include/mlibc/file-io.hpp
+++ b/options/ansi/include/mlibc/file-io.hpp
@@ -76,7 +76,7 @@ public:
 };
 
 struct fd_file : abstract_file {
-	fd_file(int fd, void (*do_dispose)(abstract_file *) = nullptr);
+	fd_file(int fd, void (*do_dispose)(abstract_file *) = nullptr, bool force_unbuffered = false);
 
 	int fd();
 
@@ -93,6 +93,7 @@ protected:
 private:
 	// Underlying file descriptor.
 	int _fd;
+	bool _force_unbuffered;
 };
 
 } // namespace mlibc
--- a/options/posix/generic/unistd-stubs.cpp
+++ b/options/posix/generic/unistd-stubs.cpp
@@ -58,6 +58,9 @@ ssize_t confstr(int name, char *buf, size_t len) {
 	} else if(name == _CS_GNU_LIBPTHREAD_VERSION) {
 		// We are not glibc, so we can return 0 here.
 		return 0;
+	} else if(name == _CS_GNU_LIBC_VERSION) {
+		// We are not glibc, so we can return 0 here.
+		return 0;
 	} else {
 		mlibc::infoLogger() << "\e[31mmlibc: confstr() request " << name << " is unimplemented\e[39m"
 				<< frg::endlog;
--- a/options/posix/generic/unistd-stubs.cpp
+++ b/options/posix/generic/unistd-stubs.cpp
@@ -45,9 +45,16 @@ int fchdir(int fd) {
 	return 0;
 }
 
-int chown(const char *, uid_t, gid_t) {
-	mlibc::infoLogger() << "\e[31mmlibc: chown() is not implemented correctly\e[39m"
-			<< frg::endlog;
+int chown(const char *path, uid_t uid, gid_t gid) {
+	if(!mlibc::sys_fchownat) {
+		MLIBC_MISSING_SYSDEP();
+		errno = ENOSYS;
+		return -1;
+	}
+	if(int e = mlibc::sys_fchownat(AT_FDCWD, path, uid, gid, 0); e) {
+		errno = e;
+		return -1;
+	}
 	return 0;
 }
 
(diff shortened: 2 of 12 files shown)


options/ansi/include/mlibc/file-io.hpp
```diff
@@ -76,7 +76,7 @@ struct abstract_file : __mlibc_file_base {
 };
 
 struct fd_file : abstract_file {
-	fd_file(int fd, void (*do_dispose)(abstract_file *) = nullptr);
+	fd_file(int fd, void (*do_dispose)(abstract_file *) = nullptr, bool is_unbuffered = false);
```
avdgrinten, line 79: Rename `is_unbuffered` -> `force_unbuffered`.

Dennisbonke: I’ll perform the rename when I’m back home again.

options/posix/generic/unistd-stubs.cpp
```diff
@@ -379,8 +379,9 @@ pid_t getsid(pid_t pid) {
 }
 
 int lchown(const char *path, uid_t uid, gid_t gid) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+	mlibc::infoLogger() << "\e[31mmlibc: lchown() is not implemented correctly\e[39m"
+			<< frg::endlog;
+	return 0;
 }
```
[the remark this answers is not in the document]
Dennisbonke: While I agree, `chown` is also stubbed like this. It seems that there is no sysdep for any of the `chown` functions at all in mlibc. That is why I did it like this.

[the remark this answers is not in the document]
avdgrinten: I think we should simply add a sysdep for chownat, the function has a simple interface anyway.

options/posix/include/mlibc/posix-sysdeps.hpp
```diff
@@ -157,6 +157,8 @@ int sys_vm_unmap(void *pointer, size_t size);
 [[gnu::weak]] int sys_before_cancellable_syscall(ucontext_t *uctx);
 [[gnu::weak]] int sys_tgkill(int tgid, int tid, int sig);
 
+[[gnu::weak]] int sys_fchownat(int dirfd, const char *pathname, uid_t owner, gid_t group, int flags, bool dereference);
```
avdgrinten, line 160: Why do we have `bool dereference`? Following symlinks is just `AT_SYMLINK_NOFOLLOW` in flags, isn't it? See https://man7.org/linux/man-pages/man2/lchown.2.html

options/posix/generic/unistd-stubs.cpp
```diff
@@ -194,14 +201,29 @@ int faccessat(int dirfd, const char *pathname, int mode, int flags) {
...
 
 int fchownat(int fd, const char *path, uid_t uid, gid_t gid, int flags) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+	if(!mlibc::sys_fchownat) {
+		MLIBC_MISSING_SYSDEP();
+		errno = ENOSYS;
+		return -1;
+	}
+	if(int e = mlibc::sys_fchownat(fd, path, uid, gid, flags, true); e) {
```
avdgrinten, line 222: Here we need to pass `AT_EMPTY_PATH`, don't we?

options/posix/include/mlibc/posix-sysdeps.hpp
Dennisbonke: I didn't think of that, it was to distinguish `chown` and `lchown`. I'll change it to AT_SYMLINK_NOFOLLOW

options/posix/generic/unistd-stubs.cpp
Dennisbonke: I don't think we do, if the application wants that behavior, it should add `AT_EMPTY_PATH` to flags itself, right?

avdgrinten: Oh, I attached the comment in the wrong function. I meant `fchown`. fchown is basically `fchownat` with AT_EMPTY_PATH, isn't it?

Dennisbonke: Hmm, that seems possible yes, I'll change that.

options/posix/generic/unistd-stubs.cpp
```diff
@@ -46,8 +46,15 @@ int fchdir(int fd) {
...
 
 int chown(const char *path, uid_t uid, gid_t gid) {
-	mlibc::infoLogger() << "\e[31mmlibc: chown() is not implemented correctly\e[39m"
-			<< frg::endlog;
+	if(!mlibc::sys_fchownat) {
+		MLIBC_MISSING_SYSDEP();
+		errno = ENOSYS;
+		return -1;
+	}
+	if(int e = mlibc::sys_fchownat(-1, path, uid, gid, 0); e) {
```
avdgrinten, line 54: This `-1` needs to be `AT_FDCWD`, right?

options/posix/generic/unistd-stubs.cpp
```diff
@@ -194,14 +201,29 @@ int faccessat(int dirfd, const char *pathname, int mode, int flags) {
...
 
 int fchown(int fd, uid_t uid, gid_t gid) {
-	mlibc::infoLogger() << "\e[31mmlibc: fchown() is not implemented correctly\e[39m"
-			<< frg::endlog;
+	if(!mlibc::sys_fchownat) {
+		MLIBC_MISSING_SYSDEP();
+		errno = ENOSYS;
+		return -1;
+	}
+	if(int e = mlibc::sys_fchownat(fd, nullptr, uid, gid, AT_EMPTY_PATH); e) {
```
avdgrinten, line 209: The description states that the path name needs to be an empty string (and not NULL) for this to work.

options/posix/generic/unistd-stubs.cpp
```diff
@@ -379,8 +401,16 @@ pid_t getsid(pid_t pid) {
...
 
 int lchown(const char *path, uid_t uid, gid_t gid) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+	if(!mlibc::sys_fchownat) {
+		MLIBC_MISSING_SYSDEP();
+		errno = ENOSYS;
+		return -1;
+	}
+	if(int e = mlibc::sys_fchownat(-1, path, uid, gid, AT_SYMLINK_NOFOLLOW); e) {
```
avdgrinten, line 409: Likewise.

PR #251, opened 2021-04-15
sysdeps/managarm: Add Scrt1 for x86_64

Dennisbonke:


sysdeps/managarm/x86_64/crt-src/Scrt1.s
```diff
@@ -0,0 +1,5 @@
+.section .text
```
avdgrinten, line 1: The filename ends in `.s` (lower case) while it should be `.S`. This results in a build failure on my local machine.

sysdeps/managarm/x86_64/crt-src/Scrt1.s
```diff
@@ -0,0 +1,5 @@
...
+.global _start
+_start:
+	lea main(%rip), %rsi
```
avdgrinten, line 4: This needs to be `rdi` (see crt0.S, Managarm's `__mlibc_entry` takes different args than Linux' `__mlibc_entry`).

avdgrinten (approved):
LGTM now, thank you for the PR.

PR #260, opened 2021-04-25
sysdeps/managarm: Handle "dismiss" errors for ioctls

Dennisbonke:
This fixes X apps crashing on startup

### DIFF
--- a/sysdeps/managarm/generic/file.cpp
+++ b/sysdeps/managarm/generic/file.cpp
@@ -1756,6 +1756,8 @@ int sys_ioctl(int fd, unsigned long request, void *arg, int *result) {
 
 		HEL_CHECK(offer.error());
 		HEL_CHECK(sendReq.error());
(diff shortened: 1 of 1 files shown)


avdgrinten:
We should probably handle `kHelErrDismissed` for all ioctls.

sysdeps/managarm/generic/file.cpp
```diff
@@ -2412,6 +2436,8 @@ int sys_ioctl(int fd, unsigned long request, void *arg, int *result) {
 
 		HEL_CHECK(offer->error);
 		HEL_CHECK(send_req->error);
+		if(recv_resp->error == kHelErrDismissed)
+			return EINVAL;
 		HEL_CHECK(send_mode->error);
```
avdgrinten, line 2441: Looks like this should be `send_mode->error` (if that's the first item after the request)?

avdgrinten (approved):

PR #261, opened 2021-04-28
Small fix related to getrandom

Dennisbonke:
This PR moves the size check from the managarm sysdep (which is the only platform that implements `sys_getentropy`) to the actual `getrandom()` function.

### DIFF
--- a/options/linux/generic/sys-random-stubs.cpp
+++ b/options/linux/generic/sys-random-stubs.cpp
@@ -23,4 +23,3 @@ ssize_t getrandom(void *buffer, size_t max_size, unsigned int flags) {
 	}
 	return max_size;
 }
-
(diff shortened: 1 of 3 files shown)


options/linux/generic/sys-random-stubs.cpp
```diff
@@ -12,6 +12,10 @@ ssize_t getrandom(void *buffer, size_t max_size, unsigned int flags) {
 		errno = EINVAL;
 		return -1;
 	}
+	if(length > 256) {
+		errno = EIO;
+		return -1;
+	}
```
avdgrinten, line 18: It is `getentropy`, not `getrandom` that should fail in this case.

avdgrinten (approved):
LGTM.

PR #265, opened 2021-05-07
Update BUFSIZ macro to 4096 with stdio.h

expnkx:
512 is extremely small and it will lead to very poor io performance.
By the way, libstdc++ will use this macro to set the buffer size of std::fstream, leading to extremely poor performance.

Thank you!

### DIFF
--- a/options/ansi/include/stdio.h
+++ b/options/ansi/include/stdio.h
@@ -62,7 +62,7 @@ typedef size_t fpos_t;
 #define _IOLBF 2
 #define _IONBF 3
 
-#define BUFSIZ 512
+#define BUFSIZ 4096
 
 #define EOF (-1)
 


64:
Do you have a benchmark? There might be a better number than 4096

Dennisbonke:
Glibc sets the value of this define to 8192, so I’m inclined to say that while 4096 might be a better value then 512, it’s probably not the best one

Dennisbonke:
> > Glibc sets the value of this define to 8192, so I’m inclined to say that while 4096 might be a better value then 512, it’s probably not the best one. However, musl defines it as 1024, so I’m interested in which is better.
> 
> No. glibc sets the value by extracting blksize from fstat. That is not a constant.

[This](https://github.com/bminor/glibc/blob/3bf0b4f2cdc763141dd850be7588c32a720f5ce5/libio/stdio.h), which also matches my host, does suggest it's a hardcoded value.

expnkx:
> > > Glibc sets the value of this define to 8192, so I’m inclined to say that while 4096 might be a better value then 512, it’s probably not the best one. However, musl defines it as 1024, so I’m interested in which is better.
> > 
> > 
> > No. glibc sets the value by extracting blksize from fstat. That is not a constant.
> 
> [This](https://github.com/bminor/glibc/blob/3bf0b4f2cdc763141dd850be7588c32a720f5ce5/libio/stdio.h), which also matches my host, does suggest it's a hardcoded value.

musl sets the value for 1024 because it is designed for embedded systems where memory usage is critical.

I can guarantee it is not a hardcoded value. We even had a bug before in WSL2 since Microsoft's Hyper-v driver sets the value as 512.
The optimal BUFSIZ is actually 131072 which is far larger than 8192 or 4096.

expnkx:
> Let's not bike-shed over the exact value, that will depend a lot on the workload.

The problem is that libstdc++ will use BUFSIZ macro to set buffer size for fstream and that will lead to serious performance issues.

Although 4096 is nowhere an optimal value, it is not a bad value, you won't get too slow because of that value.

Dennisbonke:
This was implemented with 120afb1f6853681a66d8f39f61d2da5463ba9f1d and as such can be closed I think.

avdgrinten:
This PR is not quite the same as the recent commit; the commit changes mlibc's internal buffer size but BUFSIZ is a publicly exposed macro.

PR #267, opened 2021-05-09
rtdl: Permit TLS alignments greater than 16 bytes

64:
This is required for [`exa`](https://github.com/ogham/exa).


avdgrinten:
I think the alignment assertion was done because we do not take any measures to align the TLS allocation itself (in `allocateTcb()`).

Luckily, the slab allocator that we use will usually return naturally aligned blocks, so allocations > 32 bytes should be 32 bit aligned (in reality, it's a bit more complicated than that). If we add an assertion to `allocateTcb()` to check that the allocation is aligned to `tlsMaxAlignment`, this change is fine.

64:
`tlsMaxAlignment` is currently 16, so asserting what you say wouldn’t catch the case where the requested alignment is 32 (which is what I encountered with `exa`).

Should we bump `tlsMaxAlignment` to 32, or am I misunderstanding you?

qookei:
`tlsMaxAlignment` is set to be the highest alignment value we see of all objects that are in the initial set (see `Loader::submitObject`).

We should probably tell the allocator to use `tlsMaxAlignment` as the alignment for the whole TLS segment.
And while we're at it, we should fix dynamic TLS block having possibly too small alignment (since we don't tell the allocator what alignment to use when allocating them either, in `accessDtv`)

64:
Oh, right. I assumed it was a constant.

Okay, I'll have another look at this when I have some time.

avdgrinten:
The problem is that we cannot tell the allocator to use a non-natural alignment right now (the allocator doesn't support arbitrary alignments yet). Hence, I'd just suggest to add an assertion that the pointer is properly aligned; it is highly likely that it is.

64:
Like that?

avdgrinten (approved):
Yes, LGTM. Thanks for the PR!

PR #268, opened 2021-05-10
abis/mlibc: Change uid_t, gid_t and id_t to unsigned

64:
On Linux (and apparently everywhere else), these are unsigned types.


avdgrinten (approved):
LGTM.

Issue #185, opened 2020-11-22
getpwnam_r() unimplemented


Dennisbonke:
While testing on managarm the following program suite requires `getpwnam_r()`

- `shadow`

Issue #191, opened 2020-12-23
getpwuid_r() unimplemented


Dennisbonke:
On managarm, the following programs require `getpwuid_r()`:

- `irssi`, a irc client.

Geertiebear:
Closed by #269.

PR #269, opened 2021-05-15
options/posix: implement getpw{nam,uid}_r

64:
Fixes managarm/mlibc#185 and managarm/mlibc#191.

### DIFF
--- a/options/posix/generic/pwd-stubs.cpp
+++ b/options/posix/generic/pwd-stubs.cpp
@@ -11,9 +11,11 @@ namespace {
 	thread_local passwd global_entry;
 
 	bool extract_entry(frg::string_view line, passwd *entry) {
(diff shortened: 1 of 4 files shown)


Geertiebear:
LGTM although I wonder if there is a way we can auto test this?

64:
> LGTM although I wonder if there is a way we can auto test this?

I think it would be hard without some kind of container to test in (maybe that would be a good idea in the long term?)

Geertiebear:
Yeah, I don't think I see another way either.

avdgrinten:
I guess we could add a test that checks whether the root user exists or not.

64:
Ok, done

Geertiebear:
Then LGTM

PR #271, opened 2021-06-05
rtdl: Implement the case in dladdr where the symbol cannot be found

### LINKED ISSUE #211: ld.so: Find DSO even if dladdr() symbol is not found
The Linux man page specifies the following behavior of dladdr() if the given address does not match a symbol:

> On success, these functions return a nonzero value. If the address specified in addr could be matched to a shared object, but not to a symbol in the shared object, then the info->dli_sname and info->dli_saddr fields are set to NULL.

This behavior is not currently implemented in mlibc.

Dennisbonke:
This PR implements the missing case for when `dladdr()` can't find the symbol, and should return the DSO that it should be in. Required for `clang` to function.

(Partially) fixes #211

### DIFF
--- a/options/rtdl/generic/linker.hpp
+++ b/options/rtdl/generic/linker.hpp
@@ -167,6 +167,11 @@ struct SharedObject {
 	bool wasInitialized;
 
 	Scope *objectScope;
+
+	// PHDR related stuff, we only set these for the main executable
+	void *phdrPointer = nullptr;
+	size_t phdrEntrySize = 0;
+	size_t phdrCount = 0;
 };
 
 void processCopyRelocations(SharedObject *object);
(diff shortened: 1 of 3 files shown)


options/rtdl/generic/linker.hpp
```diff
@@ -167,6 +167,11 @@ struct SharedObject {
 	bool wasInitialized;
 
 	Scope *objectScope;
+
+	// PHDR related stuff
+	void *phdr_pointer = nullptr;
+	size_t phdr_entry_size = 0;
+	size_t phdr_count = 0;
```
avdgrinten, line 174: Please use camelCase for consistency with the other members.

avdgrinten: Add a comment that we currently only set these for the main executable. After the PR is merged, add an issue to add them in the general case.

avdgrinten (approved):
LGTM, thank you.

PR #272, opened 2021-06-05
DRM atomic modesetting

no92:


options/ansi/include/stdio.h
```diff
@@ -51,6 +52,10 @@ struct __mlibc_file_base {
 
 	// EOF and error bits.
 	int __status_bits;
+
+	// PID of the popen'ed process
+	pid_t __popen_pid;
+	int __popen_pipe;
```
avdgrinten, line 58: This is an ABI break so it needs to be coordinated with the next mlibc major release.

Geertiebear:
``popen()`` has been implemented by #285 already.

no92:
Rebased on current mlibc master. The updated changeset is sufficient to get weston up and running with atomic modesetting (kmscon doesn't use it apparently). The goal of supporting sway will be shifted to another PR, as this functionality is somewhat independent of sway alone. This set obviously depends on changes to DRM in managarm.

no92:
Removed the logging commit and added the `DRM_IOCTL_MODE_DESTROYBLOB` ioctl.

options/linux-headers/include/linux/vt.h
```diff
@@ -15,7 +15,7 @@
 
 #ifdef __cplusplus
 extern "C" {
-#endif __cplusplus
+#endif // __cplusplus
```
Geertiebear, line 18: This is not correct.

options/posix/generic/unistd-stubs.cpp
```diff
@@ -292,7 +292,20 @@ int ftruncate(int fd, off_t size) {
 }
 
 char *getcwd(char *buffer, size_t size) {
-	__ensure(buffer && "glibc extension for !buffer is not implemented");
+	/* In order to support glibc's extension of allocating buffer if none is given, we have this
```
Geertiebear, line 295: Maybe rename the commit for these changes, as ``getcwd`` was implemented before. Maybe we can also add a test that compares the output of ``getcwd`` with and without the buffer?

Geertiebear (changes requested):

options/linux-headers/include/linux/vt.h
no92: This throws warnings/errors if left untouched, so the comment is necessary. [Here's](https://gcc.gnu.org/onlinedocs/cpp/Ifdef.html) what gcc says about it.

Geertiebear: Oh sorry, I completely misread, yeah this is correct.

options/posix/generic/unistd-stubs.cpp
no92: Addressed in 86fade5.

sysdeps/managarm/generic/file.cpp
```diff
@@ -1142,7 +1142,11 @@ int sys_msg_send(int sockfd, const struct msghdr *hdr, int flags, ssize_t *lengt
 }
 
 int sys_msg_recv(int sockfd, struct msghdr *hdr, int flags, ssize_t *length) {
-	__ensure(hdr->msg_iovlen);
+	if(!hdr->msg_iovlen) {
+		errno = EMSGSIZE;
+		return -1;
```
Geertiebear, line 1147: We should return the error in a sysdep, not set errno.

no92: Addressed in 1dc288a.

Geertiebear (approved):
LGTM now!

PR #274, opened 2021-06-06
Various Linux fixes for OpenSSL

Dennisbonke:
This PR adds various headers and syscalls for the Linux port, enabling it to compile and run `OpenSSL` commands.

The following headers have been added:
- `memory.h`, installed when the ansi option is enabled,
- `sys/syscall.h`, a dummy header is installed when the linux option is enabled.

Furthermore, this PR implements the following syscalls for the Linux port:
- `sys_getuid()`,
- `sys_geteuid()`,
- `sys_getgid()`,
- `sys_getegid()`.

### DIFF
--- a/options/glibc/include/memory.h
+++ b/options/glibc/include/memory.h
@@ -1,0 +1,6 @@
+#ifndef _MEMORY_H
+#define _MEMORY_H
+
+#include <string.h>
+
+#endif
--- a/options/glibc/meson.build
+++ b/options/glibc/meson.build
@@ -18,7 +18,8 @@ if not no_headers
 		'include/resolv.h',
 		'include/endian.h',
 		'include/ar.h',
-		'include/shadow.h'
+		'include/shadow.h',
+		'include/memory.h'
 	)
 	install_headers(
 		'include/sys/ioctl.h',
--- a/options/linux/include/sys/syscall.h
+++ b/options/linux/include/sys/syscall.h
@@ -1,0 +1,6 @@
+#ifndef _SYS_SYSCALL_H
+#define _SYS_SYSCALL_H
+
+// OpenSSL wants this header to exist
+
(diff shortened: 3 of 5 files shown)


options/ansi/include/memory.h
```diff
@@ -0,0 +1,9 @@
+
+#ifndef _MEMORY_H
+#define _MEMORY_H
+
+#ifndef _STRING_H
```
Geertiebear, line 5: Why this? ``string.h`` already has include guards.

Dennisbonke: It matches with what glibc does, so I copied that.

options/ansi/include/memory.h
```diff
@@ -0,0 +1,9 @@
+
+#ifndef _MEMORY_H
+#define _MEMORY_H
```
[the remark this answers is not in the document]
Dennisbonke: You're right, and POSIX doesn't define it either. Do I put it under glibc or linux in this case?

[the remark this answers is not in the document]
avdgrinten: Let's put it into `glibc`.

options/ansi/include/memory.h
avdgrinten: It's not necessary though; let's remove it.

options/glibc/include/memory.h
```diff
@@ -0,0 +1,7 @@
+#ifndef _MEMORY_H
+#define _MEMORY_H
+
+#include <string.h>
+
+#endif
+
```
avdgrinten, line 7: Unnecessary empty line outside of the include guard.

avdgrinten (commented):
Can be merged after the nitpick is fixed.

avdgrinten:
LGTM now, thank you.

PR #275, opened 2021-06-06
options/posix: add AI_NUMERICHOST support to getaddrinfo()

Geertiebear:


tests/posix/getaddrinfo.c
```diff
@@ -28,5 +28,22 @@ int main() {
...
+
+	addr = (struct sockaddr_in*)res[0].ai_addr;
+	assert((addr->sin_addr.s_addr & 0xFF) == 10);
+	assert(((addr->sin_addr.s_addr >> 8) & 0xFF) == 10);
+	assert(((addr->sin_addr.s_addr >> 16) & 0xFF) == 10);
+	assert(((addr->sin_addr.s_addr >> 24) & 0xFF) == 10);
+
+	freeaddrinfo(res);
+
+	ret = getaddrinfo("AAAA", NULL, &hints, &res);
```
avdgrinten, line 43: What about picking an existing name (say, `example.org`) here?

options/posix/generic/netdb-stubs.cpp
```diff
@@ -83,16 +83,23 @@ int getaddrinfo(const char *__restrict node, const char *__restrict service,
...
 		if (addr_count < 0)
 			return -addr_count;
 		if (!addr_count)
 			return EAI_NONAME;
 	} else {
 		/* There is no node specified */
+		if (flags & AI_NUMERICHOST)
+			return EAI_NONAME;
 		addr_count = lookup_name_null(addr_buf, flags, family);
 	}
```
avdgrinten, line 104: What is this case good for?

options/posix/generic/lookup.cpp
```diff
@@ -448,4 +448,62 @@ int lookup_name_null(struct lookup_result &buf, int flags, int family) {
...
+		struct dns_addr_buf addr_buf;
+		addr_buf.family = AF_INET;
+		memcpy(&addr_buf.addr, &addr4, 4);
+
+		buf.buf.push_back(addr_buf);
+		return 1;
+	}
+
+	struct in6_addr addr6{0};
+	res = inet_pton(AF_INET, name, &addr4);
```
avdgrinten, line 496: This doesn't actually pass AF_INET6; `addr4` also looks wrong. Copy/paste mistake?

options/posix/generic/netdb-stubs.cpp
```diff
@@ -83,16 +83,23 @@ int getaddrinfo(const char *__restrict node, const char *__restrict service,
...
-			addr_count = mlibc::lookup_name_dns(addr_buf, node, canon);
-		else
-			addr_count = 1;
+		if ((addr_count = mlibc::lookup_name_ip(addr_buf, node, family)) <= 0) {
+			if (flags & AI_NUMERICHOST)
+			       addr_count = -EAI_NONAME;
+			else if ((addr_count = mlibc::lookup_name_hosts(addr_buf, node, canon)) <= 0)
+				addr_count = mlibc::lookup_name_dns(addr_buf, node, canon);
+			else
+				addr_count = 1;
```
avdgrinten, line 92: This `else` looks a bit strange (but it was also present before). If `lookup_name_hosts` succeeds, we discard the `addr_count` and use 1 instead?

options/posix/generic/netdb-stubs.cpp
Geertiebear: The spec says that if ``AI_NUMERICHOST`` is set then ``node`` _must_ contain a valid ip address. This also implies that a null node and ``AI_NUMERICHOST`` would be an error condition.

options/posix/generic/lookup.cpp
Geertiebear: Looks like it yes.

options/posix/generic/netdb-stubs.cpp
Geertiebear: This also puzzled me, I could check the source for ``mlibc::lookup_name_hosts()`` again and see why I did that in the first place.

Geertiebear: Okay, after reading the code again I _think_ it's because there can be multiple entries in ``/etc/hosts`` but we only want to take the first one. Although I'm really not sure about it.

avdgrinten (approved):
LGTM now, thank you for the PR!

PR #280, opened 2021-06-28
Updates to lemon sysdeps

fido2020:


avdgrinten (approved):
LGTM (in general, you can merge PRs that only affect `sysdeps/lemon` without an external review if you like to).

PR #285, opened 2021-07-14
posix: Implement popen()

mintsuki:


options/posix/generic/posix_stdio.cpp
```diff
@@ -18,9 +24,103 @@ int pclose(FILE *) {
...
+
+	if (!mlibc::sys_fork || !mlibc::sys_close || !mlibc::sys_dup2 || !mlibc::sys_execve
+			|| !mlibc::sys_sigprocmask || !mlibc::sys_sigaction || !mlibc::sys_pipe) {
+		MLIBC_MISSING_SYSDEP();
+		errno = ENOSYS;
+		return nullptr;
+	}
+
+#if __MLIBC_POSIX_OPTION
+	pthread_testcancel();
```
Geertiebear, line 40: I don't see where ``popen()`` is specified as a cancellation point in POSIX.

Geertiebear (changes requested):
Is it possible to also make a small test case for ``popen()``? Something like executing ``echo "test"`` or something like that and checking if the correct output is there?

options/posix/generic/posix_stdio.cpp
Dennisbonke: [This[(https://pubs.opengroup.org/onlinepubs/000095399/functions/xsh_chap02_09.html#tag_02_09_05_02) says that it _may_ be a cancellation point so I think that is correct.

Geertiebear: Ah I completely read over that, my bad. I'm still not sure we should include ``pthread_testcancel()`` however, since POSIX says it _can_ be a cancellation point to cover implementations that use blocking syscalls to implement it. In general library functions don't have to explicitely test for cancellations, since the sysdeps should already check for it appropriately.

E.g. nowhere in the ``fputc()`` implementation do we check using ``pthread_testcancel()``, but since it calls into ``read()``, it automatically becomes a cancellation point.

options/posix/generic/posix_stdio.cpp
```diff
@@ -2,11 +2,17 @@
...
 #include <stdlib.h>
+#include <unistd.h>
 
 #include <bits/ensure.h>
 #include <mlibc/debug.hpp>
 #include <mlibc/file-io.hpp>
 #include <mlibc/posix-file-io.hpp>
+#include <mlibc/posix-sysdeps.hpp>
+
+#if __MLIBC_POSIX_OPTION
```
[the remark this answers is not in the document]
mintsuki: The `#if`? Because the include did not seem unnecessary.

options/posix/generic/posix_stdio.cpp
```diff
@@ -18,9 +24,103 @@ int pclose(FILE *) {
...
 
-FILE *popen(const char*, const char *) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+FILE *popen(const char *command, const char *typestr) {
+	bool is_write;
+	pid_t child;
+	FILE *ret = nullptr;
+
+	if (!mlibc::sys_fork || !mlibc::sys_close || !mlibc::sys_dup2 || !mlibc::sys_execve
```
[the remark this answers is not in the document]
mintsuki: I just did what `system()` already did.

options/posix/generic/posix_stdio.cpp
```diff
@@ -18,9 +24,103 @@ int pclose(FILE *) {
...
+		errno = ENOSYS;
+		return nullptr;
+	}
+
+#if __MLIBC_POSIX_OPTION
+	pthread_testcancel();
+#endif // __MLIBC_POSIX_OPTION
+
+	if (!command) {
+		return nullptr;
```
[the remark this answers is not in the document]
mintsuki: I did what `system()` did.

options/posix/generic/posix_stdio.cpp
mintsuki: Again, I did what `system()` did :^)

options/posix/generic/posix_stdio.cpp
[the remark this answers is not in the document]
Geertiebear: Yeah but you don't need to do the ``#if`` if we are already in the POSIX option.

options/posix/generic/posix_stdio.cpp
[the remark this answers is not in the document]
Geertiebear: Guess that's fine then.

options/posix/generic/posix_stdio.cpp
Geertiebear: ``system()`` has the requirement that it _has_ to be a cancellation point, ``popen()`` doens't have this requirement. So imo it's not needed and can be removed.

options/posix/generic/posix_stdio.cpp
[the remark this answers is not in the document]
Geertiebear: In the man pages for ``system()`` it explicitely mentions the case ``command == NULL`` but the man page for ``popen()`` doesn't, so I don't think any special handling needs to be done for this case (although maybe a test case to verify it doesn't cause a null-deref would be good).

mintsuki:
Done.

options/posix/generic/posix_stdio.cpp
```diff
@@ -18,9 +20,98 @@ int pclose(FILE *) {
...
+	FILE *ret = nullptr;
+
+	if (!mlibc::sys_fork || !mlibc::sys_close || !mlibc::sys_dup2 || !mlibc::sys_execve
+			|| !mlibc::sys_sigprocmask || !mlibc::sys_sigaction || !mlibc::sys_pipe) {
+		MLIBC_MISSING_SYSDEP();
+		errno = ENOSYS;
+		return nullptr;
+	}
+
+	__ensure(command);
```
Geertiebear, line 35: But now we're still checking for ``command``

options/posix/generic/posix_stdio.cpp
```diff
@@ -18,9 +20,98 @@ int pclose(FILE *) {
...
+    __ensure(typestr);
```
Geertiebear, line 36: A ``nullptr`` ``typestr`` should cause ``errno = EINVAL``.

Geertiebear (commented):
Two more comments, for the rest LGTM.

Geertiebear (approved):
LGTM now!

PR #287, opened 2021-07-14
options/posix: fix pthread syscall cancellation

Geertiebear:


avdgrinten:
What kind of issue is fixed by this PR?

Geertiebear:
This is something that was supposed to work with #230, but the behaviour wasn't implemented correctly in that PR. This PR fixes that.

avdgrinten:
Okay, but what exactly does it fix? What aspect of the feature was broken before?

Geertiebear:
Deferred cancellation never interrupted blocking syscalls properly. Only asynchronous did. This PR fixes that issue, and adds a small test case for it.

avdgrinten (approved):
LGTM.

PR #292, opened 2021-07-23
options/posix: Add several _POSIX feature defines

Dennisbonke:
This PR adds several `_POSIX_*` feature defines, which I believe mlibc supports. `_POSIX_THREAD_SAFE_FUNCTIONS` is required for `git` to compile for Managarm. It is possible that one or several of these are not actually supported, in which case we'll need to remove them or implement the missing functionality.


avdgrinten:
Why are some macros defined as `1` and others as `_POSIX_VERSION`?

Dennisbonke:
This is in accordance with musl, and seems to match the posix standard.

Geertiebear:
If it's in accordance with musl and POSIX then LGTM, let's get this merged.

PR #294, opened 2021-07-31
options/posix: Include sys/select.h in sys/time.h, include timespec i…

fido2020:
The POSIX spec states that time.h should include select function and various macros from sys/select.h. It also states that sys/select.h should define timespec. This is needed for zsh to compile as it tests for timespec in sys/time.h.


Geertiebear (approved):
LGTM

PR #295, opened 2021-08-01
Various fixes for `exa`

64:
- options/posix: fill pw_passwd in getpw{nam,uid}
- options/posix: stub pthread_attr_init, enable pthread_cond_{wait, broadcast}

image

After this PR, it just crashes on `pthread_detach`.


Geertiebear:
Can we add tests for both the getpw and pthread_cond fixes?

64:
I don't know how we can properly test the pwd change. Can we rely on the value of root pw_passwd being consistent across distros? If not maybe we can just add a check that the field is a valid c string

Geertiebear:
Ig just sanity checking that the data is not uninitialised should be enough.

Geertiebear (approved):
LGTM

PR #296, opened 2021-08-02
 options/ansi: Return SIG_ERR instead of panic when sys_sigaction fails in signal()

fido2020:
This is in line with the POSIX spec.
Programs such as zsh may pass an invalid signal number and crash on EINVAL.

Also updated lemon sysdeps.


Geertiebear:
LGTM

PR #298, opened 2021-08-11
Improvements for QEMU

qookei:


options/posix/generic/posix_signal.cpp
```diff
@@ -22,8 +30,18 @@ int pthread_sigmask(int how, const sigset_t *__restrict set, sigset_t *__restric
...
-	__ensure(!"pthread_kill() not implemented");
-	__builtin_unreachable();
+	auto tcb = reinterpret_cast<Tcb *>(thread);
+
+	if(!mlibc::sys_kill) {
+		MLIBC_MISSING_SYSDEP();
+		return ENOSYS;
+	}
+
+	if(int e = mlibc::sys_kill(tcb->tid, sig); e) {
```
Geertiebear, line 40: Shouldn't this be a ``sys_tgkill``?

Geertiebear (commented):
Some tests would be good to have, for example for the new pthread features, and maybe also the signal features if it isn't too much work.

options/posix/generic/pthread-stubs.cpp
```diff
@@ -136,8 +136,59 @@ int pthread_equal(pthread_t t1, pthread_t t2) {
...
+
+	auto hand = self->cleanupEnd;
+	while (hand) {
+		auto old = hand;
+		hand->func(hand->arg);
+		hand = hand->prev;
+		frg::destruct(getAllocator(), old);
+	}
+
+	// FIXME: This is a hack used to avoid doing syscalls when being cancelled, as that
```
Geertiebear, line 176: This can be fixed by setting the tcbExitingBit when beginning the exit procedure.

Geertiebear (commented):
LGTM besides the one comment.

options/posix/generic/pthread-stubs.cpp
```diff
@@ -136,8 +136,56 @@ int pthread_equal(pthread_t t1, pthread_t t2) {
...
+	> key_locals_{};
+
+	FutexLock key_mutex_;
+}
+
 int pthread_exit(void *ret_val) {
-	auto self = static_cast<Tcb *>(mlibc::get_current_tcb());
+	auto self = mlibc::get_current_tcb();
+
+	__atomic_fetch_or(&self->cancelBits, tcbExitingBit, __ATOMIC_RELAXED);
```
Geertiebear, line 168: Shouldn't we check before setting this if it is already set?

qookei: Does that matter? AFAIU if it's already set nothing will happen when we set it here.

PR #299, opened 2021-08-12
options/ansi: Add handling for mode 'a' and 'a+' to fopen

Dennisbonke:
This is required for `ninja` to work on Managarm.

### DIFF
--- a/options/ansi/generic/file-io.cpp
+++ b/options/ansi/generic/file-io.cpp
@@ -523,6 +523,13 @@ FILE *fopen(const char *path, const char *mode) {
 			flags = __MLIBC_O_WRONLY;
 		}
 		flags |= __MLIBC_O_CREAT | __MLIBC_O_TRUNC;
+	}else if(*mode == 'a') {
+		if(has_plus) {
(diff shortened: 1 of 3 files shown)


Geertiebear:
Would be nice if this also included a testcase for ``fopen`` but im not sure how much work that is.

sysdeps/linux/generic/sysdeps.cpp
```diff
@@ -271,21 +271,21 @@ int sys_sleep(time_t *secs, long *nanos) {
 	return 0;
 }
 
+int sys_isatty(int fd) {
+	struct winsize ws;
+	auto ret = do_syscall(NR_ioctl, fd, TIOCGWINSZ, &ws);
```
Geertiebear, line 276: This doesn't work. It relies on ``ioctl()`` numbers which are not available in ansi-only. But doing this isn't necessary, your test shouldn't be failing because of ``sys_isatty()`` missing, and if it is then that is a separate issue that needs to be addressed properly.

64:
This can be closed in favour of #446.

PR #300, opened 2021-08-13
Warning fixes

Dennisbonke:
This PR aims to fix all warnings in the directories listed below, either by fixing or ignoring the warning in special cases. Furthermore, it adds `-Wall` and `-Wextra` to the global project flags.

Directories checked in this PR:
- `options/ansi`,
- `options/elf`,
- `options/glibc`,
- `options/internal`,
- `options/linux`,
- `options/lsb`,
- `options/posix`,
- `options/rtdl`,
- `sysdeps/managarm`.

### DIFF
--- a/options/ansi/include/wchar.h
+++ b/options/ansi/include/wchar.h
@@ -9,13 +9,11 @@
 #include <bits/null.h>
 #include <bits/size_t.h>
 #include <bits/wchar_t.h>
+#include <bits/wchar.h>
 #include <bits/wint_t.h>
 #include <bits/mbstate.h>
 
 #define WEOF 0xffffffffU
-// TODO: The following declaration should be independent of gcc.
-#define WCHAR_MIN __WCHAR_MIN__
-#define WCHAR_MAX __WCHAR_MAX__
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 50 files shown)


options/posix/generic/lookup.cpp
```diff
@@ -192,7 +192,7 @@ int lookup_addr_dns(frg::span<char> name, frg::array<uint8_t, 16> &addr, int fam
 	size_t ptr = 0;
 	do {
 		size_t next = req_view.find_first('.', ptr);
-		size_t length = next != -1 ? next - ptr : req_view.size() - ptr;
+		size_t length = next != (size_t)-1 ? next - ptr : req_view.size() - ptr;
```
64, line 196: how about `SIZE_MAX`?

64:
LGTM

options/ansi/include/wchar.h
```diff
@@ -14,8 +14,12 @@
 
 #define WEOF 0xffffffffU
 // TODO: The following declaration should be independent of gcc.
+#ifndef WCHAR_MIN
```
Geertiebear, line 17: Why do these actually get re-defined? Is there some place else we define them?

Geertiebear (commented):
LGTM besides the comment

Geertiebear (changes requested):

Geertiebear (approved):
LGTM now.

PR #301, opened 2021-08-15
posix: Implement and stub more functions for shadow to function

Dennisbonke:
This PR stubs `initgroups()` and implements `execle()` like musl. With this, I can run `su --login managarm` on Managarm to become the managarm user.


Geertiebear (approved):
LGTM, depending on how hard a test case is to add that would be very nice.

PR #309, opened 2021-08-30
options/posix: implement getservbyport()

### LINKED ISSUE #289: getservbyport() unimplemented
The following program, tested on managarm, require `getservbyport()` to properly function

- `nc`, without it, `nc` has no working features.

Geertiebear:
Fixes #289.

### DIFF
--- a/options/posix/generic/services.cpp
+++ b/options/posix/generic/services.cpp
@@ -11,7 +11,90 @@
 
 namespace mlibc {
 
-static int lookup_serv_file(struct service_buf *buf, const char *name,
+static int parse_rest(service_buf &buf, char *end, int proto) {
+	if (!strncmp(end, "/udp", 4)) {
+		if (proto == IPPROTO_TCP && proto != -1)
+			return 0;
+		buf.protocol = IPPROTO_UDP;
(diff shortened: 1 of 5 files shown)


options/posix/generic/services.cpp
```diff
@@ -11,7 +11,88 @@
...
+			int rport = strtoul(pos, &end, 10);
+			if (rport != port || rport > 65535) {
+				pos = end;
+				continue;
+			}
+
+			// We have found the port, time to rewind to the start
+			// of the line.
+			for (; pos[-1]; pos--)
+				name_length++;
```
qookei, line 70: Is the character before the line buffer guaranteed to be 0?

Geertiebear: Good point, it actually isn't, but I can fix that real quick. Is the rest okay?

qookei: The `name_length` calculation seems like it includes the whitespace between the name and port as well, is that expected?

Geertiebear: Hm that's true, I guess that isn't expected. I'll add a check for whitespace too.

qookei (approved):
LGTM

PR #310, opened 2021-09-03
options/posix: Stub getloadavg

Dennisbonke:
This is required for `ninja` to compile. I thought this was already upstreamed in a different PR but apparently I was wrong.


options/posix/include/bits/posix/posix_stdlib.h
```diff
@@ -47,6 +47,8 @@ double strtod_l(const char *__restrict__ nptr, char ** __restrict__ endptr, loca
 long double strtold_l(const char *__restrict__ nptr, char ** __restrict__ endptr, locale_t loc);
 float strtof_l(const char *__restrict string, char **__restrict end, locale_t loc);
 
+int getloadavg(double [], int);
```
avdgrinten, line 50: I think we should declare this as `double *` (that's exactly the same thing and less confusing - in function declarations, arrays decay to pointers anyway).

avdgrinten:
LGTM.

PR #311, opened 2021-09-04
All pthread_attr_* support

### LINKED ISSUE #253: posix: Implement pthread_attr_*
These are used by the Rust standard library when spawning a thread, specifically:
* `pthread_attr_init`
* `pthread_attr_setstacksize`
* `pthread_attr_destroy`

Geertiebear:
Fixes #253. Is an ABI break.

### DIFF
--- a/ABI_BREAKS.md
+++ b/ABI_BREAKS.md
@@ -7,3 +7,4 @@ This document lists the ABI breaks that were made in each mlibc major version.
 - [#452](https://github.com/managarm/mlibc/pull/452): The functions `FD_{CLR,ISSET,SET,ZERO}` were renamed to `__FD_{CLR,ISSET,SET,ZERO}` and replaced by macros to match Wine's assumptions.
 - [#511](https://github.com/managarm/mlibc/pull/511): Musl's regex engine was added, implementing `regcomp` and `regexec`. This required some changes to the `regex_t` struct.
 - [#504](https://github.com/managarm/mlibc/pull/504), [#580](https://github.com/managarm/mlibc/pull/580): In both the mlibc and Linux ABIs, a `domainname` member was added to `struct utsname`, which is a glibc extension.
+- [#311](https://github.com/managarm/mlibc/pull/311): Added all necessary fields in `pthread_attr_t` required for implementing all `pthread_attr` functions.
--- a/options/ansi/generic/stdio-stubs.cpp
+++ b/options/ansi/generic/stdio-stubs.cpp
@@ -487,10 +487,10 @@ static int do_scanf(H &handler, const char *fmt, __gnuc_va_list args) {
                                 res = res * 16 + (c - '0');
                             } else if (c >= 'a' && c <= 'f') {
                                 handler.consume();
-                                res = res * 16 + (c - 'a');
+                                res = res * 16 + (c - 'a' + 10);
                             } else if (c >= 'A' && c <= 'F') {
                                 handler.consume();
-                                res = res * 16 + (c - 'A');
+                                res = res * 16 + (c - 'A' + 10);
                             } else {
                                 break;
                             }
@@ -539,10 +539,10 @@ static int do_scanf(H &handler, const char *fmt, __gnuc_va_list args) {
                         res = res * 16 + (c - '0');
                     } else if (c >= 'a' && c <= 'f') {
                         handler.consume();
-                        res = res * 16 + (c - 'a');
+                        res = res * 16 + (c - 'a' + 10);
                     } else if (c >= 'A' && c <= 'F') {
                         handler.consume();
-                        res = res * 16 + (c - 'A');
+                        res = res * 16 + (c - 'A' + 10);
                     } else {
                         break;
                     }
--- a/options/internal/aarch64-include/mlibc/thread.hpp
+++ b/options/internal/aarch64-include/mlibc/thread.hpp
@@ -11,4 +11,10 @@ inline Tcb *get_current_tcb() {
 	return reinterpret_cast<Tcb *>(ptr);
 }
 
+inline uintptr_t get_sp() {
+	uintptr_t sp;
+	asm ("mov %0, sp" : "=r"(sp));
+	return sp;
+}
+
 } // namespace mlibc
--- a/options/posix/generic/pthread-stubs.cpp
+++ b/options/posix/generic/pthread-stubs.cpp
@@ -109,6 +109,19 @@ int pthread_attr_setstackaddr(pthread_attr_t *attr, void *stackaddr) {
 	return 0;
 }
 
+int pthread_attr_getstack(const pthread_attr_t *attr, void **stackaddr, size_t *stacksize) {
+	*stackaddr = attr->__mlibc_stackaddr;
+	*stacksize = attr->__mlibc_stacksize;
+	return 0;
+}
+int pthread_attr_setstack(pthread_attr_t *attr, void *stackaddr, size_t stacksize) {
+	if (stacksize < PTHREAD_STACK_MIN)
+		return EINVAL;
+	attr->__mlibc_stacksize = stacksize;
+	attr->__mlibc_stackaddr = stackaddr;
+	return 0;
+}
+
 int pthread_attr_getguardsize(const pthread_attr_t *__restrict attr, size_t *__restrict guardsize) {
 	*guardsize = attr->__mlibc_guardsize;
 	return 0;
@@ -342,16 +355,6 @@ int pthread_setname_np(pthread_t, const char *) {
 }
 
 int pthread_getname_np(pthread_t, char *, size_t) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
-}
-
-int pthread_attr_setstack(pthread_attr_t *, void *, size_t) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
-}
-
-int pthread_attr_getstack(const pthread_attr_t *, void **, size_t *) {
 	__ensure(!"Not implemented");
 	__builtin_unreachable();
 }
--- a/options/posix/generic/pthread-stubs.cpp
+++ b/options/posix/generic/pthread-stubs.cpp
@@ -68,14 +68,17 @@ int pthread_attr_destroy(pthread_attr_t *) {
 	return 0;
 }
 
-int pthread_attr_getdetachstate(const pthread_attr_t *, int *) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+int pthread_attr_getdetachstate(const pthread_attr_t *attr, int *detachstate) {
+	*detachstate = attr->__mlibc_detachstate;
+	return 0;
 }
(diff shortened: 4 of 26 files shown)


Dennisbonke:
uhm, 3 commit messages are wrong, I don't know `posix_attr_{get,set}stackaddr()`, I assume you mean `pthread_` there?

Geertiebear:
Oh, good one, I'll fix those.

avdgrinten:
Are there any attributes supported by musl/glibc which would require additional ABI breaks?

Geertiebear:
Just one, which is ``pthread_attr_setaffinity_np``, which needs a ``cpu_set_t`` in the attr struct.

avdgrinten:
Shouldn't we add this now to prevent future ABI breaks?

avdgrinten:
We could protect against future breaks by including a version number in the structs. Should we do that?

Geertiebear:
> Shouldn't we add this now to prevent future ABI breaks?

Yeah, taking a second look at it it shouldn't be very difficult. So it should be added in this PR.

> We could protect against future breaks by including a version number in the structs. Should we do that?

How does that work exactly?

ArsenArsen:
Alias the functions to something like `_Pthread_getattr_v2` which returns a pointer to `__Pthread_attrs_v2`, which is aliased to `pthread_attr`, etc. Since these get dealiased at compile/link time, as long as all the versions work, all programs will.

avdgrinten:
Aliasing the functions doesn't necessarily help, it's the size of the struct that needs to be handled.

We could add a version field to the struct (e.g., as first member) and let the initialization macros/functions set it to `1`. If we the attributes change in an incompatible way, we can bump the value.

options/posix/include/mlibc/posix-sysdeps.hpp
```diff
@@ -89,7 +89,8 @@ int sys_close(int fd);
 [[gnu::weak]] void sys_yield();
 [[gnu::weak]] int sys_sleep(time_t *secs, long *nanos);
 [[gnu::weak]] int sys_fork(pid_t *child);
-[[gnu::weak]] int sys_clone(void *entry, void *user_arg, void *tcb, pid_t *pid_out);
+[[gnu::weak]] int sys_clone(void *tcb, pid_t *pid_out, void *stack);
+[[gnu::weak]] int sys_prepare_stack(void **stack, void *entry, void *user_arg, size_t stack_size, size_t guard_size);
```
Dennisbonke, line 93: This definition does not match the definitions in the sysdeps, where an extra `void *tcb` can be found as 4th argument. Only x86_64 linux has a matching definition at this moment.

Geertiebear: Ah, you're right. That's an easy fix tho.

options/posix/generic/pthread-stubs.cpp
```diff
@@ -213,6 +227,9 @@ int pthread_create(pthread_t *__restrict thread, const pthread_attr_t *__restric
 		MLIBC_MISSING_SYSDEP();
 		return ENOSYS;
 	}
+	new_tcb->stackSize = attr.__mlibc_stacksize;
+	new_tcb->guardSize = attr.__mlibc_guardsize;
+	new_tcb->stackAddr = attr.__mlibc_stackaddr;
```
avdgrinten, line 232: Instead of plainly storing the stack size and address from the attribute, we should store the true values. The current code won't work if no explicit stack address is given in the attribute, will it?

Geertiebear: You're right, it won't.

Geertiebear: Fixed.

options/ansi/generic/stdio-stubs.cpp
```diff
@@ -495,12 +495,12 @@ static int do_scanf(H &handler, const char *fmt, __gnuc_va_list args) {
                             if (c >= '0' && c <= '9') {
                                 handler.consume();
                                 res = res * 16 + (c - '0');
-                            } else if (c >= 'a' && c <= 'f') {
+                            } else if (c >= 'a' && c <= 'f' + 10) {
                                 handler.consume();
                                 res = res * 16 + (c - 'a');
-                            } else if (c >= 'A' && c <= 'F') {
+                            } else if (c >= 'A' && c <= 'F' + 10) {
```
qookei, line 501: These +10 are in the wrong spot?

Geertiebear: Oh yeah, they should be in the actual addition right?

qookei: Yeah

options/posix/generic/pthread-stubs.cpp
```diff
@@ -63,61 +65,207 @@ static constexpr unsigned int mutex_excl_bit = static_cast<uint32_t>(1) << 30;
...
+		}
+
+		char line[256];
+		auto sp = mlibc::get_sp();
+		while (fgets(line, 128, fp)) {
+			uintptr_t from, to;
+			if(sscanf(line, "%lx-%lx", &from, &to) != 2)
+				continue;
+			if (sp < to && sp > from) {
+				*stack_addr = reinterpret_cast<void*>(to);
```
Geertiebear, line 207: This should be ``from``. The man pages state ``pthread_attr_getstack()`` should return the lowest addressable byte of the range, not the highest one.

Dennisbonke (changes requested):
Looks good, but merging this as is would break Managarm's aarch64 port.

ABI_BREAKS.md
```diff
@@ -6,3 +6,4 @@ This document lists the ABI breaks that were made in each mlibc major version.
 
 - [#452](https://github.com/managarm/mlibc/pull/452): The functions `FD_{CLR,ISSET,SET,ZERO}` were renamed to `__FD_{CLR,ISSET,SET,ZERO}` and replaced by macros to match Wine's assumptions.
 - [#511](https://github.com/managarm/mlibc/pull/511): Musl's regex engine was added, implementing `regcomp` and `regexec`. This required some changes to the `regex_t` struct.
+- [#311](https://github.com/managarm/mlibc/pull/311): Added all necessary fieldsin `pthread_attr_t` required for implementing all `pthread_attr` functions.
```
64, line 9: `fieldsin` -> `fields in`

options/posix/generic/pthread-stubs.cpp
```diff
@@ -63,97 +66,296 @@ static constexpr unsigned int mutex_excl_bit = static_cast<uint32_t>(1) << 30;
...
+		auto fp = fopen("/proc/self/maps", "r");
+		if (!fp) {
+			mlibc::infoLogger() << "mlibc pthreads: /proc/self/maps does not exist! Producing incorrect"
+				" stack results!" << frg::endlog;
+			return;
+		}
+
+		char line[256];
+		auto sp = mlibc::get_sp();
+		while (fgets(line, 128, fp)) {
```
64, line 284: `line` holds 256 bytes but you only grab 128 here. A constexpr variable should probably be used here.

options/posix/generic/pthread-stubs.cpp
```diff
@@ -63,97 +66,296 @@ static constexpr unsigned int mutex_excl_bit = static_cast<uint32_t>(1) << 30;
...
+
+		fclose(fp);
+	}
+}
+
+int pthread_getattr_np(pthread_t thread, pthread_attr_t *attr) {
+	auto tcb = reinterpret_cast<Tcb*>(thread);
+	*attr = pthread_attr_t{};
+
+	if (!tcb->stackAddr || !tcb->stackSize) {
```
64, line 305: Under what circumstances will these be zero? Aren't they set in pthread_create?

sysdeps/linux/generic/thread.cpp
```diff
@@ -31,20 +27,34 @@ extern "C" void __mlibc_enter_thread(void *entry, void *user_arg) {
...
-			PROT_READ | PROT_WRITE);
-	__ensure(!ret);
-
-	auto sp = reinterpret_cast<uintptr_t*>(map + STACK_SIZE + GUARD_SIZE);
+static constexpr size_t default_stacksize = 0x200000;
+
+int sys_prepare_stack(void **stack, void *entry, void *user_arg, void *tcb, size_t *stack_size, size_t *guard_size) {
+	(void)tcb;
+	if (!*stack_size)
+		*stack_size = default_stacksize;
```
64, line 35: It'd be nice to use camelCasd in these functions, but maybe we should leave it to a future PR to fix that.

tests/posix/pthread_attr.c
```diff
@@ -0,0 +1,212 @@
...
+
+static void *stackaddr_worker(void *arg) {
+	void *addr = *(void**)arg;
+
+	void *sp;
+#if defined(__x86_64__)
+	asm volatile ("mov %%rsp, %0" : "=r"(sp));
+#elif defined(__aarch64__)
+	asm volatile ("mov %0, sp" : "=r"(sp));
+#elif defined (__riscv) && __riscv_xlen == 64
```
64, line 94: The __riscv_xlen check isn't needed here (that instruction will work for 32-bit too).

options/posix/generic/pthread-stubs.cpp
Geertiebear: On the main thread, which hasn't been spawned with ``pthread_create()``.

64: Ah true. I guess it makes more sense to do that lazily then (as you do), since it's unlikely most programs will need that information.

options/posix/generic/pthread-stubs.cpp
```diff
@@ -151,14 +163,16 @@ int pthread_attr_setschedparam(pthread_attr_t *__restrict, const struct sched_pa
 	__builtin_unreachable();
 }
 
-int pthread_attr_getinheritsched(const pthread_attr_t *__restrict, int *__restrict) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+int pthread_attr_getinheritsched(const pthread_attr_t *attr, int *inheritsched) {
```
Dennisbonke: Seems to be doubly implemented? Same for `pthread_attr_setinheritsched`

64:
I've force-pushed the ABI break branch, so you will need to rebase this.

options/internal/aarch64-include/mlibc/thread.hpp
```diff
@@ -11,4 +11,10 @@ inline Tcb *get_current_tcb() {
 	return reinterpret_cast<Tcb *>(ptr);
 }
 
+inline uintptr_t get_sp() {
+	uintptr_t sp;
+	asm ("mv %0, sp" : "=r"(sp));
+	return sp;
```
qookei, line 17: This is wrong, the instruction should be `mov %0, sp`

qookei:
Why does this pull request also include regex stuff etc?

64 (approved):

qookei (approved):

PR #317, opened 2021-09-26
Add several stubs and implement `getdtablesize()`

ikbenlike:
Does what it says on the tin. However, `getdtablesize()` does only return the arbitrary value that sysconf returns.


Dennisbonke:
LGTM

PR #318, opened 2021-09-28
posix: add mising define RLIM_NLIMITS

ikbenlike:
Does exactly that. Feel a bit silly doing a PR for just this but I don't need to change anything else and you'll need to have it at some point.


Geertiebear (approved):
LGTM

PR #320, opened 2021-10-03
Make openat() return error code instead of fail on assert.

ikbenlike:
This change accompanies multiple changes in the POSIX subsystem, at the moment primarily to facilitate running zsh on managarm. 

The companion of this PR is [managarm PR #397.](https://github.com/managarm/managarm/pull/397)


Geertiebear:
The commit history is a bit messy. Do you mind cleaning it up a bit? This should be possible with ``git rebase``. Furthermore, it would be nice to have some simple test cases for ``wmemcmp()``, since it's a relatively easy function to unit test.

ikbenlike:
> The commit history is a bit messy. Do you mind cleaning it up a bit? This should be possible with `git rebase`. Furthermore, it would be nice to have some simple test cases for `wmemcmp()`, since it's a relatively easy function to unit test.

Oh yeah, I wanted to fix it up a bit yesterday but I forgot. I will add a test as soon as I have time.

ikbenlike:
That is a fair point. I fixed the meson thing so if that all works fine I suppose this PR is good for merging

PR #329, opened 2021-11-02
meta: add a release procedure

Geertiebear:
Open to suggestions and comments here for sure.

### DIFF
--- a/RELEASE_PROCEDURE.md
+++ b/RELEASE_PROCEDURE.md
@@ -1,0 +1,18 @@
+# Versioning
+
+mlibc uses semantic versioning to denote releases.
+A bump in the major version signals an ABI break.
+A bump in the minor version signals a newly released set of features, while still maintaining ABI.
+A bump in the patch signals a bug-fix, while not adding new features.
+
+# Release schedule
+
+A minor or major version will be released every two months (every even month). A release is turned into a major one
+when there is a pending ABI break, given that the ABI hasn't been broken in the last 6 months.
+A patch version will be released on-demand, as bugs are fixed.
+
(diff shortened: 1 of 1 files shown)


RELEASE_PROCEDURE.md
```diff
@@ -0,0 +1,18 @@
+# Versioning
+
+Mlibc uses semantic versioning to denote releases.
```
ArsenArsen, line 3: don't we stylize as `mlibc` (lowercase)?

ArsenArsen (approved):
lgtm, generally, the only discussion that one might want to have is about the density of the minor releases

RELEASE_PROCEDURE.md
Geertiebear: Mhm yeah we could

RELEASE_PROCEDURE.md
```diff
@@ -0,0 +1,18 @@
...
+A bump in the major version signals an ABI break.
+A bump in the minor version signals a newly released set of features, while still maintaining ABI.
+A bump in the patch signals a bug-fix, while not adding new features.
+
+# Release schedule
+
+A major version will be released irregularly, as it is only necessary when breaking ABI.
+A minor version will be released bi-monthtly (read: every two months).
```
avdgrinten, line 11: It's turn that into: A minor or major version will be released every two months. We can turn the release into a major version if there is a pending ABI break. Additionally, we can promise not to break ABI more than once every 6 months or so.

We should also specify in *which* months we do the release and whether the releases are scheduled at the beginning or end of these months.

Geertiebear: Fixed.

PR #330, opened 2021-11-04
Start work on systemd compilation support

Dennisbonke:
This PR adds initial work to get `systemd` compiling on `Managarm`. For now, this PR is a draft, but currently it adds a bunch of constants and missing headers, 1 new struct and 1 function definition. As more is coming, this PR remains in draft, but can at any moment in time be merged, should fixes in this PR be requested in master.


options/linux-headers/include/linux/cdrom.h
```diff
@@ -73,6 +86,55 @@ struct cdrom_generic_command {
 	void *reserved[1];
 };
 
+typedef struct {
+	// __be16 ?
```
Geertiebear, line 90: Why not use standard stdint integers in this struct?

Dennisbonke: As this struct is directly from the Linux kernel, I figured to not change it at all where possible. I saw this behaviour in other places too, which is why I decided against replacing them.

Geertiebear: Afaict, they are just type specifiers, so they don't mean anything at runtime. But at compile time they enforce that you only ever assign a big endian value to it. It's probably really not needed in mlibc, so I'd say to just convert them to regular stdints.

Dennisbonke: Alright, I've replaced them with the normal kernel types.

options/glibc/include/net/ethernet.h
```diff
@@ -0,0 +1,21 @@
+#ifndef _NET_ETHERNET_H
+#define _NET_ETHERNET_H
+
+#include <stdint.h>
+
+// Hmm, linux option?
+#include <linux/if_ether.h>
```
Geertiebear, line 7: Probably a good idea to have this guarded by some linux option defs.

options/linux-headers/include/linux/cdrom.h
Geertiebear: Do we use ``__ux`` anywhere else in mlibc? Why not just use ``uintx_t``?

options/posix/include/fcntl.h
```diff
@@ -35,6 +35,8 @@ extern "C" {
 #define O_SYNC __MLIBC_O_SYNC
 #define O_CLOEXEC __MLIBC_O_CLOEXEC
 #define O_PATH __MLIBC_O_PATH
+#define O_LARGEFILE __MLIBC_O_LARGEFILE
+#define O_NOATIME __MLIBC_O_NOATIME
```
Geertiebear, line 39: Might want to add these to the linux abi bits too.

Geertiebear (commented):
Left some more comments.

options/glibc/include/net/ethernet.h
Dennisbonke: But what if, we build mlibc for linux, without the linux-header option, because we use the already installed kernel headers, then that would break right? Also people that don't use linux-headers but do enable glibc will not be able to use this header otherwise. How do you suggest I proceed, guarding the entire header in a linux-header option define?

Dennisbonke:
Addressed several comments, it _should_ be fine now apart from the one comment I left. I want to try to get this PR merged soon, doing a bit of cleanup, so I'm taking it off of draft status.

options/linux-headers/include/linux/neighbour.h
```diff
@@ -0,0 +1,81 @@
...
+	NDA_PROTOCOL,
+	NDA_NH_ID,
+	NDA_FDB_EXT_ATTRS,
+	__NDA_MAX
+};
+
+#define NDA_MAX (__NDA_MAX - 1)
+
+struct nda_cacheinfo {
+	__u32 ndm_confirmed;
```
Geertiebear, line 71: Here are some ``__u*`` left.

options/glibc/include/net/ethernet.h
Geertiebear: Hmm, do we really need the header include here? The man page for ``ether_aton`` says that ``eth_addr`` is defined as ``uint8_t ether_addr_octet[6];``. That would solve this issue.

avdgrinten (changes requested):
This PR adds various function declarations w/o definitions (`qsort_r`, ...). This should be avoided since it can easily lead to linking failures (but adding stubs that simply panic is OK).

avdgrinten:
Why is the `uint32_t` -> `unsigned int` etc. change done? That does not seem like the right strategy.

Dennisbonke:
> Why is the `uint32_t` -> `unsigned int` etc. change done? That does not seem like the right strategy.

Something about kernel headers and not including stdint.h (at least not on my host). I guess we could also revert that commit and just add stdint includes where needed. Shall I do that instead?

ArsenArsen:
`stdint.h` is preferable to assuming sizes.

Dennisbonke:
> `stdint.h` is preferable to assuming sizes.

Noted, I'll get to work and recheck all headers added or changed in this PR against host headers for this issue.

avdgrinten:
We also have `bits/types.h` that provides `__mlibc_uint8` and friends for situations where we do not want to include `stdint.h`.

PR #331, opened 2021-11-05
options/ansi: implement strxfrm and iswctype

Sebastian-byte:


options/ansi/generic/ctype-stubs.cpp
```diff
@@ -254,9 +254,48 @@ wctype_t wctype(const char *cs) {
...
+#define WCTYPE_BLANK  3
+#define WCTYPE_CNTRL  4
+#define WCTYPE_DIGIT  5
+#define WCTYPE_GRAPH  6
+#define WCTYPE_LOWER  7
+#define WCTYPE_PRINT  8
+#define WCTYPE_PUNCT  9
+#define WCTYPE_SPACE  10
+#define WCTYPE_UPPER  11
+#define WCTYPE_XDIGIT 12
```
ArsenArsen, line 269: prefer enums plus optional self referential macros:

```cpp
enum {
WCTYPE_ALNUM,
#define WCTYPE_ALNUM WCTYPE_ALNUM
WCTYPE_ALPHA,
#define WCTYPE_ALPHA WCTYPE_ALPHA
...
};
```

ArsenArsen (changes requested):
stylistic nit

options/ansi/generic/ctype-stubs.cpp
Sebastian-byte: What you think now?

Geertiebear (changes requested):
Please write a test for ``strxfrm`` and squash your commits into one, then it LGTM.

Sebastian-byte:
I will write the test, but, GitHub has an option to squash and merge, there's no need to me do it I think.

options/ansi/generic/ctype-stubs.cpp
ArsenArsen: sorry for the late response, I forgot to check this.

yes, that's fine, I wasn't aware we had this enum already, and also was under the impression it's standards-defined, but since it isn't, this is far better

thanks! :)

Sebastian-byte:
Now that I think about it, strxfrm is a very simple task which likely doesn't need tests, if the functions used inside it work properly.

Geertiebear:
It would still be nice to have a test for it, it shouldn't be very hard to write, and having a larger test-bed is always good.

avdgrinten:
One issue is also that this implementation is only valid for the C locale. musl gets away with it since it only supports the C locale. For now, we also do it like that, but it might need to be revised in the future.

Issue #276, opened 2021-06-06
ansi: Implement (a subset of) strptime


Dennisbonke:
Several programs require `strptime` to properly function. Notable users of this are `wget` and `xbps-install` on Managarm. Implementing this would allow these programs to behave better.

ikbenlike:
#332 has now been merged. Do the implemented directives satisfy `xbps-install` requirements or does it also depend on locale-specific directives?

Geertiebear:
grepping for ``strptime()`` shows this: ```lib/fetch/http.c:       r = strptime(p, "%a, %d %b %Y %H:%M:%S GMT", &tm);
lib/fetch/http.c:               r = strptime(p, "%a, %d %b %Y %H:%M:%S UTC", &tm);``` so I think we're okay.

ikbenlike:
The `%a` and `%b` directives are not yet implemented. They match with the name of the day and the name of the month, respectively (both the full name and abbreviations).

Dennisbonke:
This issue still stands, and blocks `wget` on Linux from functioning without hacks. `wget` triggered the ensures for `%a`, `%A`, `%b`, `%B` and `%h`. Not all of those might be needed. If someone could take this on that would be great.

PR #332, opened 2021-11-13
Implement strptime().

ikbenlike:
This PR partially implements strptime() and adds tests for the parts which are implemented. Most of the unimplemented parts depend on locale functionality. Issue #276 is related to this.

### DIFF
--- a/options/ansi/generic/time-stubs.cpp
+++ b/options/ansi/generic/time-stubs.cpp
@@ -6,6 +6,7 @@
 #include <limits.h>
 #include <wchar.h>
 #include <stdlib.h>
+#include <ctype.h>
 
 #include <bits/ensure.h>
 #include <mlibc/debug.hpp>
@@ -564,13 +565,13 @@ char *asctime_r(const struct tm *tm, char *buf) {
 		{ "Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct",
 		  "Nov", "Dec" };
 	sprintf(buf, "%.3s %.3s%3d %.2d:%.2d%.2d %d\n",
-	             weekday_names[tm->tm_wday],
-	             month_names[tm->tm_mon],
-	             tm->tm_mday,
-	             tm->tm_hour,
-	             tm->tm_min,
-	             tm->tm_sec,
-	             tm->tm_year + 1900);
+				 weekday_names[tm->tm_wday],
+				 month_names[tm->tm_mon],
+				 tm->tm_mday,
+				 tm->tm_hour,
+				 tm->tm_min,
+				 tm->tm_sec,
+				 tm->tm_year + 1900);
 	return buf;
 }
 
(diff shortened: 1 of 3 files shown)


Sebastian-byte:
You probably should implement all the missing specifiers with the C locale.

ikbenlike:
I was thinking about taking a look at the locale functions after this so I can just Do It Right immediately (instead of hardcoding the values)

options/ansi/generic/time-stubs.cpp
```diff
@@ -578,10 +579,445 @@ char *ctime_r(const time_t *clock, char *buf) {
 	return asctime_r(localtime(clock), buf);
 }
 
-char *strptime(const char *__restrict, const char *__restrict,
+namespace {
+
+int read_number(int *result, size_t *string_index, const char *s, int ndigits) {
```
Geertiebear, line 584: How is this different than a ``scanf()``?

options/ansi/generic/time-stubs.cpp
```diff
@@ -578,10 +579,445 @@ char *ctime_r(const time_t *clock, char *buf) {
...
+    return -1;
+}
+
+int is_leapyear(int year) {
+    return year % 4 == 0 && (year % 100 != 0 || year % 400 == 0);
+}
+
+int month_and_year_to_day_in_year(int month, int year){
+    int day = month_to_day(month);
+    if(is_leapyear(year) && day < month_to_day(2)) //TODO just do 2 instead of month_to_day(2) lol
```
Geertiebear, line 622: Comments should be formal

options/ansi/generic/time-stubs.cpp
```diff
@@ -578,10 +579,445 @@ char *ctime_r(const time_t *clock, char *buf) {
...
+        }
+        state->format_index++;
+    }
+
+    return (char*)input + state->input_index;
+}
+
+} //anonymous namespace
+
+/*char *strptime(const char *__restrict, const char *__restrict,
```
Geertiebear, line 967: Just remove this

options/ansi/generic/time-stubs.cpp
```diff
@@ -578,10 +579,445 @@ char *ctime_r(const time_t *clock, char *buf) {
...
 		struct tm *__restrict) {
 	mlibc::infoLogger() << "mlibc: strptime is a stub!" << frg::endlog;
 	return nullptr;
+}*/
+
+char *strptime(const char *__restrict s, const char *__restrict format, struct tm *__restrict tm){
+    struct strptime_internal_state state = {};
+
+    char *result = strptime_internal(s, format, tm, &state);
```
Geertiebear, line 976: Is there a use for having this as a seperate function? It seems to just complicate the code more.

options/ansi/generic/time-stubs.cpp
```diff
@@ -578,10 +579,445 @@ char *ctime_r(const time_t *clock, char *buf) {
...
+
+char *strptime_internal(const char *__restrict input, const char *__restrict format,
+	struct tm *__restrict tm, struct strptime_internal_state *__restrict state) {
+    while(isspace(input[state->input_index])) state->input_index++;
+    
+    if(input[state->input_index] == '\0')
+        return NULL;
+
+    while(format[state->format_index] != '\0'){
+        if(format[state->format_index] == '%'){
```
Geertiebear, line 702: This can probably be expressed more cleanly by just doing some ``if (format[state->format_index] != '%'`` and then skipping whitespace and doing the checks. It avoids having one more layer of indentation, and is more straightforward to read.

options/ansi/generic/time-stubs.cpp
```diff
@@ -578,10 +579,445 @@ char *ctime_r(const time_t *clock, char *buf) {
...
+            state->format_index++;
+            switch(format[state->format_index]){
+                case '%':
+                    if(input[state->input_index++] != '%')
+                        return NULL;
+                    break;
+                case 'a':
+                case 'A':
+                    break;
```
Geertiebear, line 711: If there are any unimplemented cases, they should at least be logged, if not just fail the function.

Geertiebear (changes requested):

Geertiebear:
Also you didn't actually add the test to the meson file.

options/ansi/generic/time-stubs.cpp
ikbenlike: yeah I realized I forgot to remove that just after I shut off my PC

options/ansi/generic/time-stubs.cpp
ikbenlike: Certain directives are equal to a known combination of other directives (`%D`, `%r`, `%R`, `%T`) and some directives are equal to other directives depending on locale (`%x`,  `%X`). Implementing it this way, as a function that simply calls itself to handle such "composite" cases greatly simplifies the control-flow (we wouldn't need any conditional fall-through, for example) and allows it to handle arbitrary directives which are only known at runtime more easily. To me, this was the most logical and simple way to implement this.

options/ansi/generic/time-stubs.cpp
ikbenlike: As it is now, there are two "known" things in the format string: a format directive, starting with `%`, and whitespace. The format directives are handled according to specific logic, and any whitespace present in the format string is matched with zero or more whitespace characters in the input string. Any other character present in the format string, which we don't necessarily know beforehand, has to be literally present in the input string. So I'm not entirely sure how doing it the way you proposed would make this cleaner.

options/ansi/generic/time-stubs.cpp
ikbenlike: This was left as a todo for myself that I forgot about.

ikbenlike:
> Also you didn't actually add the test to the meson file.

fixed this.

options/ansi/generic/time-stubs.cpp
ikbenlike: `sscanf()` doesn't return the number of characters read, but the number of items. I need the number of characters read to keep track of the index in the input string and so on. I could re-implement `read_number()` using `sscanf()` but I'm not sure if this is necessary.

options/ansi/generic/time-stubs.cpp
ikbenlike: This is now resolved.

options/ansi/generic/time-stubs.cpp
ikbenlike: Unless I missed any, it should now `__ensure()` that unimplemented cases are not reached.

options/ansi/generic/time-stubs.cpp
Geertiebear: You can use scanf's ``%n`` for this.

options/ansi/generic/time-stubs.cpp
Geertiebear: I mean doing it like (this)[https://github.com/managarm/mlibc/blob/73b86f569e7575fcafdf5f26c3fd81312611dd0d/options/ansi/generic/stdio-stubs.cpp#L355], the same way we do it in ``scanf()``. Afaict this should work, and makes the control flow a lot easier to reason about.

options/ansi/generic/time-stubs.cpp
Geertiebear: Okay, I'm not sure I follow. The function takes some inputs, and produces some outputs and saves them to this ``state`` struct. The state is then examined in the main function afterwards. This ``state`` could also just be some local variables afaict, and have the main "body" of the function just modify these local variables.

options/ansi/generic/time-stubs.cpp
```diff
@@ -564,24 +565,490 @@ char *asctime_r(const struct tm *tm, char *buf) {
...
+
+	year /= 2;
+
+	if(year % 2)
+		year += 11;
+
+	return 7 - (year % 7) + anchor;
+}
+
+//determine day of week through the doomsday algorithm,
```
Geertiebear, line 664: Here the comment should be formal also.

options/ansi/generic/time-stubs.cpp
ikbenlike: Yes, the issue is that certain inputs expand to a combination of directives. To me the cleanest way to handle this is a recursive function, especially since we do not know all expansions that can take place beforehand.

The state is passed through to keep track of the entire state of the parser, which is also necessary for deciding what to calculate at the end (we wouldn't need to calculate the day of the week if it's already given, for example.) Since the recursive calls can modify the state of the parser, I'm not sure how you'd have them as local variables instead.

Geertiebear: Ah I see, I understand now. I think still I'd prefer there to be no extra function. The duplicated code is probably not in such large quantities (afaict) that it warrants recursion like this. In this case I'd prefer to focus on readability rather than conciseness.

ikbenlike: Well, it's not necessarily just code duplication - certain directives "expand" to a combination of locale-dependent directives. I'm not sure how this would be implemented if not with recursion, and besides, I don't think recursion is atypical for parsers.

Geertiebear (changes requested):
Almost there, all the major questions are taken care of now. There are still some informal-ish comments left in the code, and if you could rebase your "fix" commit into your first one. Then everything should be okay.

ikbenlike:
> Almost there, all the major questions are taken care of now. There are still some informal-ish comments left in the code, and if you could rebase your "fix" commit into your first one. Then everything should be okay.

Everything should be in order now.

Geertiebear (approved):
LGTM now.

Geertiebear:
Thank you very much for the PR!

PR #334, opened 2021-11-23
unistd: added a basic sethostname

streaksu:


options/posix/generic/unistd-stubs.cpp
```diff
@@ -337,6 +337,19 @@ int gethostname(char *buffer, size_t bufsize) {
 	return 0;
 }
 
+int sethostname(char *buffer, size_t bufsize) {
```
Geertiebear, line 340: The signature is ``const char *``

Geertiebear (changes requested):

options/posix/generic/unistd-stubs.cpp
streaksu: fixed, sorry for the inconvenience

Geertiebear (approved):
Not a problem, thanks for the PR!

PR #337, opened 2021-11-28
Various additions for the next round of Managarm package updates

Dennisbonke:
This PR adds various headers and fixes that the new round of updates required. For now, this consists only of the addition of `arpa/nameser.h`.

Complements managarm/bootstrap-managarm#148


options/posix/generic/posix_stdlib.cpp
```diff
@@ -469,3 +469,8 @@ int getloadavg(double *, int) {
 	__ensure(!"Not implemented");
 	__builtin_unreachable();
 }
+
+char *secure_getenv(const char *name) {
+	// We don't have a secure execution environment.
```
Geertiebear, line 474: Maybe at a log here that we don't implement it properly, and submit an issue when this is merged. It's not too difficult but not really needed for this PR.

options/posix/include/arpa/nameser.h
```diff
@@ -0,0 +1,209 @@
+#ifndef _ARPA_NAMESER_H
```
Geertiebear, line 1: The entire ``resolv`` library is actually defined as a part of BSD, and not POSIX. Should we maybe move it to its own option?

options/posix/include/arpa/nameser.h
```diff
@@ -0,0 +1,209 @@
...
+	unsigned	cd: 1;
+	unsigned	ad: 1;
+	unsigned	unused :1;
+	unsigned	ra :1;
+#endif
+	unsigned	qdcount :16;
+	unsigned	ancount :16;
+	unsigned	nscount :16;
+	unsigned	arcount :16;
+} HEADER;
```
Geertiebear, line 134: What is the use of this ``HEADER`` struct, and in which standard is it specified?

Dennisbonke: Seems to be DNS query related and as for the standard, I presume BSD too. From what I could see both glibc and musl have this struct.

options/posix/include/arpa/nameser.h
Dennisbonke: We could, but so far BSD extensions have fallen under posix. For me, this is fine, but moving is always possible.

options/posix/generic/posix_stdlib.cpp
Dennisbonke: Yep, sounds good. I'll do that as soon as I have time.

options/posix/generic/arpa-nameser-stubs.cpp
```diff
@@ -0,0 +1,22 @@
+#include <arpa/nameser.h>
+
+// The ns_get* and ns_put* functions are taken from musl.
```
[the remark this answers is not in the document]
Dennisbonke: I'll try my hand at them at the earliest convenience.

options/posix/include/arpa/nameser.h
Geertiebear: Which program needs it?

Dennisbonke: I can drop a patch from `glib` with these changes.

options/posix/include/arpa/nameser.h
Dennisbonke: I've moved this to a BSD option as discussed on Discord.

options/posix/generic/arpa-nameser-stubs.cpp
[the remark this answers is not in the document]
Geertiebear: Well the BSD man pages describe how these functions are meant to be used.

tests/bsd/ns_get_put.c
```diff
@@ -0,0 +1,20 @@
...
+	uint32_t original32 = 4861984;
+	memset(buf, 0x0, sizeof(*buf));
+	ns_put16(original16, buf);
+	uint16_t result16 = ns_get16(buf);
+	assert(result16 == original16);
+	memset(buf, 0x0, sizeof(*buf));
+	ns_put32(original32, buf);
+	uint32_t result32 = ns_get32(buf);
+	assert(result32 == original32);
+	return 0;
```
Geertiebear, line 19: It would also be nice that we also test the absolute results, and don't just check if they are internally consistent.

Dennisbonke: I'm kinda lost on what you want here. As far as I can tell this tests the absolute results too no?

Geertiebear: Hmm yeah I guess.

Geertiebear (approved):
LGTM now.

no92 (approved):

PR #338, opened 2021-12-03
Hi. Can I contribute this to mainstream for thread-safety of stdio?

tearosccebe:
Although ftrylockfile cannot be implemented without FutexLock adds a try_lock() method.

Also i do not know whether it would cause a self dead lock

### DIFF
--- a/options/ansi/generic/stdio-stubs.cpp
+++ b/options/ansi/generic/stdio-stubs.cpp
@@ -1118,12 +1118,12 @@ int vasprintf(char **out, const char *format, __gnuc_va_list args) {
 
 // Linux unlocked I/O extensions.
 
-void flockfile(FILE *) {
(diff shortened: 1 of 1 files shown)


Geertiebear:
Hi! thank you for the PR. The code looks good to be merged, but if you could clean up your commit messages that would be great. The merge commit should hopefully be able to be dropped, and the original commit message should look something like: ``options/ansi: implement flockfile() and funlockfile()``.  Some general guidelines for commit messages can be found [here](https://docs.managarm.org/handbook/contributing/commit-messages.html). You can force push to your own branch to fix these things in the same PR. Once that is done I will enable the CI run, and when that passes I will merge this PR.

Geertiebear:
Closed in favour of #343.

PR #341, opened 2021-12-08
options/ansi: implement flockfile(3) and funlockfile(3) in stdio.h ba…

tearosccebe:
…sed on POSIX standard.

Unfortunately, I do not know whether calling flockfile in the same thread would cause a self dead-lock.
Another issue is that FutexLock does not provide try_lock() method, which means we cannot implement ftrylockfile(3).


ArsenArsen:
the commit message is too long, please amend it (in the same PR, `git commit --amend` and `git push` it).

PR #344, opened 2021-12-13
meson: use meson builtins instead of shell

ArsenArsen:
This saves us a risky join by space and cleans up the code to an extent

Co-Authored-By: Eli Schwartz <eschwartz@archlinux.org>


Geertiebear (approved):
LGTM

PR #349, opened 2021-12-27
Update strerror with more errno codes

Dennisbonke:
Missing errno codes in strerror are annoying, so I added all errno codes known to mlibc to `strerror()`.


Geertiebear:
The commit should be split up, right now it's not very descriptive.

Dennisbonke:
> The commit should be split up, right now it's not very descriptive.

Addressed this comment, should be better now.

Geertiebear (approved):
LGTM.

PR #353, opened 2021-12-28
options/linux-headers: add linux/fb.h

alula:
Adds a missing header for Linux fbdev interface. Used by Vinix and needed for things like weston fbdev backend and etc.


Dennisbonke:
Shouldn't this be in linux-headers option?

Geertiebear:
Yes, and I'd prefer if we don't use ``__u32`` but instead use the ``stdint.h`` equivalents.

alula:
@Dennisbonke It... already is?

alula:
videodev2.h has __uXX types as well btw

alula:
made it just use C types as other headers do

Geertiebear:
LGTM, will squash merge when CI passes.

PR #355, opened 2022-01-14
Changes for weston 9.0.91

no92:


options/linux/include/sys/epoll.h
```diff
@@ -3,7 +3,7 @@
 
 // TODO: We only need sigset_t and not the whole signal.h
 #include <stdint.h>
-#include <signal.h>
+#include <abi-bits/signal.h>
```
Geertiebear, line 6: Why is this change needed? If this fixes the above TODO, could you remove it please?

no92: This both fixes the TODO and resolves problems with weston, where includes of this do seem correct.

no92:
Addressed the issue raised.

Geertiebear (approved):
Thanks, LGTM now!

PR #358, opened 2022-01-22
Implement FIONREAD ioctl

no92:
Implement the `FIONREAD` ioctl. This depends on the other half of the PR in managarm, managarm/managarm#410.

This obsoletes #241.

### DIFF
--- a/sysdeps/managarm/generic/file.cpp
+++ b/sysdeps/managarm/generic/file.cpp
@@ -1771,6 +1771,41 @@ int sys_ioctl(int fd, unsigned long request, void *arg, int *result) {
 		}
 		return 0;
 	}
+	case FIONREAD: {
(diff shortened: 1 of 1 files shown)


avdgrinten (approved):
LGTM. argp = NULL is allowed by the ioctl?

no92:
I can't find any instance or mention of `argp` being NULL, so we might need to check this.

no92:
Added a check that returns `EINVAL` for NULL `argp`.

avdgrinten:
Minor nitpick: if we return EINVAL, we can simply return it before doing a request.

no92:
Amended the commit to return on `!argp` before making requests.

avdgrinten (approved):
LGTM now.

PR #361, opened 2022-01-27
options/posix: fix frigg error

no92:
Depends on managarm/frigg#19.


Geertiebear:
LGTM, if you push again then the CI resets and I can merge this.

no92:
Without managarm/frigg#19 merged, CI will fail.

PR #363, opened 2022-01-29
Fix getdelim and annotate fallthroughs

no92:
This fixes minor issues with `getdelim` that prevented `xbps-reconfigure` from running for some. Fallthroughs were annotated or `break`s added, where appropriate.

### DIFF
--- a/options/ansi/generic/stdio-stubs.cpp
+++ b/options/ansi/generic/stdio-stubs.cpp
@@ -470,7 +470,7 @@ static int do_scanf(H &handler, const char *fmt, __gnuc_va_list args) {
             case 'd':
             case 'u':
                 base = 10;
-                /* fallthrough */
+                [[fallthrough]];
             case 'i': {
                 unsigned long long res = 0;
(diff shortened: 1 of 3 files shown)


options/ansi/generic/stdio-stubs.cpp
```diff
@@ -1045,7 +1045,9 @@ ssize_t getdelim(char **line, size_t *n, int delim, FILE *stream) {
 	}
 
 	char *buffer = *line;
-	size_t capacity = *n, nwritten = 0;
+	/* limit the starting capacity to 512 if buffer = NULL */
+	size_t capacity = (!buffer && *n > 512) ? 512 : *n;
```
avdgrinten, line 1049: Should `*n` affect the buffer size at all if `!buffer`? What does the man page say? What do other libcs do?

no92: On `!buffer`, the man page tells us to ignore `*n`. glibc seems to set 120 if that is the case.

PR #364, opened 2022-01-30
Fix some getsockopt issues

Dennisbonke:
During testing on Managarm, a problem with a stubbed return for getsockopt was found. This PR fixes that by returning an actual success code and passing the value the way it is expected. It also adds 3 other getsockopt stub returns, pleasing several users.


Geertiebear (approved):
LGTM.

PR #365, opened 2022-01-30
Various additions to make pure X work on Managarm

Dennisbonke:
This PR adds mlibc work that eventually should allow X.org to run in pure X mode on Managarm. As we're kinda stuck at the moment, this should be merged now, and any additional work can go into it's own PR. Notably lacking is a regex suite, we should settle on one to copy or implement our own.

### DIFF
--- a/options/posix/generic/sys-resource-stubs.cpp
+++ b/options/posix/generic/sys-resource-stubs.cpp
@@ -6,9 +6,17 @@
 #include <mlibc/debug.hpp>
 #include <mlibc/posix-sysdeps.hpp>
 
-int getpriority(int, id_t) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+int getpriority(int which, id_t who) {
+	if(!mlibc::sys_getpriority) {
+		MLIBC_MISSING_SYSDEP();
+		errno = ENOSYS;
+		return -1;
+	}
+	int value = 0;
+	if(int e = mlibc::sys_getpriority(which, who, &value); e) {
+		errno = e;
+	}
+	return value;
 }
 
 int setpriority(int which, id_t who, int prio) {
--- a/options/posix/generic/sys-resource-stubs.cpp
+++ b/options/posix/generic/sys-resource-stubs.cpp
@@ -3,15 +3,25 @@
 #include <sys/resource.h>
 
 #include <bits/ensure.h>
+#include <mlibc/debug.hpp>
 #include <mlibc/posix-sysdeps.hpp>
 
 int getpriority(int, id_t) {
 	__ensure(!"Not implemented");
 	__builtin_unreachable();
 }
-int setpriority(int, id_t, int) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+
+int setpriority(int which, id_t who, int prio) {
+	if(!mlibc::sys_setpriority) {
+		MLIBC_MISSING_SYSDEP();
+		errno = ENOSYS;
+		return -1;
+	}
+	if(int e = mlibc::sys_setpriority(which, who, prio); e) {
+		errno = e;
+		return -1;
+	}
+	return 0;
 }
 
 int getrusage(int scope, struct rusage *usage) {
--- a/options/posix/generic/unistd-stubs.cpp
+++ b/options/posix/generic/unistd-stubs.cpp
@@ -15,9 +15,10 @@
 #include <mlibc/posix-sysdeps.hpp>
 #include <mlibc/thread.hpp>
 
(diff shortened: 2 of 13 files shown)


ArsenArsen:
dreadful, I love it

avdgrinten:
I suggest to split the regex stuff off from this PR. Also, the regex engine looks like it is taken from elsewhere, w/o proper attribution (which is a copyright violation, even if its own on your private fork).

Dennisbonke:
> I suggest to split the regex stuff off from this PR. Also, the regex engine looks like it is taken from elsewhere, w/o proper attribution (which is a copyright violation, even if its own on your own fork).

Yeah I didn't do this the smartest way. The regex engine here likely shouldn't be merged at all and indeed I messed up the attribution. I will fix this in the next few hours as I'm currently on the road. I'll eventually fix this PR so it conforms more to standards on other points, at which point the regex stuff will indeed be split off or dropped all together in favor of a better solution.

Dennisbonke:
Addressed copyright violation, thanks again Alexander for bringing it to my attention.

no92:
Merging 107992d as-is would break weston versions starting from mid-2019 forwards. Instead of stubbing, a complete implementation including the managarm side would be necessary.

options/posix/generic/sys-resource-stubs.cpp
```diff
@@ -3,15 +3,17 @@
...
 	__ensure(!"Not implemented");
 	__builtin_unreachable();
 }
+
 int setpriority(int, id_t, int) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+	mlibc::infoLogger() << "mlibc: setpriority is a stub!" << frg::endlog;
+	return 0;
 }
```
avdgrinten, line 17: It'd be better to stub this on the sysdep level. Or does Linux implement this on top of a different syscall?

options/posix/generic/unistd-stubs.cpp
```diff
@@ -16,8 +16,8 @@
 #include <mlibc/thread.hpp>
 
 unsigned int alarm(unsigned int) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+	mlibc::infoLogger() << "mlibc: alarm is a stub!" << frg::endlog;
+	return 0;
 }
```
avdgrinten, line 21: This should be implemented on top of `setitimer`.

options/posix/generic/sys-resource-stubs.cpp
Dennisbonke: Stubbed with a `sys_setpriority` syscall.

no92 (approved):

options/posix/generic/unistd-stubs.cpp
```diff
@@ -15,9 +15,10 @@
...
 #include <mlibc/thread.hpp>
 
-unsigned int alarm(unsigned int) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+unsigned int alarm(unsigned int seconds) {
+	struct itimerval it = {it.it_value.tv_sec = seconds}, old = {0};
+	setitimer(ITIMER_REAL, &it, &old);
+	return old.it_value.tv_sec + !! old.it_value.tv_usec;
 }
```
avdgrinten, line 22: What if `setitimer` fails here?

Dennisbonke: Musl doesn’t seem to handle that explicitly, and I guess setitimer takes care of errno in that case. On that note, the man page of alarm doesn’t list what should happen with respect to errors, nor does it list valid errno’s.

Dennisbonke: Hmm I’m not so sure about that. We could check the return value of setitimer and return 0 if it failed, but flat out crashing seems not the right solution.

Geertiebear: There are only two failing conditions for ``setitimer()``, which is if the pointers aren't valid or ``which`` is not one of ``ITIMER_REAL``, ``ITIMER_VIRTUAL``, or ``ITIMER_PROF``, neither of which happens in this case.

Geertiebear (approved):
LGTM

options/posix/generic/unistd-stubs.cpp
Geertiebear: There are only two failing conditions for ``setitimer()``, which is if the pointers aren't valid or ``which`` is not one of ``ITIMER_REAL``, ``ITIMER_VIRTUAL``, or ``ITIMER_PROF``, neither of which happens in this case.

PR #367, opened 2022-02-02
sysdeps/managarm: Implement DRM PRIME ioctls

no92:
Implements the ioctls for DRM PRIME, which depend on bragi additions in managarm.


avdgrinten (approved):
LGTM. I assume we need to merge the Managarm PR first?

PR #369, opened 2022-02-04
options/glibc: Add empty in_systm header to fix the cups port

Dennisbonke:
`cups` really wants `netinet/in_systm.h` to exist, but it doesn't need anything from it. Provide an empty file to make it happy.


Geertiebear (approved):
LGTM.

PR #370, opened 2022-02-04
Implement memfd and file sealing

no92:
This is the mlibc counterpart to managarm/managarm#418.

### DIFF
--- a/abis/mlibc/fcntl.h
+++ b/abis/mlibc/fcntl.h
@@ -26,6 +26,8 @@
 #define F_SEAL_SHRINK 0x0002
 #define F_SEAL_GROW   0x0004
 #define F_SEAL_WRITE  0x0008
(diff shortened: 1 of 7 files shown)


avdgrinten (approved):
LGTM, can be merged after Managarm changes have landed.

no92:
Minor changes to allow for both headers to be used for `memfd_create`.

avdgrinten:
Linux headers typically do not contain functions. Why is this done?

no92:
An old `mesa` included `linux/memfd.h` expecting memfd_create to be defined there, while not including `sys/mman.h`.

avdgrinten:
But even on Linux, `linux/memfd.h` does not have the function, does it? So we should not add it there either.

no92:
Fair enough, moved `memfd_create` to `sys/mman.h`.

PR #372, opened 2022-02-05
options/posix: implement setlogmask

no92:


avdgrinten (commented):
Are there any atomicity guarantees that we need to take into account?

no92:
I can't find any mention of any guarantees given about atomicity. glibc's impl looks similar to this, while musl's does have a lock around this function.

PR #373, opened 2022-02-05
options/ansi: implement wcs{ncasecmp,dup}

no92:


Geertiebear:
LGTM generally, but tests for this new functionality would be nice.

PR #374, opened 2022-02-08
Various fixes on the way to wine

Dennisbonke:
This PR adds various constants, defines, structs and functions that `wine` requires. It certainly isn’t everything but this can already get reviewed and merged.

Has an abi-break counterpart in #452, which is not required for this to be merged, but is required for wine.
Also includes some fixes for Chromium incidentially, so part of the Chromium on Managarm project.

### DIFF
--- a/abis/mlibc/in.h
+++ b/abis/mlibc/in.h
@@ -89,6 +89,8 @@ struct group_source_req {
 #define IPPROTO_RAW 4
 #define IPPROTO_TCP 5
 #define IPPROTO_UDP 6
+#define IPPROTO_IGMP 7
+#define IPPROTO_IPIP 8
 #define IPPROTO_ESP 50
 #define IPPROTO_AH 51
 
--- a/abis/mlibc/in.h
+++ b/abis/mlibc/in.h
@@ -126,6 +126,8 @@ struct group_source_req {
 #define IP_MULTICAST_LOOP         34
 #define IP_ADD_MEMBERSHIP         35
 #define IP_DROP_MEMBERSHIP        36
+#define IP_UNBLOCK_SOURCE         37
+#define IP_BLOCK_SOURCE           38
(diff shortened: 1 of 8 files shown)


avdgrinten:
Renaming functions is an ABI break.

options/posix/include/sys/select.h
```diff
@@ -16,15 +16,15 @@
...
-#define FD_ZERO FD_ZERO
+void __FD_CLR(int fd, fd_set *);
+int __FD_ISSET(int fd, fd_set *);
+void __FD_SET(int fd, fd_set *);
+void __FD_ZERO(fd_set *);
+
+#define FD_CLR(fd, set) __FD_CLR(fd, set)
+#define FD_ISSET(fd, set) __FD_ISSET(fd, set)
+#define FD_SET(fd, set) __FD_SET(fd, set)
+#define FD_ZERO(set) __FD_ZERO(set)
```
64, line 27: Why do these need to be macros? Is it not possible to do `#define FD_CLR(fd, set) FD_CLR(fd, set)`?

ArsenArsen: selfref macros would be preferable

ArsenArsen: update: microsoft (and in turn wine) are insane, FD_SET is declared as a type on windows and it's part of the api, so we might have to rename them

I'm open to and would prefer literally any other solution

qookei: Maybe we could rename FD_xxx to __FD_xxx, and provide the FD_xxx symbols as aliases for ABI backwards compat?

64: If we are going to break ABI for pthread anyway then we might as well get rid of such hacks when we do so

Dennisbonke: I've removed this part of the PR and split it out into #452, that means we can merge this PR now.

PR #378, opened 2022-02-12
meta: adjust .gitignore file

no92:


.gitignore
```diff
@@ -1,10 +1,14 @@
 build
 
-# the menson vsocde extension auto generates another build directory
-# called `builddir`
-builddir
+# exclude dotfiles
+.*
+!.git*
+!.mailmap
```
ArsenArsen, line 6: why exclude these?

.gitignore
```diff
@@ -1,10 +1,14 @@
...
+!subprojects/*.wrap
+*.xbstrap
 
-subprojects/cxxshim
-subprojects/frigg
-.vscode
-.clang-format
+# generated by editors
+builddir
+*.swp
```
ArsenArsen, line 14: use global gitignores to ignore products of personalized setups

no92:
Addressed comments and rebased.

PR #379, opened 2022-02-13
sysdeps/managarm: fix return value of fcntl(F_SETFL)

64:
As per the man page we must ensure that the return value is set to zero on success. Otherwise, the calling function reads this variable when uninitialized.


Geertiebear (approved):
LGTM

PR #382, opened 2022-02-13
Small fixes to fix CI

Dennisbonke:
This PR fixes an error in 060b995c318c2fb08e88525815a32719f5236b81 where the installation directory of `in_systm.h` was inadvertently changed to go to the wrong location, breaking the `cups` port. It also adds a `struct iphdr` to the `ip.h` linux kernel header in order to satisfy `qemu` picking up said header and wanting to use its contents.


Geertiebear (approved):
LGTM

PR #385, opened 2022-02-14
posix: Make accept() fill struct sockaddr

mintsuki:


mintsuki:
Yeah, that would also be a possibility.

no92:
Superseded by #427.

PR #386, opened 2022-02-14
posix: Implement {get,set}itimer()

mintsuki:


Geertiebear (changes requested):
The implementations should check if the sysdep exists.

mintsuki:
> The implementations should check if the sysdep exists.

Done.

Geertiebear (approved):
LGTM.

PR #387, opened 2022-02-14
posix: Implement unlink() using sys_unlinkat()

mintsuki:


Geertiebear:
Why is this change needed? If ``sys_unlink`` is redundant, can we remove it?

mintsuki:
> Why is this change needed? If `sys_unlink` is redundant, can we remove it?

Yeah, you probably can. The change is needed because I believe it makes more sense for mlibc to implement lesser functions like `unlink()` by means of more generic syscalls like `unlinkat()`. Some OSes may not have an `unlink()` syscall at all. It also puts it more in line with other sysdeps.

mintsuki:
OTOH you could make the argument that OSes without an `unlink()` syscall can just emulate it in their own sysdeps, thus allowing OSes without `unlinkat()` to not have to emulate *that*, as it is significantly trickier.

In that case it would make sense to review some of the other sysdeps as well, to make them more granular.

Geertiebear:
I think in general it's fine to have less granular sysdeps, as platforms can just implement the functionality they're interested in.

>Yeah, you probably can. The change is needed because I believe it makes more sense for mlibc to implement lesser functions like unlink() by means of more generic syscalls like unlinkat(). Some OSes may not have an unlink() syscall at all. It also puts it more in line with other sysdeps.

This makes sense, and in this case I'd ask you to also remove the ``sys_unlink`` sysdep, since it's no longer needed then.

mintsuki:
Done.

PR #398, opened 2022-02-22
options/posix: Implement execlp

Dennisbonke:
This PR implements `execlp()` which is needed for `havoc`.

### DIFF
--- a/options/posix/generic/unistd-stubs.cpp
+++ b/options/posix/generic/unistd-stubs.cpp
@@ -123,9 +123,24 @@ int execle(const char *path, const char *arg0, ...) {
 	return execve(path, argv, envp);
 }
 
-int execlp(const char *, const char *, ...) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+// This function is taken from musl
+int execlp(const char *file, const char *argv0, ...) {
+	int argc;
+	va_list ap;
+	va_start(ap, argv0);
+	for(argc = 1; va_arg(ap, const char *); argc++);
+	va_end(ap);
+	{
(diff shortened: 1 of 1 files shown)


no92 (approved):

options/posix/generic/unistd-stubs.cpp
```diff
@@ -123,9 +123,24 @@ int execle(const char *path, const char *arg0, ...) {
...
+// This function is taken from musl
+int execlp(const char *file, const char *argv0, ...) {
+	int argc;
+	va_list ap;
+	va_start(ap, argv0);
+	for(argc = 1; va_arg(ap, const char *); argc++);
+	va_end(ap);
+	{
+		int i;
+		char *argv[argc + 1];
```
Geertiebear, line 135: I do not like the use of VLA's here. I think it's fine here to malloc an array and never free it, since ``argv`` for a program is never freed anyways.

Geertiebear (changes requested):

options/posix/generic/unistd-stubs.cpp
Dennisbonke: Considering it's already in mlibc master (see [here](https://github.com/managarm/mlibc/blob/bc8dd2e192018ddb0c49be49a4885ddd978f4028/options/posix/generic/unistd-stubs.cpp#L115)), I'd say this is fine for now.

64: I agree, we should open a separate issue for use of VLAs.

Geertiebear (approved):
LGTM.

PR #400, opened 2022-03-07
options/ansi: Implement scanf

Kyota-exe:
This PR implements `scanf` using the same code as `fscanf`.
The only difference is that instead of passing the file stream argument, it passes `stdin` to `vfscanf`.


avdgrinten (approved):
LGTM.

Kyota-exe:
Wasn’t compiling because of a typo. I thought I had checked that it built fine locally, but I guess I forgot, my bad.

ArsenArsen:
squash these commits, and wait for CI, after it's probably alright.

should there be a test too?

Kyota-exe:
Squashed the commits.

PR #403, opened 2022-03-29
options/ansi: Handle EOF in ungetc properly

Dennisbonke:
This, together with a locking patch in `gtk+3` (or a change in the way we lock things in `file-io.cpp`), allows `gtk+3` to run again on Managarm.


options/ansi/generic/file-io.cpp
```diff
@@ -221,10 +221,14 @@ int abstract_file::write(const char *buffer, size_t max_size, size_t *actual_siz
...
 
-void abstract_file::unget(char c) {
+int abstract_file::unget(char c) {
+	if(!__offset) {
+		return EOF;
+	}
 	__ensure(__offset);
 	__offset--;
 	__buffer_ptr[__offset] = c;
+	return 0;
```
ArsenArsen, line 231: why not return ``c`` here and remove the entire block of changes below in favor of just return unget

ArsenArsen (changes requested):
also commit messages shouldn't be terminated by punctuation

options/ansi/generic/file-io.cpp
Dennisbonke: Done as requested, also updated the commit message as requested.

64:
ABI diff fails due to internal function change -- should be okay.

Geertiebear (changes requested):
Could you add a test for this behaviour?

avdgrinten (commented):
I think this behavior is not correct. ungetc() should also work at the start of the file. We should simply store a `optional<char>` inside the `mlibc::abstract_file` to make ungetc work.

Dennisbonke:
With the generous help of @64, both a test and the comments by @avdgrinten should be addressed now.

Geertiebear:
I'm still not sure I agree with this behaviour. The man page of ``ungetc()`` says that there can be an infinite amount of characters pushed to the stream, so long as buffering allows for it. Right now, the way I read this code there can only be one character pushed. Technically this is within spec, but also not entirely satisfactory. I'm not sure that the current behaviour for handling subsequenct ungets is correct at all though, since it erases the current unget character. 

Also at the end of the test please unlink the temporary file.

PR #405, opened 2022-04-09
posix: multiple resolv.conf entries, avoid ipv6 failure on dns functions

AtieP:
* Allow any resolv.conf entry to be retrieved.
* Allow only IPV4 resolv.conf entries on lookup_name_dns() and lookup_addr_dns().
* If one resolv.conf entry doesn't work on lookup_nane_dns() and lookup_addr_dns(), use the next one.

### DIFF
--- a/options/posix/generic/lookup.cpp
+++ b/options/posix/generic/lookup.cpp
@@ -18,6 +18,8 @@ namespace {
 	constexpr unsigned int RECORD_A = 1;
 	constexpr unsigned int RECORD_CNAME = 5;
 	constexpr unsigned int RECORD_PTR = 12;
+
+	struct nameserver_data default_nameserver("127.0.0.1", AF_INET);
 }
 
 static frg::string<MemoryAllocator> read_dns_name(char *buf, char *&it) {
@@ -86,10 +88,27 @@ int lookup_name_dns(struct lookup_result &buf, const char *name,
 	// for dns
 	sin.sin_port = htons(53);
 
-	auto nameserver = get_nameserver();
-	if (!inet_aton(nameserver ? nameserver->name.data() : "127.0.0.1", &sin.sin_addr)) {
-		mlibc::infoLogger() << "lookup_name_dns(): inet_aton() failed!" << frg::endlog;
-		return -EAI_SYSTEM;
+	int nameserver_idx = 0;
+	while (true) {
+		auto nameserver = get_nameserver(nameserver_idx++);
+		if (!nameserver) {
+			// no more nameservers in resolf.conf,
+			// use default
+			if (!inet_pton(default_nameserver.af, default_nameserver.name.data(), &sin.sin_addr)) {
+				mlibc::infoLogger() << "lookup_name_dns(): no nameservers available!" << frg::endlog;
+				return -EAI_SYSTEM;
+			}
+			break;
+		}
+
+		// TODO: ipv6
+		if (nameserver->af != AF_INET)
+			continue;
+
+		if (!inet_pton(nameserver->af, nameserver->name.data(), &sin.sin_addr))
+			continue;
+
+		break;
 	}
 
 	int fd = socket(AF_INET, SOCK_DGRAM, 0);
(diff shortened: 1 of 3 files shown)


ArsenArsen:
> Allow only IPV4 resolv.conf entries on lookup_name_dns() and lookup_addr_dns().

Why is this needed?

AtieP:
> > Allow only IPV4 resolv.conf entries on lookup_name_dns() and lookup_addr_dns().
> 
> Why is this needed?

Because having an IPV6 address breaks the code, since there isn't support for them yet in inet_pton().

ArsenArsen:
What special work is even needed for them to work in the DNS code? This is not AAAA records, this is socket(AF_INET6) instead of AF_INET.

AtieP:
> 

I meant that there isn't IPV6 inet_pton() support yet, so passing a IPV6 address to it just breaks, and what lookup_name_dns()/lookup_addr_dns() do is use that exact function.

options/posix/include/mlibc/resolv_conf.hpp
```diff
@@ -10,11 +10,15 @@ namespace mlibc {
 struct nameserver_data {
 	nameserver_data()
 	: name(getAllocator()) {}
+	nameserver_data(const char *name_, int af_)
```
Geertiebear, line 13: This is a really minor nitpick, but we don't do this in the managarm code style, nor is it necessary.

options/posix/generic/lookup.cpp
```diff
@@ -86,10 +88,27 @@ int lookup_name_dns(struct lookup_result &buf, const char *name,
...
 
-	auto nameserver = get_nameserver();
-	if (!inet_aton(nameserver ? nameserver->name.data() : "127.0.0.1", &sin.sin_addr)) {
-		mlibc::infoLogger() << "lookup_name_dns(): inet_aton() failed!" << frg::endlog;
-		return -EAI_SYSTEM;
+	int nameserver_idx = 0;
+	while (true) {
+		auto nameserver = get_nameserver(nameserver_idx++);
+		if (!nameserver) {
+			// no more nameservers in resolf.conf,
```
Geertiebear, line 95: Another style note: comments should be formally written. So with proper punctuation and capitalization.

options/posix/generic/lookup.cpp
```diff
@@ -215,10 +234,27 @@ int lookup_addr_dns(frg::span<char> name, frg::array<uint8_t, 16> &addr, int fam
 	// for dns
 	sin.sin_port = htons(53);
 
-	auto nameserver = get_nameserver();
-	if (!inet_aton(nameserver ? nameserver->name.data() : "127.0.0.1", &sin.sin_addr)) {
-		mlibc::infoLogger() << "lookup_name_dns(): inet_aton() failed!" << frg::endlog;
-		return -EAI_SYSTEM;
+	int nameserver_idx = 0;
```
Geertiebear, line 237: I feel like this routine can be split up into a function.

options/posix/generic/lookup.cpp
```diff
@@ -215,10 +234,27 @@ int lookup_addr_dns(frg::span<char> name, frg::array<uint8_t, 16> &addr, int fam
...
+	while (true) {
```
Geertiebear, line 238: Also this while loop doesn't need to be a ``while(true)`` loop, you can also write ``while(nameserver = get_nameserver())``.

options/posix/generic/resolv_conf.cpp
```diff
@@ -20,15 +24,30 @@ frg::optional<struct nameserver_data> get_nameserver() {
...
+				cur++;
+				continue;
+			}
+
 			char *end;
 			for (pos = line + 11; isspace(*pos); pos++);
 			for (end = pos; *end && !isspace(*end); end++);
 			*end = '\0';
-			ret.name = frg::string<MemoryAllocator>(
+			auto str = frg::string<MemoryAllocator>(
```
Geertiebear, line 37: Why this change?

Geertiebear (commented):
Some comments, nothing major. Another improvement that could be made is caching ``resolv.conf`` entries so we don't have to parse the file on every request.

ArsenArsen:
> Some comments, nothing major. Another improvement that could be made is caching `resolv.conf` entries so we don't have to parse the file on every request.

`resolv.conf` changes dynamically

PR #410, opened 2022-04-18
More work towards shadow support

Dennisbonke:
This PR implements `getspnam()` and `getspnam_r`, fixes the function signature of `setpgrp` to match the System V version, just like glibc and musl do, adds a stub `fgetspent` and stubs `sys_tcdrain` in Managarm. All of this works towards running the `shadow` package on Managarm.

As more work is expected, this is a draft for now.

### DIFF
--- a/options/glibc/generic/shadow-stubs.cpp
+++ b/options/glibc/generic/shadow-stubs.cpp
@@ -1,9 +1,24 @@
 #include <shadow.h>
+#include <errno.h>
+#include <limits.h>
+#include <pthread.h>
+#include <stdlib.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <sys/stat.h>
 
 #include <bits/ensure.h>
 #include <mlibc/debug.hpp>
 
-// Code taken from musl
+/*
+ * The code in this file is largely based on or taken from musl.
+ * This includes:
+ * - xatol
+ * - __parsespent
+ * - cleanup
+ * - getspnam_r
+ * - getspnam
+ */
 #define NUM(n) ((n) == -1 ? 0 : -1), ((n) == -1 ? 0 : (n))
 
 int putspent(const struct spwd *sp, FILE *f) {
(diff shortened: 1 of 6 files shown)


64:
is that code copied from musl?

Dennisbonke:
> is that code copied from musl?

It is, I believe I left a comment about it in the source code.

options/glibc/generic/shadow-stubs.cpp
```diff
@@ -17,6 +24,150 @@ int putspent(const struct spwd *sp, FILE *f) {
...
+	if(**s == ':' || **s == '\n') {
+		return -1;
+	}
+	for(x = 0; **s - '0' < 10U; ++*s) {
+		x = 10 * x + (**s - '0');
+	}
+	return x;
+}
+
+int __parsespent(char *s, struct spwd *sp) {
```
Geertiebear, line 38: Is this supposed to be a public function?

Dennisbonke: Whoops, certainly not. As far as I can tell this should be internal. Good catch.

Dennisbonke: Fixed

options/posix/include/unistd.h
```diff
@@ -200,7 +200,7 @@ int setegid(gid_t);
 int seteuid(uid_t);
 int setgid(gid_t);
 int setpgid(pid_t, pid_t);
-pid_t setpgrp(pid_t, pid_t);
+int setpgrp(void);
```
Geertiebear, line 203: Maybe we can conform to the POSIX.1 version, which has ``pid_t`` as return type.

options/posix/generic/unistd-stubs.cpp
```diff
@@ -680,8 +680,8 @@ int setpgid(pid_t pid, pid_t pgid) {
 	return 0;
 }
 
-pid_t setpgrp(pid_t pid, pid_t pgid) {
-	return setpgid(pid, pgid);
+int setpgrp(void) {
```
Geertiebear, line 683: Here too of course.

Geertiebear (commented):
For the code taken from musl, the comment at the top of the file doesn't really make it clear that multiple blocks of code are taken from musl. Please fix that to be more clear.

options/posix/include/unistd.h
Dennisbonke: My man page does not list a `setpgrp` with a `pid_t` return type, I only have these two
```
int setpgrp(void);                   /* System V version */
int setpgrp(pid_t pid, pid_t pgid);  /* BSD version */
```

Geertiebear: https://linux.die.net/man/2/setpgrp

options/posix/generic/unistd-stubs.cpp
Dennisbonke: Done as requested.

options/posix/include/unistd.h
Dennisbonke: Done as requested.

Dennisbonke:
I've improved the attribution as requested, ready for another review.

64:
ABI checker complains since we changed `setpgrp` to take no arguments, but it was wrong before so I'm not sure that's an issue.

64:
This needs to be rebased to pass CI.

no92 (approved):

PR #411, opened 2022-04-18
ci: Several glibc related fixes

Dennisbonke:
This should fix the ansi only build on the ci.

### DIFF
--- a/options/ansi/musl-generic-math/libm.h
+++ b/options/ansi/musl-generic-math/libm.h
@@ -16,8 +16,6 @@
 #include <stdint.h>
 #include <float.h>
 #include <math.h>
-//#include <complex.h>
-#include <endian.h>
 
 #define INT_MAX __INT_MAX__
 #define INT_MIN (-INT_MAX - 1)
--- a/ci/bootstrap.yml
+++ b/ci/bootstrap.yml
@@ -125,6 +125,8 @@ packages:
         - "-Dbuild_tests=true"
         - "-Ddisable_posix_option=true"
         - "-Ddisable_linux_option=true"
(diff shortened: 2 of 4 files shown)


options/ansi/musl-generic-math/libm.h
```diff
@@ -17,7 +17,7 @@
 #include <float.h>
 #include <math.h>
 //#include <complex.h>
-#include <endian.h>
+// #include <endian.h>
```
ArsenArsen, line 20: why was this done? why is it a comment

Dennisbonke: endian.h is provided by the glibc option. I followed the style of complex.h and just commented it out.

ArsenArsen: why isn't it included with it if it is? is it even required or just a false transitive include?

Dennisbonke: We can't use that if we build ansi-only?

Geertiebear, line 20: Don't comment it out, just remove it. Same for the ``complex.h`` include.

Geertiebear (changes requested):

options/ansi/musl-generic-math/libm.h
Geertiebear: This is not a public header, and the CI passes without the ``endian.h`` include, so we can assume nothing will break with this change.

Dennisbonke: Done as requested, should be fine now pending CI completion.

Geertiebear (approved):
LGTM now.

PR #415, opened 2022-04-27
Various additions to compile NSPR and NSS

Dennisbonke:
This PR adds some defines and structs to please `NSPR` and `NSS`.


options/bsd/include/sys/queue.h
```diff
@@ -0,0 +1,106 @@
...
+}
+#define TAILQ_ENTRY(type)	_TAILQ_ENTRY(struct type,)
+
+/*
+ * Tail queue functions.
+ */
+#define	TAILQ_INIT(head) do {						\
+	(head)->tqh_first = NULL;					\
+	(head)->tqh_last = &(head)->tqh_first;				\
+} while (/*CONSTCOND*/0)
```
Geertiebear, line 30: Remove these ``CONSTCOND`` comments.

Geertiebear (changes requested):

avdgrinten:
The queue implementation seems to be taken from somewhere else. Remember to add licenses!

options/bsd/include/sys/queue.h
Dennisbonke: Done as requested.

Dennisbonke:
> The queue implementation seems to be taken from somewhere else. Remember to add licenses!

The header was taken from glibc, but the copyright header in it seems to be a BSD-3-Clause style header, so I think we're good, but I could be wrong.

Geertiebear (approved):
LGTM now.

PR #416, opened 2022-04-28
options/ansi: Avoid multiple definition errors in math.h

Dennisbonke:
This issue came to light when compiling `libvpx`, and this seems to fix the problems.

### DIFF
--- a/options/ansi/include/math.h
+++ b/options/ansi/include/math.h
@@ -71,21 +71,21 @@ int __fpclassifyl(long double x);
 // [C11/7.12.14 Comparison macros]
 #define isunordered(x,y) (isnan((x)) ? ((void)(y),1) : isnan((y)))
 
-inline int __mlibc_isless(double_t x, double_t y) { return !isunordered(x, y) && x < y; }
-inline int __mlibc_islessf(float_t x, float_t y) { return !isunordered(x, y) && x < y; }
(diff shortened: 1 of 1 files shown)


64:
I think we reached the conclusion that it's sufficient to just mark the functions as `static __inline__` and keep them in the header

Dennisbonke:
> I think we reached the conclusion that it's sufficient to just mark the functions as `static __inline__` and keep them in the header
> 
> The issue arises from gnu89 treating inline functions differently to c99. Maybe a comment should be added to this effect (explaining why we use `__inline__`)

Would this constitute an ABI break? Don't think so right?

64:
According to @ArsenArsen: no, because those symbols aren't present in libc.so (since they're never used by us internally)

64:
And frankly while we're at it, we could just do what glibc does and use the gcc builtins?

https://github.com/bminor/glibc/blob/118a2aee07f64d605b6668cbe195c1f44eac6be6/math/math.h#L1299-L1310

Geertiebear (changes requested):
Please adjust this PR to the strategy discussed on Discord.

Dennisbonke:
Performed the requested changes, tested `libvpx` and that builds successfully. Should be good to go now.

PR #419, opened 2022-05-03
options/posix: Implement uname

streaksu:


abis/mlibc/utsname.h
```diff
@@ -0,0 +1,13 @@
+#ifndef _ABIBITS_UTSNAME_T_H
+#define _ABIBITS_UTSNAME_T_H
+
+struct utsname {
+	char sysname[65];
+	char nodename[65];
+	char release[65];
+	char version[65];
+	char machine[65];
+	// char domainname[65]; Uncommenting this makes the ABI gods unhappy.
```
ArsenArsen, line 10: where does this field come from?

also if it's not appropriate to add it, just drop it and we can add it to abi-break later

sysdeps/dripos/meson.build
```diff
@@ -35,6 +35,7 @@ if not no_headers
 		'include/abi-bits/uid_t.h',
 		'include/abi-bits/wait.h',
 		'include/abi-bits/limits.h',
+		'include/abi-bits/utsname.h',
```
ArsenArsen, line 38: was formatting changed here?

abis/mlibc/utsname.h
streaksu: `uname` for some platforms like linux includes that field (https://man7.org/linux/man-pages/man2/uname.2.html), hidden behind a define, the plan was uncommenting it in a following pr to abi-break.

sysdeps/dripos/meson.build
streaksu: I don't think so, at least it does not show in my editor, is that file not meant to use tabs?

abis/mlibc/utsname.h
ArsenArsen: I'm not aware of any use in practice. We could merge without it and follow up on abi-break to bunch it in whenever the next break happens.

sysdeps/dripos/meson.build
ArsenArsen: Disregard, fault of the GitHub app.

options/posix/generic/sys-utsname.cpp
```diff
@@ -0,0 +1,24 @@
...
+#include <sys/utsname.h>
+#include <errno.h>
+
+#include <bits/ensure.h>
+#include <mlibc/debug.hpp>
+#include <internal-config.h>
+#include <mlibc/posix-sysdeps.hpp>
+
+int uname(struct utsname *p) {
+	__ensure(p);
```
Geertiebear, line 12: Instead of this ``__ensure`` we should just set ``errno`` to ``EINVAL`` and return -1. This is also what glibc does.

sysdeps/linux/include/abi-bits/utsname.h
```diff
@@ -0,0 +1 @@
+../../../../abis/mlibc/utsname.h
```
Geertiebear, line 1: Linux should use the Linux ABI. The header should also be in the Linux abi-bits.

Geertiebear (changes requested):
I don't like the commented out field. Either just remove it and open an issue, or uncomment the field and submit this PR to the ``abi-break`` branch.

streaksu:
The comment was removed and the `__ensure` was replaced for setting the errno to `EFAULT`, which imo makes more sense than `EINVAL`, Linux does this as well.
I'll submit the correct `utsname` for Linux in another PR to the `abi-break` branch (since it breaks the Linux ABI).

ArsenArsen:
Why is that even checked here? I'd leave it up to the sysdep (as mlibc itself doesn't need to access the struct at all).

options/posix/meson.build
```diff
@@ -43,7 +43,7 @@ libc_sources += files(
 	'generic/sys-times.cpp',
 	'generic/sys-time-stubs.cpp',
 	'generic/sys-uio.cpp',
-	'generic/sys-utsname-stubs.cpp',
+	'generic/sys-utsname.cpp',
```
Geertiebear, line 46: Why was this file renamed?

Geertiebear (commented):
The Linux ABI still isn't fixed.

options/posix/meson.build
streaksu: Because its not a stub anymore

streaksu:
I said I would break the ABI of Linux in a PR to abi-break following this one, given that the other changes, that are essentially adding a sysdep, dont require that treatment. Is this the wrong approach? Do I move this all to abi-break?

Geertiebear:
Then for style I'd still rather add the "incomplete" struct to the Linux ABI right now and then submit a different PR when this is merged.

Geertiebear (approved):
LGTM now.

PR #420, opened 2022-05-04
options/glibc: Include byteswap.h in endian.h

borrrden:
Otherwise anyone including it will get build errors from the lack of __bswap family functions


avdgrinten (approved):
LGTM.

PR #421, opened 2022-05-11
options/posix: make timer{add,sub} take 2 const args

no92:
While making the first 2 args `const` is not required by the spec, some code seems to assume that they are. Because an implementation should not modify them anyways, making them const is justified to avoid patching.

### DIFF
--- a/options/posix/generic/sys-time-stubs.cpp
+++ b/options/posix/generic/sys-time-stubs.cpp
@@ -21,12 +21,12 @@ int gettimeofday(struct timeval *__restrict result, void *__restrict unused) {
 	return 0;
 }
 
-void timeradd(struct timeval *, struct timeval *, struct timeval *) {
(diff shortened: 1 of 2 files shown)


Dennisbonke:
https://man7.org/linux/man-pages/man3/timeradd.3.html does not mention them being const tho?

Dennisbonke:
> > https://man7.org/linux/man-pages/man3/timeradd.3.html does not mention them being const tho?
> 
> 
> 
> That's because glibcs implementation is a macro for some reason.
> 
> 
> 
> Please also implement this function while you're at it, it should be possible while taking const args, and it should be trivial. I wonder why it wasn't done already

You can lift the implementations from the chromium PR, I had to the entire family for it.

Geertiebear (approved):
These changes LGTM. This PR doesn't need an implementation, it's just trying to fix a definition issue, an implementation (while trivial), can be left to another PR.

PR #422, opened 2022-05-11
sysdeps/managarm: implement sys_uname

no92:


sysdeps/managarm/generic/file.cpp
```diff
@@ -4682,5 +4682,23 @@ int sys_memfd_create(const char *name, int flags, int *fd) {
...
+	strcpy(buf->nodename, "?");
+	strcpy(buf->release, "?");
+	strcpy(buf->version, "?");
+#if defined(__x86_64__)
+	strcpy(buf->machine, "x86_64");
+#elif defined (__aarch64__)
+	strcpy(buf->machine, "aarch64");
+#else
+#	error Unknown architecture
+#endif
```
ArsenArsen, line 4698: What about RISC-V?

ArsenArsen (changes requested):

sysdeps/managarm/generic/file.cpp
Geertiebear: Considering we currently don't support risc-v at all in mlibc, I think this is okay.

Geertiebear (approved):
LGTM.

PR #423, opened 2022-05-11
options/posix: make freelocale a noop

no92:
This follows the precedent of the other locale functions and allows sway to run.


Geertiebear (changes requested):
Please add a warning log to say that it is a no-op.

Geertiebear (approved):
LGTM.

PR #429, opened 2022-05-15
abis,options: add missing defines for alacritty

64:
Also moves some defines from mman.h into abi bits.


options/linux/generic/pty-stubs.cpp
```diff
@@ -15,7 +15,6 @@
 int openpty(int *mfd, int *sfd, char *name, const struct termios *ios, const struct winsize *win) {
 	__ensure(!name);
 	__ensure(!ios);
-	__ensure(!win);
```
Geertiebear, line 18: Maybe add a warning that says the win argument is ignored.

Geertiebear (changes requested):

Geertiebear (approved):
LGTM.

PR #433, opened 2022-05-18
options/ansi: fix alignment checks in aligned_alloc

64:
`posix_memalign` requires that align is a power of two multiple of `sizeof(void *)`, but C11's `aligned_alloc` does not. Hence we must make sure the alignment is adjusted correctly when passing to `posix_memalign`.

We also add an extra check for the case that a bad alignment is passed to `aligned_alloc`.

Also adds some regression tests for these cases.

### DIFF
--- a/options/ansi/generic/stdlib-stubs.cpp
+++ b/options/ansi/generic/stdlib-stubs.cpp
@@ -184,6 +184,11 @@ void srand(unsigned int s) {
 
 void *aligned_alloc(size_t alignment, size_t size) {
 	void *ptr;
+
+	// posix_memalign requires that the alignment is a multiple of sizeof(void *).
+	if (alignment < sizeof(void *))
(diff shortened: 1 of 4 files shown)


options/ansi/generic/stdlib-stubs.cpp
```diff
@@ -184,6 +184,15 @@ void srand(unsigned int s) {
 
 void *aligned_alloc(size_t alignment, size_t size) {
 	void *ptr;
+
+	// alignment must be a power of two, and size % alignment must be 0
+	if (alignment & (alignment - 1) || size & (alignment - 1))
+		return nullptr;
```
avdgrinten, line 190: According to the Linux man page, `aligned_alloc` sets `errno`.

avdgrinten (approved):
LGTM.

PR #435, opened 2022-05-20
meson: find frigg and cxxshim via pkg-config

64:
This allows people to have multiple build folders on the go at the same time (e.g one for Managarm and one for testing on Linux).

Also gets rid of the meson complaint about c++20 language settings.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -37,14 +37,13 @@ internal_conf = configuration_data()
 mlibc_conf = configuration_data()
 
 if not headers_only
-	cxxshim_dep = dependency('cxxshim', method: 'pkg-config', fallback: ['cxxshim', 'cxxshim_dep'])
+	cxxshim_dep = dependency('cxxshim', fallback: ['cxxshim', 'cxxshim_dep'])
 	libc_deps += cxxshim_dep
 	rtdl_deps += cxxshim_dep
 
 	frigg_dep = dependency(
 		'frigg',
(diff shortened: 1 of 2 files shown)


meson.build
```diff
@@ -37,19 +37,27 @@ internal_conf = configuration_data()
...
 
 if not headers_only
-	cxxshim_dep = subproject('cxxshim').get_variable('cxxshim_dep')
+	cxxshim_dep = dependency('cxxshim', method: 'pkg-config', fallback: ['cxxshim', 'cxxshim_dep'])
 	libc_deps += cxxshim_dep
 	rtdl_deps += cxxshim_dep
-	frigg_dep = subproject('frigg',
-		default_options: ['frigg_no_install=true']).get_variable('frigg_dep')
+
+	# Ideally we'd use the same one-liner as above, but we can't pass default_options that way
```
eli-schwartz, line 44: Why not? The `dependency()` function absolutely does document that you can pass default_options to it and those options will be applied in the case that subproject fallback is done.

eli-schwartz (changes requested):

meson.build
64: Oh, you're absolutely right, I just missed that in the docs

64: Thanks :)

avdgrinten (approved):

avdgrinten (approved):

PR #439, opened 2022-05-30
Port to RISC-V

64:
This PR adds support for the RISC-V architecture (64-bits only for now), and refactors the Linux sysdeps to be architecture independent. The result is that the test suite may be successfully run under `qemu-riscv64`.

Porting the Linux sysdeps to AArch64 won't be too hard now with these changes.

BLOCKED ON:
- [x] https://github.com/managarm/mlibc/pull/444

TODO (THIS PR):
- [x] Dynamic linking
- [x] Threads
- [x] `setjmp` / `longjmp`
- [x] `sigaltstack` test
- [x] Cancellation points
- [x] split into multiple PRs
- [x] Debug `realpath` test

TODO (FUTURE PRs):
- Implement all startupfiles (`crti.o`, `crtn.o`, `Scrt1.o`)
- Implement `pltRelocateStub`
- Implement `sigsetjmp`
- Rename `abis/linux/x86_64/` to `abis/linux/`
- Link to `libgcc` or `compiler-rt` in `meson.build`
- Don't call into a sysdep for `sys_tcb_set` (it only needs to be a syscall on x86_64)
- Run Linux RISC-V tests on CI


abis/linux/x86_64/signal.h
```diff
@@ -255,6 +257,92 @@ typedef struct {
...
 
+typedef struct __ucontext {
+	unsigned long uc_flags;
+	struct __ucontext *uc_link;
+	stack_t uc_stack;
+	mcontext_t uc_mcontext;
+	sigset_t uc_sigmask;
+} ucontext_t;
+
+#elif defined(__riscv) && __riscv_xlen == 64
```
avdgrinten, line 268: Why do we define the RISC-V ABI in `linux/x86_64`?

abis/mlibc/signal.h
```diff
@@ -133,8 +133,9 @@ struct sigaction {
 	void (*sa_sigaction)(int, siginfo_t *, void *);
 };
 
-// TODO: this struct won't work on all arches (for example aarch64) but
-// we don't have an arch specific abi folder for mlibc yet.
+#if defined(__x86_64__) || defined(__aarch64__)
+// TODO: This is wrong for AArch64.
```
avdgrinten, line 137: Why is this file changed? The `mlibc` ABI shouldn't be used for `riscv64-linux` anyway, right?

tests/posix/sigaltstack.c
```diff
@@ -38,7 +38,7 @@ int main() {
 #elif defined(__aarch64__)
 	asm volatile ("mov sp, %0\n\tstp x0, x1, [sp, #-16]!" :: "r"(uint64_t{0}));
 #elif defined(__riscv) && __riscv_xlen == 64
-	asm volatile ("unimp"); // TODO
+	asm volatile ("li sp, 0\n\tsd zero, 0(sp)");
```
avdgrinten, line 41: If you use `\n\t` in inline assembly please also split the inline asm onto two lines.

abis/linux/x86_64/stat.h
```diff
@@ -60,6 +63,30 @@ struct stat {
 	long __unused[3];
 };
 
+#elif defined(__riscv) && __riscv_xlen == 64
```
avdgrinten, line 66: We should not define the RISC-V (or aarch64, see above) ABIs in `linux/x86_64`.

options/lsb/generic/tls.cpp
```diff
@@ -2,11 +2,17 @@
...
+	extern "C" void *__tls_get_addr(struct __abi_tls_entry *entry) {
+		Tcb *tcbPtr = mlibc::get_current_tcb();
+		auto dtvPtr = reinterpret_cast<char *>(tcbPtr->dtvPointers[0]);
+		return reinterpret_cast<void *>(dtvPtr + entry->offset);
+	}
+#else
+	extern "C" void *__tls_get_addr(struct __abi_tls_entry *entry) {
+		return __dlapi_get_tls(entry);
+	}
+#endif
```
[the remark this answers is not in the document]
64: IIRC it uses `entry->object` which hasn't been set up properly (since we don't perform dynamic relocs), in the static case.

abis/mlibc/signal.h
64: 1) I've moved the `ucontext_t` struct from a posix internal header to here (and also defined it in the Linux ABI folder) so that we can `ucontext_t` from different ABIs 2) I've added an `#error` below in case a RISC-V user does accidentally consume this without modification.

options/lsb/generic/tls.cpp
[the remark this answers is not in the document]
avdgrinten: Do relocations exist that we could perform to fix this? This `#ifdef` approach breaks if a statically linked binary `dlopen()`s a shared library at runtime.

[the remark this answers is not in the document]
64: Does dlopen usually work from static binaries? The DSO would have to be linked against libc statically too (or not at all)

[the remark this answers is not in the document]
avdgrinten: I guess it can be made to work but OTOH it might be more reasonable to merge this PR now; we can always worry about this later. Maybe add a comment that the current implementation fails if dlopen() is used from a static binary.

64:
Rebased to fix sigma merge conflict and added the comment about dlopen.

avdgrinten (approved):
LGTM, thank you for the PR!

PR #441, opened 2022-05-31
options/ansi: fix ignored * specifiers in scanf

### LINKED ISSUE #440: scanf doesn’t implement * properly
Specifiers with * are supposed make scanf read the input without consuming a varg, but the current scanf implementation just skips over * without parsing the input.

Kyota-exe:
This PR fixes `scanf` so that it handles the "assignment-suppression character" `*` correctly.
Resolves #440.


Geertiebear (approved):
LGTM, thanks.

PR #442, opened 2022-06-01
tests: fix getservby{name,port}

no92:
About the asserts, quoting the man page: `Port numbers are assigned by the IANA (Internet Assigned Numbers Authority), and their current policy is to assign both TCP and UDP protocols when assigning a port number. Therefore, most entries will have two entries, even for TCP-only services.`

In the loop, the name length did not get reset; this caused a reproducible segfault in local testing for me and @64.

### DIFF
--- a/options/posix/generic/services.cpp
+++ b/options/posix/generic/services.cpp
@@ -46,8 +46,8 @@ static int lookup_serv_file_port(service_result &buf, int proto, int port) {
 
 	char line_buf[129] = {0};
 	char *line = line_buf + 1;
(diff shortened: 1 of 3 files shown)


no92:
Combing through the `/etc/services` of the CI, Debian and Arch, the common denominator of a tcp- or udp-only service seems to be `babel`. Thus, the check for a tcp or udp-only server has been changed from `http` to `babel`.

Geertiebear (approved):
LGTM, thanks.

PR #443, opened 2022-06-02
abis: place {F,W,R,X}_OK in abi-bits

64:
Fixes the ABI for `access` on Linux. Also adds an `access` test (and one for `longjmp` too, because why not)

### DIFF
--- a/sysdeps/managarm/meson.build
+++ b/sysdeps/managarm/meson.build
@@ -66,6 +66,7 @@ endif
 if not no_headers
 	install_headers(
 		'include/abi-bits/abi.h',
+		'include/abi-bits/access.h',
 		'include/abi-bits/auxv.h',
 		'include/abi-bits/seek-whence.h',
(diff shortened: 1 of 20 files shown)


no92 (approved):

ArsenArsen:
lgtm, but squash the last into the first commit

sysdeps/managarm/meson.build
```diff
@@ -88,6 +88,7 @@ if not no_headers
 		'include/abi-bits/nlink_t.h',
 		'include/abi-bits/pid_t.h',
 		'include/abi-bits/uid_t.h',
+		'include/abi-bits/unistd.h',
```
Dennisbonke, line 91: This is wrong, please change this to be `access.h`.

Dennisbonke (changes requested):
Breaks the Managarm build, please fix.

Geertiebear (approved):
LGTM.

PR #444, opened 2022-06-02
rtdl: refactor `getCurrentTcb` and `__abi_tls_entry`

64:
- use `get_current_tcb` instead of `getCurrentTcb`
- move `__abi_tls_entry` to header


64:
ABI check fails because the internal function `getCurrentTcb` is removed, but that shouldn't be an issue.

avdgrinten (approved):
LGTM.

PR #436, opened 2022-05-22
meta: fix minor warnings

no92:

PR #445, opened 2022-06-02
meta: fix all warnings and enable -Werror on CI

64:
Supercedes https://github.com/managarm/mlibc/pull/436.

### DIFF
--- a/ci/bootstrap.yml
+++ b/ci/bootstrap.yml
@@ -69,6 +69,7 @@ packages:
         - '--libdir=lib'
         - '--buildtype=debugoptimized'
         - "-Dbuild_tests=true"
+        - "-Dwerror=true"
         - '@THIS_SOURCE_DIR@'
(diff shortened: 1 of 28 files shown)


Dennisbonke:
LGTM, one question, does this also enable `-Wall` and `-Wextra`? If not, we should do that too.

no92 (approved):

64:
> LGTM, one question, does this also enable `-Wall` and `-Wextra`? If not, we should do that too.

We already have `warning_level=2` in the meson.build

Dennisbonke:
> > LGTM, one question, does this also enable `-Wall` and `-Wextra`? If not, we should do that too.
> 
> We already have `warning_level=2` in the meson.build

Awesome, then this should be good to go from my point of view.

Geertiebear (approved):
LGTM.

PR #447, opened 2022-06-03
options/internal: increase log stack buffer size

64:
This should make some assertion failures more readable.

### DIFF
--- a/options/internal/generic/debug.cpp
+++ b/options/internal/generic/debug.cpp
@@ -5,8 +5,8 @@
 
 namespace mlibc {
 
-frg::stack_buffer_logger<InfoSink> infoLogger;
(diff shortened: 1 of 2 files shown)


64:
ABI diff seems fine.

avdgrinten:
This only reduces the number of line breaks right? Messages should already be printed in full.

64:
> This only reduces the number of line breaks right? Messages should already be printed in full.

yes, but 1) it's harder to read and copy/paste things if they're split too much and 2) it means that messages are less likely to be interleaved with ones from other processes

Geertiebear (approved):
LGTM, can't merge myself tho since it's an ABI break, so @avdgrinten or @qookei should merge it.

PR #448, opened 2022-06-04
rtdl: implement RTLD_LOCAL and RTLD_NOLOAD

### LINKED ISSUE #277: ld.so: Consider objectScope in relocation code
Right now, we only consider globalScope; this leads to incorrect behavior for libraries that have been opened by `dlopen(RTLD_LOCAL)`.

64:
This PR implements `RTLD_LOCAL` and `RTLD_NOLOAD`.

We also add some `rtdl` tests to make sure that we've implemented this correctly.

Fixes #277.


64:
Latest commit fixes #450.

options/rtdl/generic/linker.cpp
```diff
@@ -251,12 +246,31 @@ SharedObject *ObjectRepository::requestObjectAtPath(frg::string_view path, uint6
 
 	_parseDynamic(object);
 
-	_nameMap.insert(path, object);
+	_nameMap.insert(name, object);
```
64, line 249: This change is not correct. I'll fix it when I can.

64:
There are a few things I still need to fix before this should be merged.

avdgrinten:
The last commit's message should (obviously) be fixed, too.

64:
Yeah, this is not ready

64:
Most of the commits here are fairly self-explanatory apart from 53dbeb52b9a4864753cfa8d8f6ffae1b1f80051a. I'll try to give a few details about what changes I've made:

1. `dlopen` now records the return address of the caller. This is used to figure out which object called it (see `findCaller`), and determine which RPATH / RUNPATH to look things up in.
2. Fixed a bug in RPATH processing where the final entry wasn't considered.
3. `object->phdrPointer` is correctly set for all types of objects. Before, we were only setting it for the main executable, meaning that `dl_iterate_phdr` was reporting no phdr entries for those objects.
4. Reworked the way that scopes are handled to support `RTLD_LOCAL`. Instead of promoting everything to the global scope, we: 
    1. Set `object->loadScope` to the scope of the loading object *only when `object` is first loaded*. `object->loadScope` is used to determine the scope of resolutions during relocation processing.
    2. Append `object` to the scope of the loading object (even if `object` was already loaded). This allows symbols in `object` to be visible to the global scope even if it was previously loaded with `RTLD_LOCAL`. We should also add `object`'s dependencies to the scope too, but we leave this unimplemented for now.
    3. Update the symbol resolution algorithm used during relocation processing so that the global scope is searched first, and if the symbol isn't found the local scope is used second. The `rtdl/scope2` test ensures this behaviour works correctly.
5. Delete `object->globalRts` as it's obsoleted by `object->loadScope`.
6. Reworked the way that filenames are interpreted in `DT_NEEDED` entries or `dlopen` calls, to roughly match musl's behaviour. 
    1. If an object is referenced by a pathname (i.e, it contains a '/'), then we interpret that directly and the object to `_nameMap` under that (relative or absolute) pathname. This logic is implemented in `requestLogicWithName`.
    2. If an object is referenced by a 'shortname' (i.e, it *doesn't* contain a '/'), then we interpret that as a search relative to the RPATH / default libdirs etc, and add the object to the `_nameMap` under that shortname. 
    3. The above two points mean that if you do `dlopen("test/libfoo.so")` followed by `dlopen("libfoo.so")` (where both paths resolve to the same underlying file) in the second call the `_nameMap` will not contain any entry for `"libfoo.so"` and the library would be loaded into memory twice. To fix this behaviour (as it is mandated by POSIX), we use two approaches:
        1. Iterate the `_nameMap` to see if any already loaded objects have the same inode and device as requested file. This first requires locating the referenced file; if it is given as a shortname e.g `"libfoo.so"` as above then we will need to process the RPATH etc.
        2. If `sys_stat` isn't available in the dynamic linker to retrieve the inode and device, we will fall back to matching basenames (again by iterating over `_nameMap`), so that `"libfoo.so"` and `"test/libfoo.so"` match the same object.
        3. We leave SONAME handling for a future PR if needed, for simplicity.
        5. Note that this strategy roughly follows musl (they do not look at `SONAME` at all), except they don't implement b. because stat is always available.

64:
I've readded `globalRts` in case we need to use it later for lazy binding.

64:
This PR is basically 'ready', just waiting on some approval on the general approach described in https://github.com/managarm/mlibc/pull/448#issuecomment-1151383060 before I clean it up.

Geertiebear:
This PR should be rebased after merging of #538.

64:
Rebased.

64:
I'm going to split this PR up into smaller chunks tomorrow.

options/rtdl/generic/linker.cpp
```diff
@@ -1131,12 +1137,31 @@ void Loader::submitObject(SharedObject *object) {
...
+			// The object was already loaded into a different scope.
+			// This may happen if e.g we load DSO foo with RTLD_LOCAL, and then
+			// load DSO bar with RTLD_GLOBAL where bar depends on foo.
+			// In this case, foo and all its dependencies must be added to
+			// the global scope. Note that the localScope of foo itself
+			// does not change here (it is fixed at first load, see below).
+
+			_loadScope->appendObject(*it);
+
+			// TODO: Add dependencies to the _loadScope too, in the correct order.
```
Geertiebear, line 1159: Failing silently here could lead to some nasty bugs. Could we maybe warn about it?

options/rtdl/generic/main.cpp
```diff
@@ -335,65 +328,68 @@ void *__dlapi_get_tls(struct __abi_tls_entry *entry) {
...
 		return executableSO;
 
-	if(local)
-		mlibc::infoLogger() << "\e[31mrtdl: RTLD_LOCAL " << file << " is not supported properly\e[39m"
-				<< frg::endlog;
+	// TODO: Thread-safety!
+	auto rts = rtsCounter++;
+
+	// To know which RPATH / RUNPATH to process, we need to know which object called us.
+	auto findOrigin = [&](){
```
Geertiebear, line 348: Why does this need to be a lambda? It also doesn't seem to be used in all of the relevant cases.

options/rtdl/generic/linker.cpp
ArsenArsen: is this that hard to implement even?

64: It's a bit annoying as we have to do a breadth-first traversal of the dependencies, but let me give it a go.

64: I've implemented this now and added a test for it.

avdgrinten:
`_linkBfs` already contains all dependencies, doesn't it?

64:
Going to split this up again.

Dennisbonke (approved):
LGTM

options/rtdl/generic/linker.hpp
```diff
@@ -240,20 +241,29 @@ frg::optional<ObjectSymbol> resolveInObject(SharedObject *object, frg::string_vi
 struct Scope {
 	using ResolveFlags = uint32_t;
 	static inline constexpr ResolveFlags resolveCopy = 1;
+	static inline constexpr ResolveFlags skipGlobalAfterRts = 1 << 1;
+
+	static frg::optional<ObjectSymbol> resolveGlobalOrLocal(Scope &globalScope,
```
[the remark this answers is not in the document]
64: We don't, but it's maybe a little clearer (the same reason that Scope has an `isGlobal` variable, as korona requested). I don't mind either way, though.

Dennisbonke:
With korona's sign off on discord, and with the knowledge that you will be addressing the lifetime issues discussed on discord in a later PR, I think we can merge this.

PR #451, opened 2022-06-05
options/ansi: Implement several of the fenv family functions

Dennisbonke:
Chromium made extensive use of some of these functions, so I decided to implement them like musl does. Fixes some ensures in Chromium.

### DIFF
--- a/options/ansi/generic/fenv-stubs.cpp
+++ b/options/ansi/generic/fenv-stubs.cpp
@@ -2,22 +2,11 @@
 #include <bits/ensure.h>
 #include <fenv.h>
 
-int feclearexcept(int) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
-}
-
-int fegetenv(fenv_t *) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
-}
+// The functions that are not in this file but are defined in the header
+// are implemented like musl does in assembly.
+extern "C" __attribute__((__visibility__("hidden"))) int __fesetround(int);
 
 int fegetexceptflag(fexcept_t *, int) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
-}
-
-int fegetround(void) {
 	__ensure(!"Not implemented");
 	__builtin_unreachable();
 }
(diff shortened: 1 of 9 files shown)


options/ansi/generic/fenv-stubs.cpp
```diff
@@ -2,54 +2,45 @@
...
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
-}
-
-int fegetenv(fenv_t *) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+// The functions that are not in this file but are defined in the header
+// are implemented like musl does in assembly.
+#ifdef __cplusplus
```
Geertiebear, line 7: Why this ifdef?

Geertiebear (changes requested):
These functions should be tested.

no92:
I'm unclear on how these should be tested, TBH, and given that these are taken from musl (and have already been tested with bionic's tests), I don't see _that_ great of a need to do so.

Dennisbonke:
I agree with Leo here. What do you guys think?

avdgrinten:
Why is it hard to test them? Isn't it enough to cause an FPU exception and see if it behaves as expected?

ArsenArsen:
in process floating point tests seem like the easiest thing a libc could be testing for

options/ansi/generic/fenv-stubs.cpp
```diff
@@ -2,54 +2,39 @@
...
-	__builtin_unreachable();
-}
-
-int fegetenv(fenv_t *) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
-}
+// The functions that are not in this file but are defined in the header
+// are implemented like musl does in assembly.
+extern "C" __attribute__((__visibility__("hidden"))) int __fesetround(int);
```
64, line 7: Isn't making the assembly hidden sufficient? Also, you could use C++-style attributes

options/ansi/include/bits/ansi/fenv.h
```diff
@@ -0,0 +1,53 @@
+#ifndef MLIBC_FENV
```
64, line 1: _H

options/ansi/include/bits/ansi/fenv.h
```diff
@@ -0,0 +1,53 @@
...
+#define FE_TONEAREST 0
+#define FE_TOWARDZERO 1
+#define FE_DOWNWARD 2
+#define FE_UPWARD 3
+
+#else
+#error Unknown architecture
+#endif
+
+#endif // MLIBC_FENV
```
64, line 53: _H

tests/ansi/fenv.c
```diff
@@ -0,0 +1,90 @@
+#include <assert.h>
+#include <stdbool.h>
+#include <fenv.h>
+#include <math.h>
+
+#define NO_OPTIMIZE(x) asm volatile("" :: "r,m" (x) : "memory")
```
64, line 6: Dodgy, but I guess as long as the tests keep passing then who cares.

64 (approved):

options/ansi/generic/fenv-stubs.cpp
64: Nvm, disregard. But C++ attribute would be nice

Dennisbonke: Couldn't get it to work, so we can fix that later if need be.

PR #452, opened 2022-06-05
options/posix: Refactor FD_SET and friends to work with wine

Dennisbonke:
This PR refactors FD_SET and friends to be consistent with the way glibc defines them, which makes wine happy during compilation. No functional changes are made.

Abi break counterpart of #374


Geertiebear (approved):
LGTM.

PR #453, opened 2022-06-06
options/ansi: Implement timegm

Dennisbonke:
This PR implements `timegm()` as implemented by Sortix.

### DIFF
--- a/options/ansi/generic/time-stubs.cpp
+++ b/options/ansi/generic/time-stubs.cpp
@@ -275,7 +275,11 @@ size_t strftime(char *__restrict dest, size_t max_size,
 }
 
 size_t wcsftime(wchar_t *__restrict, size_t, const wchar_t *__restrict,
-		const struct tm *__restrict) MLIBC_STUB_BODY
+		const struct tm *__restrict) {
+	mlibc::infoLogger() << "mlibc: wcsftime is a stub" << frg::endlog;
+	return 0;
+}
+
 namespace {
 
 struct tzfile {
--- a/options/ansi/generic/time-stubs.cpp
+++ b/options/ansi/generic/time-stubs.cpp
@@ -1093,7 +1093,19 @@ time_t timelocal(struct tm *) {
 	__builtin_unreachable();
 }
 
-time_t timegm(struct tm *) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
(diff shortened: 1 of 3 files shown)


options/ansi/generic/time-stubs.cpp
```diff
@@ -1089,7 +1092,155 @@ time_t timelocal(struct tm *) {
...
+static const int DAYS_MARCH = 31;
+static const int DAYS_APRIL = 30;
+static const int DAYS_MAY = 31;
+static const int DAYS_JUNE = 30;
+static const int DAYS_JULY = 31;
+static const int DAYS_AUGUST = 31;
+static const int DAYS_SEPTEMBER = 30;
+static const int DAYS_OCTOBER = 31;
+static const int DAYS_NOVEMBER = 30;
+static const int DAYS_DECEMBER = 31;
```
apache-hb, line 1107: might it be preferable to put these in an anonymous namespace rather than static, use constexpr instead of const, and update these to match style elsewhere (camelCase?)

options/ansi/generic/time-stubs.cpp
```diff
@@ -1089,7 +1092,155 @@ time_t timelocal(struct tm *) {
...
+		DAYS_NOVEMBER,
+		DAYS_DECEMBER,
+	};
+
+	for(uint8_t m = 0; m < month; m++) {
+		int month_leaps = get_leap_second(year, m);
+		int month_days = month_days_list[m];
+		int month_seconds = month_days * 24 * 60 * 60 + month_leaps;
+		ret += month_seconds;
+	}
```
apache-hb, line 1236: further up time_t was assigned to directly, cant it be used here to avoid the possibly narrowing implicit cast from int

options/ansi/generic/time-stubs.cpp
```diff
@@ -1089,7 +1092,155 @@ time_t timelocal(struct tm *) {
...
+
+	ret += (time_t)day * 24 * 60 * 60;
+	ret += (time_t)hour * 60 * 60;
+	ret += (time_t)minute * 60;
+	ret += (time_t)second * 1;
```
apache-hb, line 1241: day, hour, minute, and second are all `time_t`, is this cast needed?

Geertiebear (changes requested):
Please add a test for the newly added functionality.

Geertiebear (approved):
LGTM now.

PR #455, opened 2022-06-06
options/linux-headers: Add rt_class_t enum

Dennisbonke:
This PR adds a missing enum to `rtnetlink.h`, which is used by Chromium and parts of the Managarm netstack.


Dennisbonke:
A failure in the `pthread_cond` test is completely unrelated to this change and can be disregarded. As it is just an enum getting added, we can ignore the test failure here.

Geertiebear (approved):
LGTM.

PR #456, opened 2022-06-06
sysdeps/managarm: Stub sys_madvise

Dennisbonke:
Chromium relies on `madvise()` returning 0, which we can only satisfy if we stub the actual sysdep.


Geertiebear (approved):
LGTM.

PR #457, opened 2022-06-06
options/glibc: Stub res_nclose and res_ninit

Dennisbonke:
This PR stubs `res_nclose()` and `res_ninit()`. As far as I'm aware, we can safely leave them as no-ops for now.

### DIFF
--- a/options/glibc/generic/resolv-stubs.cpp
+++ b/options/glibc/generic/resolv-stubs.cpp
@@ -1,5 +1,6 @@
 #include <resolv.h>
 #include <bits/ensure.h>
+#include <mlibc/debug.hpp>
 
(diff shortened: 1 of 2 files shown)


options/glibc/generic/resolv-stubs.cpp
```diff
@@ -16,6 +16,15 @@ int res_init() {
 	return 0;
 }
 
+int res_ninit(res_state) {
```
Geertiebear, line 19: Add a log here to say it's a stub please.

Geertiebear (changes requested):
LGTM, just add a log.

Geertiebear (approved):
LGTM now.

PR #418, opened 2022-05-02
Various additions for QtWebEngine and Chromium

Dennisbonke:
This PR aims to implement missing functions, defines and structs for `QtWebEngine` and `Chromium`. As my first aim was `QtWebEngine`, the branch is named after that, but the current focus really is `Chromium`. At this moment in time, chromium does compile and link, but I expect more additions to be needed that we'll find while running and debugging it. To avoid making this too big, I'd like to get this merged in as is and do the rest in a separate PR.

Related PRs can be found in #458 and #461

PR #458, opened 2022-06-06
Chromium part 2

Dennisbonke:
As requested on #418, I've split the PR into two separate pieces. This PR contains all changes into the `abis` folder or into `options/elf`, which should only be defines and structs. It is my intention that this makes #418 smaller and easier to review.

Sister PRs are #418 and #461 


abis/mlibc/signal.h
```diff
@@ -19,6 +19,7 @@ typedef struct {
 	void *si_addr;
 	int si_status;
 	union sigval si_value;
+	int si_fd;
```
Geertiebear, line 22: This is an ABI break

Geertiebear (approved):
LGTM.

PR #438, opened 2022-05-25
options/glibc: implement missing functionality in getopt_long

no92:
This fixes the issues raised in #170.

PR #460, opened 2022-06-06
tests: run test suite under host libc too

64:
This PR adds support for running the test suite with the host `libc`, which should serve as a useful sanity check that the tests are functioning correctly. We have already found a few bugs in our implementation from this.


tests/glibc/getopt.c
```diff
@@ -12,25 +17,27 @@ int main() {
...
 	int c;
 	c = getopt_long(test_argc, test_argv, shortopts, longopts, NULL);
+	debug(c);
 	assert(c == 'f');
 
 	c = getopt_long(test_argc, test_argv, shortopts, longopts, NULL);
-	assert(!strcmp(optarg, "abc"));
+	debug(c);
 	assert(c == 'b');
+	assert(!strcmp(optarg, "abc"));
```
64, line 40: cc @no92, these changes were necessary to get the `getopt` test to run on `glibc` but it implies that our implementation in `mlibc` is broken.

tests/posix/popen.c
```diff
@@ -2,7 +2,12 @@
 #include <string.h>
 #include <assert.h>
 
+#ifdef USE_HOST_LIBC
 #define TEST_FILE "popen.tmp"
+#else
+#define TEST_FILE "popen-host-libc.tmp"
```
no92, line 8: This is the wrong way around.

tests/glibc/getopt.c
no92: Weird, they work for me on both mlibc and glibc. Also, I remember checking against the man pages, and our mlibc behavior should match that, IIRC.

As far as I can tell, these changes should not really change the behavior anyways other than using concatenated short options.

Also, #438 touches this test and implements missing code paths in `getopt_long`.

64: With the double colon, the man page doesn't say that it should eat the next argument into optarg. What glibc version do you use?

ArsenArsen:
is the build system code here duplicated?

64:
meson doesn't have functions, so yes

ArsenArsen:
it has loops, and arrays

64:
> is the build system code here duplicated?

How's that for a refactor now?

I've also made the testing code use meson's `suite` feature, so that the output looks like this:
```
$ meson test abort
1/2 mlibc:host-libc+posix / abort        EXPECTEDFAIL    0.04s   killed by signal 6 SIGABRT
2/2 mlibc:posix / abort                  EXPECTEDFAIL    0.05s   killed by signal 6 SIGABRT
```

I think this is more useful for people since it automatically runs both the host libc test AND the mlibc test without any extra arguments. Plus one can now easily run all the posix tests with `meson test --suite posix`, or all the host libc tests with `meson test --suite host-libc`.

Geertiebear:
#438 was merged, so this PR can be rebased.

64:
Blocked on a fix for `strtol` triggering UBSan.

tests/ansi/alloc.c
```diff
@@ -18,6 +18,8 @@ int main() {
 	p = aligned_alloc(1, 1);
 	assert(p != NULL);
 
+	// It seems that glibc doesn't report error in these cases.
+#ifndef USE_HOST_LIBC
```
ArsenArsen, line 22: perhaps this should test for glibc?

ArsenArsen (approved):

tests/ansi/alloc.c
```diff
@@ -18,6 +18,8 @@ int main() {
 	p = aligned_alloc(1, 1);
 	assert(p != NULL);
 
+	// It seems that glibc doesn't report error in these cases.
```
Geertiebear, line 21: We should probably submit a bug report then...

Geertiebear (approved):
LGTM, with one comment.

tests/ansi/alloc.c
64: Meh... it'd probably break some programs if they changed that behaviour. We'd still have to work around it here anyway, so we need to decide on a solution.

Geertiebear:
Conflicts on ``tests/meson.build`` right now.

PR #468, opened 2022-06-06
meta: move stubbed libs to dummy-libs folder

64:
Hopefully this should remove some of the confusion that people have when looking at `mlibc` source code for the first time.


Geertiebear (approved):
LGTM.

Issue #469, opened 2022-06-06
pthread_cond test fails sometimes (particularly under heavy load)


64:
e.g
```
 46/122 posix/pthread_cond FAIL 0.20s   killed by signal 4 SIGILL
```
with stderr showing
```
In function pthread_mutex_lock, file ../options/posix/generic/pthread-stubs.cpp:794
__ensure(!"sys_futex_wait() failed") failed
```

PR #470, opened 2022-06-06
options/posix: loop if sys_futex_wait returns EAGAIN

64:
Fixes https://github.com/managarm/mlibc/issues/469.

### DIFF
--- a/options/posix/generic/pthread-stubs.cpp
+++ b/options/posix/generic/pthread-stubs.cpp
@@ -790,7 +790,11 @@ int pthread_mutex_lock(pthread_mutex_t *mutex) {
 
 			// Wait on the futex if the waiters flag is set.
 			if(expected & mutex_waiters_bit) {
-				if(int e = mlibc::sys_futex_wait((int *)&mutex->__mlibc_state, expected, nullptr); e)
+				int e = mlibc::sys_futex_wait((int *)&mutex->__mlibc_state, expected, nullptr);
+
+				// If the wait returns EAGAIN, that means that the mutex_waiters_bit was just unset by
+				// some other thread. In this case, we should loop back around.
+				if (e && e != EAGAIN)
 					__ensure(!"sys_futex_wait() failed");
 
 				// Opportunistically try to take the lock after we wake up.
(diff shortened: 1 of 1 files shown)


options/posix/generic/pthread-stubs.cpp
```diff
@@ -790,7 +790,11 @@ int pthread_mutex_lock(pthread_mutex_t *mutex) {
 
 			// Wait on the futex if the waiters flag is set.
 			if(expected & mutex_waiters_bit) {
-				if(int e = mlibc::sys_futex_wait((int *)&mutex->__mlibc_state, expected, nullptr); e)
+				int e = mlibc::sys_futex_wait((int *)&mutex->__mlibc_state, expected, nullptr);
+
+				// If the wait returns EAGAIN, that means that the mutex_waiters_bit was just unset by
+				// some other thread. In this case, we should loop back around.
+				if (e && e != EAGAIN)
```
ArsenArsen, line 797: Why the refactor to separate out this variable? (also is there no real mechanism for reporting this error out of libc?)

ArsenArsen (changes requested):

options/posix/generic/pthread-stubs.cpp
64: how else would it be written

avdgrinten (approved):
LGTM.

options/posix/generic/pthread-stubs.cpp
avdgrinten: I think this is the right fix. The code is called in a loop and EAGAIN simply means that we need to loop once more because there was a race (the futex state did not match our expectation).

ArsenArsen: just as before: `if (int e = mlibc::sys_futex_wait((int *)&mutex->__mlibc_state, expected, nullptr); e && e != EAGAIN)`

> I think this is the right fix. The code is called in a loop and EAGAIN simply means that we need to loop once more because there was a race (the futex state did not match our expectation).

I mean non-EAGAIN errors, my bad for the confusion.

64: True, just the line is really long that way. I don't think there is anything more we can do with the error here; at least the man page has a list of only a few errors lock can return.

PR #474, opened 2022-06-07
sysdeps: update sys_mkdir to new prototype

no92:


sysdeps/managarm/generic/file.cpp
```diff
@@ -163,7 +163,7 @@ int sys_chroot(const char *path) {
 	return 0;
 }
 
-int sys_mkdir(const char *path) {
+int sys_mkdir(const char *path, mode_t) {
 	return sys_mkdirat(AT_FDCWD, path, 0);
```
64, line 167: We should probably pass the mode parameter to mkdirat here, even though it's currently unused.

Geertiebear (approved):
LGTM but please squash the fixup.

PR #478, opened 2022-06-08
options/ansi: Handle EOF in ungetc properly

Dennisbonke:
This, together with a locking patch in gtk+3 (or a change in the way we lock things in file-io.cpp), allows gtk+3 to run again on Managarm. Many thanks to @64 for the test and fixing ungetc even further.


Geertiebear (approved):
LGTM, thank you.

PR #480, opened 2022-06-10
options/internal: implement ubsan hooks for libc.so and ld.so internal usage

64:
The `ffsl-ffsll` test had to also be fixed because it was crashing with UBSan enabled. We refactor it along the lines of #445.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -223,6 +223,7 @@ internal_sources = [
 	'options/internal/generic/essential.cpp',
 	'options/internal/generic/frigg.cpp',
 	'options/internal/generic/sigset.cpp',
+	'options/internal/generic/ubsan.cpp',
 	'options/internal/gcc/stack_protector.cpp',
 	'options/internal/gcc/guard-abi.cpp',
 	'options/internal/gcc/initfini.cpp',
@@ -266,6 +267,7 @@ rtdl_sources += [
 	'options/internal/generic/ensure.cpp',
 	'options/internal/generic/essential.cpp',
 	'options/internal/generic/frigg.cpp',
+	'options/internal/generic/ubsan.cpp',
 	'options/rtdl/generic/main.cpp',
 	'options/rtdl/generic/linker.cpp',
 	'options/rtdl' / host_machine.cpu_family() / 'runtime.S'
--- a/options/internal/generic/essential.cpp
+++ b/options/internal/generic/essential.cpp
@@ -5,7 +5,8 @@ namespace {
 	// Needed since we cannot declare a templated enum.
 	template<typename T>
 	struct word_helper {
-		enum class [[gnu::may_alias, gnu::aligned(1)]] word_enum : T { };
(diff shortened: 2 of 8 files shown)


meson.build
```diff
@@ -296,10 +298,17 @@ if not headers_only
 			rtdl_deps += libgcc
 		endif
 
+		if get_option('buildtype') == 'debug'
+			sanitize_args = ['-fsanitize=undefined']
+		else
+			sanitize_args = []
+		endif
```
avdgrinten, line 305: Meson has a `-Db_sanitize=undefined` built-in flag for this purpose.

options/internal/generic/essential.cpp
```diff
@@ -13,6 +13,7 @@ namespace {
 
 	template<typename T>
 	[[gnu::always_inline]]
+	[[gnu::no_sanitize("all")]] // Since we use unaligned loads.
```
avdgrinten, line 16: Why are unaligned loads a problem? `word<T>` should have an alignment of 1.

64: We discussed this a bit on Discord. `word<T>` does not have an alignment of 1 despite the attribute. I don't think there's an obvious alternative, so I'm inclined to keep the `no_sanitize` for now.

meson.build
64: This will build the tests with `-fsanitize=undefined` which causes them to break (since there's nothing linked in to provide the ubsan hooks). I'm not sure how to deal with that.

ArsenArsen: We can either override it it for the target or see about including the hooks

64: We could build the hooks as a static library and then link them into both?

ArsenArsen: that would work, yeah, as a static library with install: false or such

avdgrinten: That probably doesn't work since the mlibc namespace has hidden linkage.

For now, we could just pass `override_options: ['b_sanitize=none']`.

64: Yeah, but there's probably a linker flag for that? `--exclude-libs` should hide the symbols defined by the static library

ArsenArsen: > That probably doesn't work since the mlibc namespace has hidden linkage.

yes, but static libs don't do visibility (they're just archives of objects), so I'm pretty sure it would (worth trying anyway, it shouldn't be very hard)

avdgrinten: Well, the issue is that the ubsan hooks use `mlibc::infoLogger` etc. which is not visible outside `libc.so`.

64: Fair enough. I guess we can see about enabling sanitizers for the host libc tests (when that PR is merged), just as another sanity check.

64:
Blocked on https://github.com/managarm/frigg/pull/32.

no92 (approved):

Geertiebear (approved):
LGTM.

PR #485, opened 2022-06-11
sysdeps: add reminder to call __dlapi_enter for static builds to work properly

64:
This should reduce some confusion for new ports.


Geertiebear:
Why not just add the correct call to the sysdeps, instead of this PR?

64:
I think it requires kernel support to pass the stack in, right?

avdgrinten:
No kernel support is needed but a change in `crt*.o` (which is a serious ABI break as it's compiled into each and every program).

Geertiebear (approved):
Then adding a reminder is fine.

PR #489, opened 2022-06-15
sysdeps: remove sigma

64:
As per @thomtl's request.


Dennisbonke:
Why are you removing the sigma port? It's a nice sample port that you can use as a reference. (Quote from Electrode in OSDev #voice).

64:
Thom said we can remove it, plus it's not tested on CI and thus is quite likely to be broken.

PR #496, opened 2022-06-17
options/ansi: fix wrong pointer subtract in fread

64:
This is the wrong way around since `__buffer_ptr >= __unget_ptr` holds as an invariant.


64:
This should probably be backported into the 2.3.0 release, as it breaks any large `fread`s.

avdgrinten (approved):

PR #499, opened 2022-06-17
{options,sysdeps}/linux: implement ptrace

64:
This is part of the mlibc on Linux project.


64:
I think that's a false positive on the ABI checker.

ArsenArsen:
The members did change, maybe this warning would be avoided if the old aliases remained?
Whether this matters also depends on the ucontext consumers

64:
Well, that is true but it shouldn't count as an ABI break

Geertiebear:
Seeing as this is both how glibc and musl define the struct layout to be, I don't think that there will be any downstream API issues.

Geertiebear (approved):
LGTM, merging this with the knowledge that the previous ``mcontext_t`` definition was incorrect.

PR #501, opened 2022-06-17
options/posix: move some termios defines to abi-bits

64:
This is part of the mlibc for Linux project.


options/linux/include/sys/vt.h
```diff
@@ -0,0 +1,6 @@
+#ifndef _VT_H
+#define _VT_H
+
+#include <abi-bits/vt.h>
+
+#endif // _VT_H
```
avdgrinten, line 6: The include guard should be `_SYS_VT_H`.

Geertiebear (approved):
LGTM.

Geertiebear:
Actually, can the fixup commit be rebased? Then this is good to merge.

PR #503, opened 2022-06-17
options/elf: make ELF64_ST_* self-referential macros

64:
This PR is part of the mlibc on Linux project.

### DIFF
--- a/options/elf/include/elf.h
+++ b/options/elf/include/elf.h
@@ -144,6 +144,11 @@ __MLIBC_INLINE_DEFINITION unsigned char ELF64_ST_INFO(unsigned char bind, unsign
 	return (bind << 4) | type;
 }
 
+// Some programs check for definitions of these macros.
+#define ELF64_ST_BIND ELF64_ST_BIND
+#define ELF64_ST_TYPE ELF64_ST_TYPE
+#define ELF64_ST_INFO ELF64_ST_INFO
+
 enum {
 	STB_GLOBAL = 1,
 	STB_WEAK = 2,


Dennisbonke:
This change was needed because `libbacktrace`, part of gcc, was complaining about multiple definitions. @czapek1337 told me that glibc has them as macros, and after changing it gcc didn’t complain.

options/elf/include/elf.h
```diff
@@ -90,15 +94,11 @@ typedef struct {
...
-	return info & 0x0F;
-}
-extern inline unsigned char ELF64_ST_INFO(unsigned char bind, unsigned char type) {
-	return (bind << 4) | type;
-}
+#define ELF64_ST_BIND(info) ((info) >> 4)
+
+#define ELF64_ST_TYPE(info) ((info) & 0x0F)
+
+#define ELF64_ST_INFO(bind, type) (((bind) << 4) | (type))
```
ArsenArsen, line 101: Why? If someone is checking for these being defined, rely on self referential macros instead (see https://github.com/managarm/mlibc/blob/314ff87092b93f7a186de7652a7ed0875f2db2cc/options/posix/include/sys/select.h#L25-L33).

ArsenArsen (changes requested):

options/elf/include/elf.h
64: That might work; cc @czapek1337. Though it's probably better to get rid of the inline functions anyway

ArsenArsen: Why?

64: Because of the weirdness around how inline functions work in ISO / GNU variants of C / C++. The inline functions here might even be broken like (the math functions were) in that the symbols weren't exported from libc.so. In this particular case, using a macro is what other libcs do and I don't see a good reason to depart from that.

ArsenArsen: The only issue wrt this arises on ISO C89, and we've already established a solution.

The reason not to prefer macros is manifold, the most important of which is to prevent double evaluation of parameters, with a close second being the fact that macros are untyped (it also would let us integrate macros into abidiff checks, as macros cannot be abidiffed but inline functions and types can as long as they're invoked at least once).

64: Of course in general macros are worse. But the double argument evaluation does not apply here. And since we'll need macros here anyway the ABI check is hardly useful. But whatever, I'll let someone else weigh in on this.

64:
Sorry for the pings everyone :^)

64:
Closing as the issue was fixed by #507.

PR #504, opened 2022-06-17
abi-bits: add domainname to utsname

64:


ArsenArsen (approved):

Geertiebear (approved):
Conflict on ``ABI_BREAKS.md`` right now.

no92:
Given that this only touched `abis/linux/utsname.h` and this is a fix to be in line with the kernel ABI, why is this an ABI break that is merged into abi-breaks and not into master?

64:
Good question. We should probably merge a similar change into abis/mlibc

PR #506, opened 2022-06-18
rtdl: fix rpaths under 7 characters crashing rtdl

ArsenArsen:
(detected by netbsduser while working on SCAL-UX)

With an rpath of "/", rtdl would fail to substring the rpath, causing an
assertion failure and hence a crash. As the code here implements a
replace functionality, such a substring failure should simply not do a
replace instead.

Depends on https://github.com/managarm/frigg/pull/33 (re-run checks when that gets merged)


64:
The RPATH processing is more broken: it skips the last RPATH entry. This is fixed by #448 (I'll try to split that into a separate PR).

ArsenArsen:
I can cherry pick that later if you'd prefer.

64:
> The RPATH processing is more broken: it skips the last RPATH entry. This is fixed by #448 (I'll try to split that into a separate PR).

Done in #508.

ArsenArsen:
This is ready for merge then?

64:
LGTM.

PR #507, opened 2022-06-18
options: use __MLIBC_INLINE_DEFINITION for header inline functions

64:
As discussed on Discord. This should solve our issues w.r.t `inline` behaviour in C and C++.

See also #503, that PR should be changed to use `__MLIBC_INLINE_DEFINITION` once this is merged.


avdgrinten (approved):
LGTM.

PR #510, opened 2022-06-18
meta: add bors.toml

64:
This should allow us to use bors. Some more steps will be required by someone with perms, see here: https://bors.tech/documentation/getting-started/


avdgrinten:
IMHO we should wait until GitHub's [merge queues](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/merging-a-pull-request-with-a-merge-queue) are generally available.

PR #511, opened 2022-06-19
options/posix: port musl regex engine

### LINKED ISSUE #98: regcomp() unimplemented
The following programs, tested on managarm, require `regcomp()` to properly function

- `file`, `file` works for executables, but for other files, `regcomp()` is called, thus severely limiting it's use.
- `flex`, all programs from `flex` want `regcomp()`. Without it, even `flex --help` fails

64:
Fixes #98.


no92 (approved):

tests/posix/regex.c
```diff
@@ -0,0 +1,30 @@
+#include <regex.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <assert.h>
+ 
+int main(void) {
+	char *testString = "mlibc is the best best best libc";
```
ArsenArsen, line 7: singing hard praise for a WIP project :P

options/posix/musl-generic-regex/glob.c
```diff
@@ -305,3 +306,6 @@ void globfree(glob_t *g)
 	g->gl_pathc = 0;
 	g->gl_pathv = NULL;
 }
+
+// weak_alias(glob, glob64);
+// weak_alias(globfree, globfree64);
```
ArsenArsen, line 311: do we need these?

options/posix/meson.build
```diff
@@ -54,9 +53,19 @@ libc_sources += files(
...
+libc_sublibs += static_library('musl-generic-regex',
+	'musl-generic-regex/fnmatch.c',
+	'musl-generic-regex/glob.c',
+	'musl-generic-regex/regcomp.c',
+	'musl-generic-regex/regerror.c',
+	'musl-generic-regex/regexec.c',
+	'musl-generic-regex/tre-mem.c',
+	pic: true,
+	include_directories: libc_include_dirs,
+	c_args: ['-Wno-unused', '-Wno-implicit', '-Wno-parentheses', '-Wno-sign-compare', '-Wno-attributes', '-Wno-unknown-pragmas', '-Wno-implicit-fallthrough']
```
ArsenArsen, line 68: have we tried fixing these?

ArsenArsen (commented):
Skimming through the code, this seems fine.

options/posix/musl-generic-regex/fnmatch.c
```diff
@@ -18,7 +18,7 @@
 #include <stdlib.h>
 #include <wchar.h>
 #include <wctype.h>
-//#include "locale_impl.h"
+// #include "locale_impl.h"
```
Geertiebear, line 21: Why not just simply remove this?

avdgrinten:
Yes, fixing the warnings would probably be quite annoying.

I wonder if we should simply add `musl` as a subproject and write a Meson generator that copies the files and patches them.

64:
> Yes, fixing the warnings would probably be quite annoying.
> 
> I wonder if we should simply add `musl` as a subproject and write a Meson generator that copies the files and patches them.

I'm not against that, but we should probably do that in a separate PR

PR #513, opened 2022-06-19
options/ansi: use attribute((format)) in stdio.h

64:
The GCC manual states:

> The compiler always (unless -ffreestanding or -fno-builtin is used) checks formats for the standard library functions printf, fprintf, sprintf, scanf, fscanf, sscanf, strftime, vprintf, vfprintf and vsprintf whenever such warnings are requested (using -Wformat), so there is no need to modify the header file stdio.h.

However since we use `-fno-builtin`, our uses of `printf` et al. aren't checked. We add the attributes manually to fix this.


no92 (approved):

Geertiebear (approved):
LGTM

PR #515, opened 2022-06-19
meson: allow static and shared libs to be built simultaneously

### LINKED ISSUE #514: Static and shared libraries cannot be built simultaneously
To fix this, we should probably replace our ad-hoc `static` option with meson's built in `default_library` option (which can be set to a value of 'both').

64:
Fixes #514.

Tests for now only build against the shared libraries when `default_library=both`, but this could probably be fixed in a future PR (maybe once the big refactor of `tests/meson.build` in #460 is merged).

It does make the main meson.build slightly messier, but it's probably worthwhile (and I don't see an obvious way to rewrite it more cleanly).

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -350,7 +350,6 @@ if not headers_only
 			link_whole: libc_sublibs,
 			install: true
 		)
-
 	endif
 
 	library('pthread', 'dummy-libs/libpthread/src/dummy.cpp', install: true)
--- a/meson.build
+++ b/meson.build
@@ -16,6 +16,7 @@ libc_include_dirs = [
 ]
 
 rtdl_sources = [ ]
+rtdl_dso_sources = [ ]
 libc_sources = [ ]
 libc_sublibs = [ ]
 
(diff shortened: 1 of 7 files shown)


meson.build
```diff
@@ -283,76 +279,82 @@ subdir('options/crypt')
...
+			],
+			cpp_args: static_cpp_args + ['-fno-stack-protector'],
+			include_directories: libc_include_dirs,
+			dependencies: libc_deps,
+			link_with: [ld_static_lib],
+			link_whole: [libc_sublibs, ld_static_lib],
+			install: true
+		)
+	endif
+	if library_type == 'shared' or library_type == 'both'
```
ArsenArsen, line 317: `library_type in ['shared', 'both']`

ArsenArsen (changes requested):
Overall, this is pretty good, maybe factor out the common variables (list of sources namely)  between the if static and if shared block?

And a style nit

EDIT: might also be worth adding a dynamic-only test

ArsenArsen (approved):

meson.build
```diff
@@ -303,69 +299,68 @@ if not headers_only
...
+		ansi_sources,
+		lsb_sources,
+	]
+
+	# Our library have different behaviour when built as static and shared libraries.
+	# Hence we need to rebuild the object files with a different define for each mode.
+	if library_type in ['static', 'both']
+		static_cpp_args = [
+			'-DMLIBC_STATIC_BUILD',
+			'-DFRIGG_HAVE_LIBC',
```
Geertiebear, line 320: Is this define even still used? It seems weird to me that it's defined for the static case but not the dynamic case.

Geertiebear (approved):
LGTM in general, one small comment though.

PR #525, opened 2022-06-21
sysdeps/linux: implement syscall macro

### LINKED ISSUE #523: Implement the `syscall` macro on Linux
Given that the alternative is patching everything, this is probably the best option. We might be able to rip this from musl, or maybe it'd be not too hard to implement ourselves.

64:
Fixes #523.


64:
Hmm... the I guess that `__linux__` is still defined when compiling the other sysdeps on CI, so we can't use a simple `#ifdef __linux__` guard to pull in the linux syscall macros. Any ideas? Maybe we should use a different macro that can be detected and set by meson.

avdgrinten:
Can't we pass `-U __linux__` when compiling the other sysdeps?

tests/meson.build
```diff
@@ -79,7 +79,8 @@ posix_fail_test_cases = [
 
 glibc_test_cases = [
 	'getopt',
-	'ffsl-ffsll'
+	'ffsl-ffsll',
+	'syscall',
 ]
```
avdgrinten, line 84: This test only works on Linux, it should probably be called `linux-syscall` and only be added if we are compiling for Linux.

options/posix/include/unistd.h
```diff
@@ -12,6 +12,10 @@
 #include <abi-bits/gid_t.h>
 #include <abi-bits/pid_t.h>
 
+#ifdef __linux__
+#include <mlibc/syscall.h>
+#endif
```
avdgrinten, line 17: `mlibc/` are internal headers that are not installed. This should be in `bits/.

avdgrinten: And the header also needs to be installed.

avdgrinten (approved):

PR #529, opened 2022-06-21
options: use MLIBC_CHECK_OR_ENOSYS macro

64:


no92 (approved):
LGTM, us LFS people can reasonably rebase against this.

options/posix/generic/unistd-stubs.cpp
```diff
@@ -753,11 +642,7 @@ void swab(const void *__restrict, void *__restrict, ssize_t) {
 }
 
 int symlink(const char *target_path, const char *link_path) {
-	if(!mlibc::sys_symlink) {
-		MLIBC_MISSING_SYSDEP();
-		errno = ENOSYS;
-		return -1;
-	}
+	MLIBC_CHECK_OR_ENOSYS(mlibc::sys_symlink);
```
ArsenArsen, line 645: It's redundant to repeat `mlibc::sys_` everywhere. It'd also be best if this macro returned the value of the syscall via statement expressions (and returned out of the function otherwise, so that we can eliminate even that bit of repetition).

options/internal/include/bits/ensure.h
```diff
@@ -21,6 +21,15 @@ void __ensure_warn(const char *assertion, const char *file, unsigned int line,
 #define MLIBC_MISSING_SYSDEP() __ensure_warn("Library function fails due to missing sysdep", \
 		__FILE__, __LINE__, __func__)
 
+#define MLIBC_CHECK_OR_ENOSYS(sysdep) ({ \
+			if (!(sysdep)) { \
+				__ensure_warn("Library function fails due to missing sysdep", \
+					__FILE__, __LINE__, __func__); \
+				errno = ENOSYS; \
+				return -1; \
+			} \
```
ArsenArsen, line 30: Return `sysdep` here.

options/internal/include/bits/ensure.h
```diff
@@ -21,6 +21,15 @@ void __ensure_warn(const char *assertion, const char *file, unsigned int line,
 #define MLIBC_MISSING_SYSDEP() __ensure_warn("Library function fails due to missing sysdep", \
 		__FILE__, __LINE__, __func__)
 
+#define MLIBC_CHECK_OR_ENOSYS(sysdep) ({ \
```
ArsenArsen, line 24: As I mentioned in some random line, it's probably best to make this also not repeat `mlibc::sys_`.

ArsenArsen (changes requested):
It'd be quite nice to use sysdeps like this:

```cpp
auto sysdep = MLIBC_CHECK_OR_ENOSYS(sigprocmask);

if (int e = sysdep(how, set, retrieve); e) { ...
```

(rationale: sysdeps are often near identical, and having to replace the name of a sysdep when duplicating them even twice rather than once can and will introduce bugs)

options/internal/include/bits/ensure.h
64: We'd need to be passed the arguments then, wouldn't we

options/internal/include/bits/ensure.h
64: The downside of that is we won't be able to do `CHECK(mlibc::sys_foo & mlibc::sys_bar)`, and it's a little less obvious to the reader what is happening (+ less greppable). But maybe that's not a big deal and we can split it into separate lines

options/posix/generic/unistd-stubs.cpp
avdgrinten: Returning the function is very ugly, especially for functions that use multiple sysdeps (not every sysdeps is used in only a simple wrapper function)

The `sys_` prefix is redundant but at least it makes grepping easier.

avdgrinten: In general, the less magic that there is within the macro, the better. The previous solution was a few more lines to type but at least it was obvious what's going on.

options/internal/include/bits/ensure.h
64: Ah no you don't mean to call it, just return the function pointer like you wrote above. I'm not sure I agree with that tbh. I think the macro is okay how it is now.

options/posix/generic/unistd-stubs.cpp
ArsenArsen: Fair enough

options/internal/include/bits/ensure.h
ArsenArsen: Well, consider this:

```cpp
ssize_t listxattr(const char *path, char *list, size_t size) {
	MLIBC_CHECK_OR_ENOSYS(mlibc::sys_listxattr);

	ssize_t nread;
	if (int e = mlibc::sys_listxattr(path, list, size, &nread); e) {
		errno = e;
		return -1;
	}
	return nread;
}

ssize_t llistxattr(const char *path, char *list, size_t size) {
	MLIBC_CHECK_OR_ENOSYS(mlibc::sys_llistxattr);

	ssize_t nread;
	if (int e = mlibc::sys_listxattr(path, list, size, &nread); e) {
		errno = e;
		return -1;
	}
	return nread;
}

ssize_t flistxattr(int fd, char *list, size_t size) {
	MLIBC_CHECK_OR_ENOSYS(mlibc::sys_flistxattr);

	ssize_t nread;
	if (int e = mlibc::sys_flistxattr(fd, list, size, &nread); e) {
		errno = e;
		return -1;
	}
	return nread;
}
```

I hope a realistic scenario like this (this happened while I was implementing these) demonstrates my concern. This subtle of an error would've been missed had I merely not also been writing tests that target edge cases (the difference in the two confused functions is that small).

64: I'll let korona decide

Geertiebear: I think I'll vote for the solution as it currently is, mostly because returning the sysdep wouldn't allow for code such as ``MLIBC_CHECK_OR_ENOSYS(mlibc::sys_getpid && mlibc::sys_kill, -1);``, which is a useful usecase. Unless there are some structured binding memes involved...

Geertiebear: For the structured binding idea we could simply have the macro call into a C++ function that generates the structured binding right?

You make a good point on the code duplication, and I understand the concern raise w.r.t. bug resilience.

avdgrinten: For example, if I look at glibc, it's quite difficult to understand how the syscall entry points are generated unless you spend some hours to read the code. This should definitely be avoided.

avdgrinten: The number of sysdep wrappers is bounded so a bit of code duplication does not hurt. Plus, there are subtle differences. For example, compare `madvise()` (which uses errno) vs. `posix_madvise()` (which returns an error code).

ArsenArsen: I'll take it over barely detectable (even at runtime) bugs. This behaviour isn't that unpredictable if you look at the call site of such a return either.

64: IMO the code readability is more important than the small code writabiloty benefit here.

64:
Merging as this has been approved in Discord.

PR #533, opened 2022-06-23
riscv: use far jump to init/fini

64:
This should fix the 'relocation truncated to fit' errors we see on the `abi-break` branch.


avdgrinten (approved):
LGTM. In general, `j` on RISC-V is only suitable for jumps within the same function.

PR #538, opened 2022-06-25
options/rtdl: Use frg::string and fix a nullptr dereference

czapek1337:


64:
LGTM. There's probably a cleaner way to do this, but this way should be good enough for now.

avdgrinten:
Where did we dereference a null pointer before?

64:
The dereference was printing file when null

64:
Maybe we could have frigg print 'nullptr' in that case?

czapek1337:
> Where did we dereference a null pointer before? Or rather, where does frg::String prevent a bug now?

The nullptr deref was coming from valid calls to dlopen with nullptr, I prevented that by moving the print after null check.

Regarding frg::string we had a nasty bug when trying to port Python to mlibc LFS, it was allocating the dlopen path on the heap and then reusing the memory, and we just stored the pointer instead of copying the path so we ended up with broken DSO paths all over the place.

options/rtdl/generic/main.cpp
```diff
@@ -527,7 +528,7 @@ int __dlapi_iterate_phdr(int (*callback)(struct dl_phdr_info *, size_t, void*),
 	for (auto object : globalScope->_objects) {
 		struct dl_phdr_info info;
 		info.dlpi_addr = object->baseAddress;
-		info.dlpi_name = object->name;
+		info.dlpi_name = object->name.data(); // TODO: this pointer will dangle if we modify object->name
```
qookei, line 531: I don't think this is a concern for us; I can't think of a reason why the name we recorded of an already-loaded library would change.

options/rtdl/generic/linker.hpp
```diff
@@ -61,7 +61,7 @@ struct ObjectRepository {
 
 	void _discoverDependencies(SharedObject *object, uint64_t rts);
 
-	frg::hash_map<frg::string_view, SharedObject *,
+	frg::hash_map<frg::string<MemoryAllocator>, SharedObject *,
```
qookei, line 64: Using a full allocated string shouldn't be necessary here if the string view is over the allocated string in `SharedObject`.

options/rtdl/generic/linker.cpp
```diff
@@ -256,7 +256,7 @@ SharedObject *ObjectRepository::requestObjectAtPath(frg::string_view path, uint6
 
 	_parseDynamic(object);
 
-	_nameMap.insert(path, object);
+	_nameMap.insert(frg::string<MemoryAllocator>(path, getAllocator()), object);
```
qookei, line 259: Construct a `string_view` based on `object->name` instead of a full copy of the string here (and other places that insert into `_nameMap`).

qookei (commented):
Looks good apart from the comments.

qookei (approved):
LGTM

options/rtdl/generic/main.cpp
64: The way musl does it (and the way I implement it in #448) they do modify the recorded name if a DSO is loaded first by path (e.g `foo/bar.so`) and then by 'shortname' (`bar.so`), then the recorded name will be modified to `bar.so`. Another case where we'd need to be careful about `.data()` is if we ever implement dlclose and free the backing memory for the object's name. It's true that we don't implement either of these *yet* but it's probably prudent to keep the comment here.

Geertiebear (approved):
LGTM.

PR #539, opened 2022-06-25
Fix inline-emitter not including mlibc-config.h

Dennisbonke:
Also add `extern "C"` to the macros in `sys/sysmacros.h`


64 (approved):
LGTM

PR #540, opened 2022-06-25
options/posix: Implement `buf_file`

### LINKED ISSUE #527: posix: fmemopen() unimplemented
Mount from util-linux requires that to work.

48cf:
Fixes #527


Geertiebear:
This is probably best implemented on top of the already existing ``mem_file``. It seems that ``fmemopen`` is just a non-allocating case of ``open_memstream``.

PR #541, opened 2022-06-25
Various fixes for mlibc LFS part 3

Dennisbonke:
Round 3, mostly defines again, a few structs, two missing headers and several compatibility defines.

### DIFF
--- a/options/posix/include/dirent.h
+++ b/options/posix/include/dirent.h
@@ -28,6 +28,8 @@ struct dirent {
 	char d_name[1024];
 };
 
+#define d_fileno d_ino
+
 struct __mlibc_dir_struct {
 	int __handle;
 	__mlibc_size __ent_next;
--- a/options/posix/include/sys/syslog.h
+++ b/options/posix/include/sys/syslog.h
@@ -1,0 +1,1 @@
+#include <syslog.h>
--- a/options/posix/meson.build
+++ b/options/posix/meson.build
@@ -128,6 +128,7 @@ if not no_headers
 		'include/sys/utsname.h',
 		'include/sys/vfs.h',
 		'include/sys/wait.h',
+		'include/sys/syslog.h',
 		subdir: 'sys'
 	)
 	install_headers(


options/glibc/include/sys/kd.h
```diff
@@ -0,0 +1,17 @@
+#ifndef _SYS_KD_H
+#define _SYS_KD_H
+
+/* Make sure the <linux/types.h> header is not loaded.  */
```
64, line 4: Is this file copied from musl? The spacing should also be fixed to match mlibc's style.

options/posix/include/sys/syslog.h
```diff
@@ -0,0 +1 @@
+#include <syslog.h>
```
64, line 1: This file should have a newline at the end

options/posix/include/dirent.h
```diff
@@ -28,6 +28,8 @@ struct dirent {
 	char d_name[1024];
 };
 
+#define d_fileno d_ino
```
64, line 31: Why is this needed? Seems like a bit of a hack

options/elf/include/elf.h
```diff
@@ -144,6 +176,17 @@ __MLIBC_INLINE_DEFINITION unsigned char ELF64_ST_INFO(unsigned char bind, unsign
 	return (bind << 4) | type;
 }
 
+typedef struct {
+  Elf64_Half si_boundto;
+  Elf64_Half si_flags;
+} Elf64_Syminfo;
+
+#define ELF32_ST_BIND(val)		(((unsigned char) (val)) >> 4)
```
64, line 184: These should probably be made inline to match `ELF64_ST_*`.

64 (changes requested):

options/posix/include/dirent.h
Dennisbonke: It's a compatibility hack, used by `strace`. See [here](https://github.com/bminor/musl/blame/cfdfd5ea3ce14c6abf7fb22a531f3d99518b5a1b/include/dirent.h#L22) for what musl thought when adding it.

options/glibc/include/sys/kd.h
Dennisbonke: Based on indeed, I'll fix the spacing.

options/elf/include/elf.h
Dennisbonke: No longer in this PR, moved to #547

options/posix/include/sys/syslog.h
```diff
@@ -0,0 +1,2 @@
...
+
```
64, line 2: There is one newline too many :^) otherwise LGTM

options/linux/include/sys/mount.h
```diff
@@ -13,6 +13,8 @@ extern "C" {
 #define MS_SYNCHRONOUS 32
 #define MS_NOATIME 64
 
+#define MNT_FORCE 1
```
ArsenArsen, line 16: is this ABI?

Dennisbonke: As far as I know it isn't (and if it is, it matches Linux now).

ArsenArsen: this isn't consumed by the libc, so it is OS ABI

64: > As far as I know it isn't (and if it is, it matches Linux now).

Arsen meant that the change should be in `abis`, not that the change is an ABI break

64 (approved):

options/glibc/include/sys/kd.h
Geertiebear: What is this actually used for?

Dennisbonke: As far as I'm aware, this is just some glibc compatibility header.

Geertiebear: I meant the "Make sure linux/types.h is not loaded". What is that used for? Our ``linux/kd.h`` does not have any inclusion of ``linux/types.h``.

abis/mlibc/mount.h
```diff
@@ -0,0 +1,6 @@
+#ifndef _ABIBITS_MOUNT_H
+#define _ABIBITS_MOUNT_H
+
+#define MNT_FORCE 1
+
+#endif // _ABIBITS_MOUNT_H
```
avdgrinten, line 6: Why does mlibc get its own ABI even though it matches the Linux ABI anyway?

avdgrinten (changes requested):

abis/mlibc/mount.h
Geertiebear: For consistency's sake I suppose.

options/glibc/include/sys/kd.h
```diff
@@ -0,0 +1,17 @@
...
+#define _LINUX_TYPES_H 1
+#define __undef_LINUX_TYPES_H
+#endif
+
+#include <linux/kd.h>
+
+#ifdef __undef_LINUX_TYPES_H
+#undef _LINUX_TYPES_H
+#undef __undef_LINUX_TYPES_H
+#endif
```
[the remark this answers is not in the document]
ArsenArsen: It's also not necessary, and should be in the Linux option I think. See: https://github.com/ArsenArsen/mlibc/commit/27d1035f368373a07641cbc532e3208cf6b77dd1

abis/mlibc/mount.h
avdgrinten: That doesn't make sense, sysdeps can just link to `abis/linux` instead of `abis/mlibc`.

Dennisbonke:
Dropped the two commits giving issues, I'll split those out.

64 (approved):

Geertiebear (approved):
Since the PR slimme down a bunch, the left over changes are good to merge

PR #543, opened 2022-06-25
Various fixes for mlibc LFS part 2

Dennisbonke:
Some more fixes and defines, a compatibility header, some legacy types and an update to `struct res_state`.


options/bsd/include/arpa/nameser_compat.h
```diff
@@ -0,0 +1 @@
+#include <arpa/nameser.h>
```
64, line 1: missing newline

abis/linux/limits.h
```diff
@@ -2,5 +2,6 @@
 #define _ABIBITS_LIMITS_H
 
 #define IOV_MAX 1024
+#define NAME_MAX 255
```
64, line 5: Conflicts with #558.

options/glibc/include/net/ethernet.h
```diff
@@ -2,11 +2,16 @@
...
 #include <stdint.h>
+#include <mlibc-config.h>
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
+#ifdef __MLIBC_LINUX_OPTION
+#	include <linux/if_ether.h>
+#endif
```
64, line 13: This is a kernel header, so I don't know if this should be hidden behind `__MLIBC_LINUX_OPTION`? Maybe it needs to be an ABI header or hidden behind `#ifdef __linux__`, not sure which.

64 (changes requested):

options/glibc/include/net/ethernet.h
ArsenArsen: This implementation is (de-facto) "mandated" by Linux, hence my choice to choose based on that, though, this should `#define __UAPI_DEF_ETHHDR 1` before the include, really, unless we want to reimplement it (as musl chooses to for instance). While which one of the two is used is mostly bikeshedding (there's no real standard on this), this forward would be useful on non-Linux systems.

Alternatively, we can just bring out the definitions into mlibc. It's only a struct and a bunch of ethtype defines (and this is what musl does).

abis/linux/limits.h
Dennisbonke: I'll drop it from #558

options/glibc/include/net/ethernet.h
Geertiebear: So does this current implementation actually work? You say we should do ``#define __UAPI_DEF_ETHHDR 1`` before the include, so it's unclear to me whether this solution right now is sufficient.

ArsenArsen: yeah, for pedantics pretty much, the header defaults it to 1:

```c
/* SPDX-License-Identifier: GPL-2.0+ WITH Linux-syscall-note */

/* allow libcs like musl to deactivate this, glibc does not implement this. */
#ifndef __UAPI_DEF_ETHHDR
#define __UAPI_DEF_ETHHDR		1
#endif

#if __UAPI_DEF_ETHHDR
struct ethhdr {
	unsigned char	h_dest[ETH_ALEN];	/* destination eth addr	*/
	unsigned char	h_source[ETH_ALEN];	/* source ether addr	*/
	__be16		h_proto;		/* packet type ID field	*/
} __attribute__((packed));
#endif
```

avdgrinten: How does musl handle this?

Dennisbonke: Glibc can use cheats, as the kernel has hardcoded checks for glibc in the libc compat header. Musl doesn’t have that and has to do it themselves. We can’t compare with glibc unless we either claim we’re glibc (stupid) or get a fix into the Linux kernel (have fun).

Geertiebear (approved):
This conversation seems resolved so I will merge this.

PR #544, opened 2022-06-25
abis: Define PACKET_HOST

Dennisbonke:
This PR adds a bunch of missing constants into the Linux ABI. It also stubs 2 functions and adds 2 new sysdeps.

### DIFF
--- a/options/linux/include/netpacket/packet.h
+++ b/options/linux/include/netpacket/packet.h
@@ -1,6 +1,8 @@
 #ifndef _NETPACKET_PACKET_H
 #define _NETPACKET_PACKET_H
 
+#include <abi-bits/packet.h>
+
 struct sockaddr_ll {
 	unsigned short int sll_family;
 	unsigned short int sll_protocol;
--- a/abis/linux/packet.h
+++ b/abis/linux/packet.h
@@ -1,0 +1,6 @@
+#ifndef _ABIBITS_PACKET_H
+#define _ABIBITS_PACKET_H
+
+#define PACKET_HOST 0
+
+#endif // _ABIBITS_PACKET_H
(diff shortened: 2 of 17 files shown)


avdgrinten:
Huge PRs with lots of unrelated commits are quite annoying to review; it'd be better to group commits into semantically related changes.

Dennisbonke:
> Huge PRs with lots of unrelated commits are quite annoying to review; it'd be better to group commits into semantically related changes.

If requested, I can do that. However, that will lead to a _lot_ of PRs as most commits aren't really related. I'll split out part 1, 2 and 3 shortly.

avdgrinten:
Lots of PRs are not an issue. A small PR and be reviewed + merged in 5min, a long one takes weeks.

options/linux/include/netpacket/packet.h
```diff
@@ -1,6 +1,8 @@
 #ifndef _NETPACKET_PACKET_H
 #define _NETPACKET_PACKET_H
 
+#define PACKET_HOST 0
```
64, line 4: LGTM now that this PR is only a single line change :^)

64 (approved):

options/linux/include/netpacket/packet.h
ArsenArsen: Isn't this an ABI thing?

Dennisbonke: If it is, it matches Linux so it should be good.

64: I think Arsen means that it should live in `abis/`

64 (changes requested):
As per the above comments, this should live in `abis/`.

Geertiebear (approved):
LGTM.

PR #546, opened 2022-06-25
options/posix: Define CREPRINT and CDISCARD

Dennisbonke:
Split off from #541 


64 (approved):
LGTM (matches musl).

PR #549, opened 2022-06-25
abis: Split off EPOLL_CLOEXEC into it's own abi header

Dennisbonke:
Split off from #541 

### DIFF
--- a/options/linux/include/sys/epoll.h
+++ b/options/linux/include/sys/epoll.h
@@ -3,9 +3,7 @@
 
 #include <stdint.h>
 #include <abi-bits/signal.h>
-
-#define EPOLL_CLOEXEC 1
-#define EPOLL_NONBLOCK 2
+#include <abi-bits/epoll.h>
 
 // These constants match the Linux definitions.
 #define EPOLLIN 0x001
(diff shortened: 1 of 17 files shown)


options/linux/include/sys/epoll.h
```diff
@@ -3,8 +3,8 @@
 
 #include <stdint.h>
 #include <abi-bits/signal.h>
+#include <abi-bits/epoll.h>
 
-#define EPOLL_CLOEXEC 1
 #define EPOLL_NONBLOCK 2
```
avdgrinten, line 8: Shouldn't we also move `EPOLL_NONBLOCK` at the same time?

Dennisbonke: I'm unsure, I don't see `EPOLL_NONBLOCK` on my host so I can't tell if the value is correct.

avdgrinten: Seems to be a glibc artifact: https://sourceware.org/bugzilla/show_bug.cgi?id=14985
We should probably remove it entirely.

Dennisbonke: Will do

64:
Conflicts.

64 (approved):
LGTM

PR #550, opened 2022-06-25
options/linux: Move the scsi headers to the linux option

Dennisbonke:
Split off from #541 


avdgrinten (approved):

avdgrinten:
Maybe add a comment to the `meson.build` to state that these headers are indeed provided by libc on Linux (which is quite stupid).

64:
Merge conflict on `scsi.h`

64 (approved):

PR #552, opened 2022-06-25
options/posix: Define more legacy types in use by acl

Dennisbonke:
Split off from #543 


64 (approved):
LGTM.

PR #553, opened 2022-06-25
options/glibc: Add more members to struct res_state

Dennisbonke:
Split off from #543 

### DIFF
--- a/options/glibc/include/resolv.h
+++ b/options/glibc/include/resolv.h
@@ -34,6 +34,18 @@ typedef struct __res_state {
 	struct sockaddr_in nsaddr_list[MAXNS];
 	char *dnsrch[MAXDNSRCH + 1];
 	char defdname[256];
+	union {
+		char pad[52];
+		struct {
+			uint16_t nscount;
+			uint16_t nsmap[MAXNS];
+			int nssocks[MAXNS];
+			uint16_t nscount6;
+			uint16_t nsinit;
+			struct sockaddr_in6	*nsaddrs[MAXNS];
+			unsigned int _initstamp[2];
+		} _ext;
+	} _u;
 } *res_state;
 struct __res_state *__res_state(void);
 #define _res (*__res_state())


avdgrinten (commented):
This is an ABI break.

avdgrinten:
Why is that struct even needed? Do we use it? Looks like an internal implementation detail.

Dennisbonke:
> This is an ABI break.

Discussed before, it was fine back then. Why wouldn't it be fine now?
See [here](https://github.com/managarm/mlibc/commit/c45dd47a397c6943a9ad0483374f8fa425b8d03d)

Dennisbonke:
> Why is that struct even needed? Do we use it? Looks like an internal implementation detail.

Yeah it's old and not used according to musl, but some programs rely on it. I'm basically following musl here.

avdgrinten:
Which programs rely on this?

Dennisbonke:
> Which programs rely on this?

`glib` and `Qt6` come to mind, the latter only on Linux it seems.

64 (changes requested):
~~This should have a comment in `ABI_BREAKS.md`.~~

64:
Oh sorry, if we decided that this doesn't require a formal ABI break, then can it be merged?

Dennisbonke:
> Oh sorry, if we decided that this doesn't require a formal ABI break, then can it be merged?

Yeah this is ready to go from my end.

Geertiebear (approved):
LGTM, can you rebase this so the vinix sysdep check passes?

Dennisbonke:
> yeah, this isn't an abi break (the struct is only ever accessed via pointer and we are extending it), however, I'm unsure about whether we should expose it if we don't use it, since it might force a specific resolver implementation on us later. We should deliberate more on how this is intended to be used
> 
> EDIT: relevant page: `resolver(3)` and [sourceware.org/git/?p=glibc.git;a=blob;f=resolv/README;h=514e9bb617e710f16126c1474561965a2b35653d;hb=HEAD](https://sourceware.org/git/?p=glibc.git;a=blob;f=resolv/README;h=514e9bb617e710f16126c1474561965a2b35653d;hb=HEAD) the latter appears to imply that this is BINDs libresolv, does Qt have a way to use that?

Unsure, but musl does expose it, noting that it's only to satisfy broken programs and that they themselves don't use it. See [here](https://github.com/bminor/musl/blob/cfdfd5ea3ce14c6abf7fb22a531f3d99518b5a1b/include/resolv.h#L25-L62).

ArsenArsen:
can we fix programs instead? having a struct like this one that will intentionally never get implemented seems silly, at least as a default (i.e. without a `-DMLIBC_ADD_RES_HACK` cppflag or such)

Dennisbonke:
> can we fix programs instead? having a struct like this one that will intentionally never get implemented seems silly, at least as a default (i.e. without a `-DMLIBC_ADD_RES_HACK` cppflag or such)

While we might be able to do that, I prefer to go the musl route, as it is consistent with what we have been doing, and matches other libc's, and reduces the patch burden on us.

64 (approved):

PR #555, opened 2022-06-25
abis/linux: Define _NSIG

Dennisbonke:
Split off from #544 


avdgrinten (commented):
If the ABIs provide `NSIG` we can move `#define _NSIG NSIG` to a public header.

Dennisbonke:
> If the ABIs provide `NSIG` we can move `#define _NSIG NSIG` to a public header.

True, I'll move it.

64 (approved):

PR #558, opened 2022-06-25
abis/linux: Define LOGIN_NAME_MAX

Dennisbonke:
Split off from #544

### DIFF
--- a/abis/linux/limits.h
+++ b/abis/linux/limits.h
@@ -2,5 +2,6 @@
 #define _ABIBITS_LIMITS_H
 
 #define IOV_MAX 1024
+#define LOGIN_NAME_MAX 256
 
 #endif //_ABIBITS_LIMITS_H


abis/linux/limits.h
```diff
@@ -2,5 +2,7 @@
 #define _ABIBITS_LIMITS_H
 
 #define IOV_MAX 1024
+#define NAME_MAX 255
```
64, line 5: The commit message / PR name is wrong, since it adds `NAME_MAX` too. Though we already define `NAME_MAX` in `ansi/limits.h`, so we shouldn't define it here too. Perhaps that + a few of the other constants there should be moved here.

64 (changes requested):

abis/linux/limits.h
ArsenArsen: fwiw, name_max is the system-specific variant. See also: https://github.com/ArsenArsen/mlibc/commit/6d62dbce05ffb6353611a027650784c10f6c8253

64 (approved):
LGTM

PR #560, opened 2022-06-25
options/posix: Define quad_t and u_quad_t

Dennisbonke:
Split off from #544

### DIFF
--- a/options/posix/include/sys/types.h
+++ b/options/posix/include/sys/types.h
@@ -43,5 +43,9 @@ typedef uint16_t u_int16_t;
 typedef uint32_t u_int32_t;
 typedef uint64_t u_int64_t;
 
+// BSD extensions
+typedef int64_t quad_t;
+typedef uint64_t u_quad_t;
+
 #endif // _SYS_TYPES_H
 


64:
Conflicts.

ArsenArsen (changes requested):
This definition is also not necessarily accurate, from OpenBSD:
```c
/* quads, deprecated in favor of 64 bit int types */
typedef	__int64_t	quad_t;
typedef	__uint64_t	u_quad_t;
```

These appear to also be BSD specifics? I can't find anything about it in POSIX.

Dennisbonke:
> This definition is also not necessarily accurate, from OpenBSD:
> 
> ```c
> /* quads, deprecated in favor of 64 bit int types */
> typedef	__int64_t	quad_t;
> typedef	__uint64_t	u_quad_t;
> ```
> 
> These appear to also be BSD specifics? I can't find anything about it in POSIX.

While the BSD stuff might be accurate, this is how musl and glibc define them.

ArsenArsen:
Sure, but I reckon BSD is the authoritative source on BSD extensions

Dennisbonke:
> Sure, but I reckon BSD is the authoritative source on BSD extensions

I'll change them.

avdgrinten:
`musl` defines them as (`unsigned`) `long long`. We can probably use `__mlibc_int64` / `__mlibc_uint64`.

avdgrinten:
Also, add a comment to state that these are BSD extensions.

Dennisbonke:
> musl defines them as (`unsigned`) `long long`. We can probably use `__mlibc_int64` / `__mlibc_uint64`.

We already have stdint.h in that file, so I used `int64_t` / `uint64_t` if that is alright.

ArsenArsen (approved):
(u)int64_t is defined at this point in the header, so this is fine

avdgrinten (approved):

Geertiebear:
This can be rebased to fix the tests.

PR #561, opened 2022-06-25
options/posix: Stub get and setdomainname

Dennisbonke:
Split off from #544

### DIFF
--- a/options/posix/generic/unistd-stubs.cpp
+++ b/options/posix/generic/unistd-stubs.cpp
@@ -1140,6 +1140,16 @@ int setresgid(gid_t rgid, gid_t egid, gid_t sgid) {
 	return 0;
 }
 
+int getdomainname(char *, size_t) {
+	__ensure(!"Not implemented");
+	__builtin_unreachable();
+}
+
+int setdomainname(const char *, size_t) {
+	__ensure(!"Not implemented");
+	__builtin_unreachable();
+}
+
 #ifdef __MLIBC_CRYPT_OPTION
 void encrypt(char[64], int) {
 	__ensure(!"Not implemented");
--- a/options/posix/include/unistd.h
+++ b/options/posix/include/unistd.h
@@ -252,6 +252,10 @@ int pipe2(int *pipefd, int flags);
 int setresuid(uid_t ruid, uid_t euid, uid_t suid);
 int setresgid(gid_t rgid, gid_t egid, gid_t sgid);
 
+/* Glibc extensions. */
+int getdomainname(char *name, size_t len);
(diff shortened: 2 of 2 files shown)


avdgrinten (approved):

64:
These should probably live in `options/glibc` instead, right? But that'd probably require some other changes anyway, so maybe we should make an issue for that and leave it to a future PR.

ArsenArsen (changes requested):
Correct, but it's silly to stub this.

```
       On most Linux architectures (including x86), there is no getdomainname() system call; instead, glibc implements getdomainname() as a  library  func‐
       tion that returns a copy of the domainname field returned from a call to uname(2).
```
... and `setdomainname` is a syscall.

options/posix/generic/unistd-stubs.cpp
```diff
@@ -1122,6 +1122,7 @@ char *ctermid(char *s) {
 	return s ? strcpy(s, "/dev/tty") : const_cast<char *>("/dev/tty");
 }
 
+<<<<<<< HEAD
```
ArsenArsen, line 1125: Merge markers

ArsenArsen (changes requested):
The combined implementation for these two functions and their Linux sysdeps is under 50LoC, boilerplate included

Dennisbonke:
> 

Unfortunately we don't have domainname in the mlibc abi, so that is blocked on abi-break. I move to implement this at a later point, and merge the stubs instead for now/

Dennisbonke:
> Didn't we just merge an abi-break because a struct was incorrect on Linux?

Yes, so the Linux abi is fine, but the mlibc abi is not fine until abi-break is merged into master. I don't want to wait on that.

ArsenArsen:
The struct access would be inside sysdeps to accomodate systems that have a getdomainname syscall anyway, and hence does not rely on the mlibc ABI on Linux

Geertiebear:
I don't think an implementation of these is required to merge this PR, that can always be done in a future PR. But the merge marker needs to be removed for it to build of course.

64 (approved):

options/posix/include/unistd.h
```diff
@@ -252,6 +252,10 @@ int pipe2(int *pipefd, int flags);
 int setresuid(uid_t ruid, uid_t euid, uid_t suid);
 int setresgid(gid_t rgid, gid_t egid, gid_t sgid);
 
+// Glibc extensions
```
ArsenArsen, line 255: line comment

ArsenArsen (changes requested):

Geertiebear (approved):
LGTM.

PR #563, opened 2022-06-25
options/linux: Include certain linux specific headers in sys/syscall.h if we're linux

Dennisbonke:
After #525 was merged, programs still refused to use the syscall macro, claiming it couldn't be found. This fixes that problem.

Part of mlibc LFS project.

### DIFF
--- a/options/linux/include/sys/syscall.h
+++ b/options/linux/include/sys/syscall.h
@@ -2,5 +2,9 @@
 #define _SYS_SYSCALL_H
 
 // OpenSSL wants this header to exist
+#ifdef __linux__
+#include <abi-bits/arch-syscall.h>
+#include <asm/unistd.h>
+#endif
 
 #endif // _SYS_SYSCALL_H


64 (approved):
LGTM

options/linux/include/sys/syscall.h
```diff
@@ -2,5 +2,9 @@
 #define _SYS_SYSCALL_H
 
 // OpenSSL wants this header to exist
+#ifdef __linux__
+#include <abi-bits/arch-syscall.h>
+#include <asm/unistd.h>
+#endif
```
avdgrinten, line 8: The comment is outdated, right? Also, `asm/unistd.h` is provided by the Linux kernel headers, right?

64: `asm/unistd.h` provides syscall numbers in the form e.g `__NR_write` while `sys/syscall.h` provides syscall numbers in the form `SYS_write`. Different applications require different ones, but yes the kernel provides `asm/unistd.h`.

64:
Hmm, actually I don't think this is the correct place. This change should probably live in `sysdeps/linux/include/sys/syscall.h`.

Alternatively, perhaps we should move `sysdeps/linux/include/sys/syscall.h`'s contents here and surround it by a large `#ifdef __linux__`. ATM it seems that both of these headers can't coexist.

avdgrinten:
Instead of putting a lot of code into `#ifdef`, put it into a `bits/` header.

64:
Yeah that's true. It's just getting slightly messy since we have one header which provides the syscall macro itself, and another header which provides the syscall numbers. The latter is (currently) what `sys/syscall.h` is, and the former is what `bits/syscall.h` is. So we might have to rename the headers a bit.

64 (changes requested):
Requesting changes as per discussion above.

64:
Merge conflict hee (+ still changes needed)

Dennisbonke:
Closing in favor of #641

PR #565, opened 2022-06-25
tests: use PIE where available

64:


avdgrinten:
It might be good to run the tests twice: once with PIE and once without. That should be doable simply by adjusting the build matrix, right?

64:
The way this PR is set up, the pie tests will get run when we're building a shared library, and non-pie tests when we're building a static library. Wouldn't that be sufficient?

64:
Had to work around a pretty nasty linker bug by ignoring `DT_TEXTREL` on `RISC-V`. The linker seems to interpret the `R_RISCV_NONE` relocs that result from the bug as relocations against a read-only section of the elf, which causes it to slap `DT_TEXTREL` / `DF_TEXTREL` on the binary when it's not required.

avdgrinten:
> The way this PR is set up, the pie tests will get run when we're building a shared library, and non-pie tests when we're building a static library. Wouldn't that be sufficient?

In principle, there is also `-static -pie`, right? Although that should still link against `Scrt1.o`.

64:
Well yes, that's true (though it'd use `rcrt1.o`). It just sits a bit wrong with me that we'd be doubling the CI time to effectively test a few lines of hand written assembly. If we want to test both pie and non-pie, perhaps we can change the CI recipe so that pie is used when building both static and shared libs simultaneously, otherwise non-pie.

avdgrinten (approved):
LGTM. Can this be merged?

PR #580, opened 2022-06-29
abi-bits: add domainname to utsname in mlibc abi

64:
As per the discussion in #504.

### DIFF
--- a/ABI_BREAKS.md
+++ b/ABI_BREAKS.md
@@ -6,4 +6,4 @@ This document lists the ABI breaks that were made in each mlibc major version.
 
 - [#452](https://github.com/managarm/mlibc/pull/452): The functions `FD_{CLR,ISSET,SET,ZERO}` were renamed to `__FD_{CLR,ISSET,SET,ZERO}` and replaced by macros to match Wine's assumptions.
 - [#511](https://github.com/managarm/mlibc/pull/511): Musl's regex engine was added, implementing `regcomp` and `regexec`. This required some changes to the `regex_t` struct.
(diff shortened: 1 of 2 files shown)


ABI_BREAKS.md
```diff
@@ -6,4 +6,4 @@ This document lists the ABI breaks that were made in each mlibc major version.
 
 - [#452](https://github.com/managarm/mlibc/pull/452): The functions `FD_{CLR,ISSET,SET,ZERO}` were renamed to `__FD_{CLR,ISSET,SET,ZERO}` and replaced by macros to match Wine's assumptions.
 - [#511](https://github.com/managarm/mlibc/pull/511): Musl's regex engine was added, implementing `regcomp` and `regexec`. This required some changes to the `regex_t` struct.
-- [#504](https://github.com/managarm/mlibc/pull/504): A `domainname` member was added to `struct utsname`, which is a glibc extension.
+- [#580](https://github.com/managarm/mlibc/pull/504): A `domainname` member was added to `struct utsname`, which is a glibc extension.
```
Geertiebear, line 9: But the old entry is still true? We should just add two entries here.

Geertiebear (changes requested):

ABI_BREAKS.md
64: The old entry changed the Linux ABI, which isn't really something that downstream users will care about

Geertiebear: There should still be two entries. We cannot assume that no user is interested in the changelog of linux ABI changes.

64: Fair point.

PR #581, opened 2022-06-29
Add remaining sysdeps from LFS

no92:


options/linux/generic/sys-prctl-stubs.cpp
```diff
@@ -1,12 +1,32 @@
...
+		MLIBC_MISSING_SYSDEP();
+		errno = ENOSYS;
+		return -1;
+	}
+
+	unsigned long x[4];
+	va_list ap;
+	va_start(ap, op);
+	for(int i=0; i<4; i++)
+		x[i] = va_arg(ap, unsigned long);
```
64, line 22: Why do we not simply make the function accept 4 longs?

options/linux/generic/sys-prctl-stubs.cpp
```diff
@@ -1,12 +1,32 @@
...
+	if(!mlibc::sys_prctl) {
+		MLIBC_MISSING_SYSDEP();
+		errno = ENOSYS;
+		return -1;
+	}
+
+	unsigned long x[4];
+	va_list ap;
+	va_start(ap, op);
+	for(int i=0; i<4; i++)
```
64, line 21: The formatting should be fixed here.

options/linux/generic/sys-prctl-stubs.cpp
```diff
@@ -1,12 +1,32 @@
...
 #include <sys/prctl.h>
 
 #include <mlibc/debug.hpp>
 
-int prctl(int, ...) {
-	mlibc::infoLogger() << "\e[31mmlibc: prctl() is not implemented correctly\e[39m"
-			<< frg::endlog;
+#include "mlibc/linux-sysdeps.hpp"
+
+int prctl(int op, ...) {
```
64, line 11: Why do we use varargs and not simply take 4 longs?

sysdeps/linux/generic/sysdeps.cpp
```diff
@@ -625,14 +718,206 @@ int sys_reboot(int cmd) {
...
+int sys_eventfd_create(unsigned int initval, int flags, int *fd) {
+	auto ret = do_syscall(SYS_eventfd2, initval, flags);
+	if (int e = sc_error(ret); e)
+		return e;
+	*fd = sc_int_result<int>(ret);
+	return 0;
+}
+
+int sys_signalfd_create(const sigset_t *masks, int flags, int *fd) {
+	auto ret = do_syscall(SYS_signalfd4, *fd, masks, 65/8, flags);
```
64, line 782: Did you mean to use `NSIG` here?

sysdeps/linux/generic/sysdeps.cpp
```diff
@@ -625,14 +718,206 @@ int sys_reboot(int cmd) {
...
+	if (int e = sc_error(ret); e)
+		return e;
+	return 0;
+}
+
+int sys_inotify_create(int flags, int *fd) {
+	sc_result_t ret;
+
+	if(flags) {
+		ret = do_syscall(SYS_inotify_init1, flags);
```
64, line 808: Why not just call `inotify_init1` in both cases?

sysdeps/linux/generic/sysdeps.cpp
```diff
@@ -625,14 +718,206 @@ int sys_reboot(int cmd) {
...
+}
+
+int sys_inotify_rm_watch(int ifd, int wd) {
+	auto ret = do_syscall(SYS_inotify_rm_watch, ifd, wd);
+	if (int e = sc_error(ret); e)
+		return e;
+	return 0;
+}
+
+int sys_ttyname(int fd, char *buf, size_t size) {
```
64, line 870: This sysdep doesn't actually perform any raw syscalls, so I wonder if we could move it to `options/`? The only linux-specific behaviour here is assuming that `/proc/self/fd` is available, but we also rely on `procfs` in some other areas don't we?

64 (changes requested):

options/linux/generic/sys-prctl-stubs.cpp
no92: `musl` does also take varargs, while the man page talks about `unsigned long`s. I'm unsure what to follow here, but varargs feel like the less troublesome variant.

sysdeps/linux/generic/sysdeps.cpp
no92: The `procfs` way is presumably linux-specific and other sysdeps do implement this as a syscall, so moving this into `options/` doesn't look reasonable to me.

sysdeps/linux/generic/sysdeps.cpp
avdgrinten: Is that even correct? Why not `sizeof(sigset_t)`?

options/linux/generic/sys-prctl-stubs.cpp
```diff
@@ -1,12 +1,32 @@
...
+
+	unsigned long x[4];
+	va_list ap;
+	va_start(ap, op);
+	for(int i = 0; i < 4; i++)
+		x[i] = va_arg(ap, unsigned long);
+	va_end(ap);
+	int val;
+
+	if(int e = mlibc::sys_prctl(op, x[0], x[1], x[2], x[3], &val); e) {
```
avdgrinten, line 26: I dislike this interface to sys_prctl. What if Linux decides to add another argument? We could also pass the va_list here.

sysdeps/linux/generic/sysdeps.cpp
```diff
@@ -125,6 +125,45 @@ int sys_seek(int fd, off_t offset, int whence, off_t *new_offset) {
 	return 0;
 }
 
+int sys_chmod(const char *pathname, mode_t mode) {
+#ifdef SYS_chmod
+	auto ret = do_cp_syscall(SYS_chmod, pathname, mode);
+#else
+	auto ret = do_cp_syscall(SYS_fchmodat, AT_FDCWD, pathname, mode);
+#endif
```
avdgrinten, line 133: What's the advantage of this `#ifdef`? Can't we simply use SYS_fchmodat on all archs?

options/linux/include/sys/sysinfo.h
```diff
@@ -0,0 +1,28 @@
...
+	unsigned long freeram;
+	unsigned long sharedram;
+	unsigned long bufferram;
+	unsigned long totalswap;
+	unsigned long freeswap;
+	unsigned short procs;
+	unsigned long totalhigh;
+	unsigned long freehigh;
+	unsigned int mem_unit;
+	char _f[20 - 2 * sizeof(long) - sizeof(int)]; // Padding to 64 bytes according to my man page
```
avdgrinten, line 21: What do the Linux kernel headers say about this?

options/posix/include/mlibc/posix-sysdeps.hpp
```diff
@@ -155,15 +157,14 @@ int sys_vm_unmap(void *pointer, size_t size);
 [[gnu::weak]] int sys_accept(int fd, int *newfd, struct sockaddr *addr_ptr, socklen_t *addr_length);
 [[gnu::weak]] int sys_bind(int fd, const struct sockaddr *addr_ptr, socklen_t addr_length);
 [[gnu::weak]] int sys_connect(int fd, const struct sockaddr *addr_ptr, socklen_t addr_length);
-[[gnu::weak]] int sys_sockname(int fd, struct sockaddr *addr_ptr, socklen_t max_addr_length,
-		socklen_t *actual_length);
-[[gnu::weak]] int sys_peername(int fd, struct sockaddr *addr_ptr, socklen_t max_addr_length,
-	socklen_t *actual_length);
+[[gnu::weak]] int sys_sockname(int fd, struct sockaddr *addr_ptr, socklen_t *max_addr_length);
+[[gnu::weak]] int sys_peername(int fd, struct sockaddr *addr_ptr, socklen_t *max_addr_length);
```
avdgrinten, line 161: Why are these signatures changed?

avdgrinten: The previous approach looks much saner to me here.

options/linux/include/mlibc/linux-sysdeps.hpp
```diff
@@ -32,7 +32,7 @@ int sys_write(int fd, const void *buf, size_t count, ssize_t *bytes_written);
 [[gnu::weak]] int sys_timerfd_create(int flags, int *fd);
 [[gnu::weak]] int sys_timerfd_settime(int fd, int flags,
 		const struct itimerspec *value);
-[[gnu::weak]] int sys_signalfd_create(sigset_t, int flags, int *fd);
+[[gnu::weak]] int sys_signalfd_create(const sigset_t *, int flags, int *fd);
```
ArsenArsen, line 36: Why was this change made?

sysdeps/managarm/generic/file.cpp
```diff
@@ -992,6 +992,11 @@ int sys_tcsetattr(int fd, int when, const struct termios *attr) {
 	return 0;
 }
 
+int sys_tcdrain(int) {
+	mlibc::infoLogger() << "\e[35mmlibc: tcdrain() is a stub\e[39m" << frg::endlog;
```
ArsenArsen, line 996: Why not just implement this? It's one ioctl call.

sysdeps/linux/generic/sysdeps.cpp
```diff
@@ -541,6 +541,24 @@ int sys_setsockopt(int fd, int layer, int number, const void *buffer, socklen_t
 	return 0;
 }
 
+int sys_sockname(int fd, struct sockaddr *addr_ptr, socklen_t max_addr_length,
+		socklen_t *actual_length) {
+	auto ret = do_syscall(NR_getsockname, fd, addr_ptr, max_addr_length);
```
ArsenArsen: Are we using `SYS_` or `NR_`?

options/linux/generic/sys-timerfd.cpp
```diff
@@ -18,10 +18,8 @@ int timerfd_create(int clockid, int flags) {
 
 int timerfd_settime(int fd, int flags, const struct itimerspec *value,
 		struct itimerspec *oldvalue) {
-	__ensure(!oldvalue);
```
ArsenArsen, line 21:
```
       If  the  old_value argument is not NULL, then the itimerspec structure that it points to is used to return the setting of the timer that was current
       at the time of the call; see the description of timerfd_gettime() following.
```

Why ensure this to NULL?

sysdeps/linux/generic/sysdeps.cpp
```diff
@@ -747,6 +747,13 @@ int sys_timerfd_create(int clockid, int flags, int *fd) {
 	return 0;
 }
 
+int sys_timerfd_settime(int fd, int flags, const struct itimerspec *value, const struct itimerspec *oldvalue) {
+	auto ret = do_syscall(SYS_timerfd_settime, fd, flags, value, oldvalue);
```
ArsenArsen, line 802: Do we want to use the 32 bit variant here?

options/posix/generic/sys-stat-stubs.cpp
```diff
@@ -100,9 +100,18 @@ int mknodat(int dirfd, const char *path, mode_t mode, dev_t dev) {
...
+		errno = ENOSYS;
+		return -1;
+	}
+
+	if (int e = mlibc::sys_umask(mode); e) {
+		errno = e;
+		return -1;
+	}
+
 	return 0;
```
ArsenArsen, line 115: This is wrong. `umask(2)` is meant to return the old umask.

sysdeps/linux/generic/sysdeps.cpp
```diff
@@ -784,6 +784,20 @@ int sys_inotify_create(int flags, int *fd) {
...
+	return 0;
+}
+
+int sys_delete_module(const char *name, unsigned flags) {
+	auto ret = do_syscall(SYS_delete_module, name, flags);
+	if (int e = sc_error(ret); e)
+		return e;
+	return 0;
+}
+
```
ArsenArsen, line 829: We don't implement `init/delete_module` (and neither does glibc).

ArsenArsen (changes requested):

sysdeps/linux/generic/sysdeps.cpp
```diff
@@ -729,13 +1011,38 @@ int sys_mkdirat(int dirfd, const char *path, mode_t mode) {
...
 }
 
+int sys_symlinkat(const char *target_path, int dirfd, const char *link_path) {
+	auto ret = do_syscall(SYS_symlinkat, target_path, dirfd, link_path);
+	if (int e = sc_error(ret); e)
+		return e;
+	return 0;
+}
+
+mode_t sys_umask(mode_t mode) {
```
64, line 1039: This should return the old umask.

sysdeps/linux/generic/sysdeps.cpp
64: That is not an entirely satisfying answer, but okay.

64 (changes requested):

sysdeps/linux/generic/sysdeps.cpp
no92: What exactly do you mean by '32 bit variant'?

options/linux/include/sys/inotify.h
```diff
@@ -2,13 +2,14 @@
...
+#include <abi-bits/fcntl.h>
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
-#define IN_CLOEXEC 1
-#define IN_NONBLOCK 2
+#define IN_CLOEXEC O_CLOEXEC
+#define IN_NONBLOCK O_NONBLOCK
```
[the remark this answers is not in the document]
no92: Yes, this updates the constants to match what the Linux kernel expects.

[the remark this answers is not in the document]
avdgrinten: That means that this commits breaks the ABI though.

sysdeps/linux/generic/sysdeps.cpp
no92: Using the dedicated syscall, if available, feels like the correct choice to me. musl also prefers the dedicated syscall, if possible.

options/linux/generic/sys-prctl-stubs.cpp
64: We could pass a view type, for something more C++ friendly.

sysdeps/linux/generic/sysdeps.cpp
64: We already use the more general syscalls where possible, so we should be consistent.  IMO less ifdefs make the code easier to follow and easier to test.

options/linux/include/mlibc/linux-sysdeps.hpp
no92: This now matches the definition given by the `signalfd(2)` man page.

sysdeps/linux/generic/sysdeps.cpp
ArsenArsen: Ah, wait, never mind. There is a `SYS_timerfd_settime64` syscall but it only exists on 32-bit architectures.

options/posix/include/mlibc/posix-sysdeps.hpp
avdgrinten: Well, IMHO the in-out argument (which uses the in-argument for a completely different purpose than the out-argument) is quite ugly. Just that Linux also does it that way doesn't mean that we should change the sysdeps to force every port to do it. IMHO Linux is "wrong" here.

sysdeps/linux/generic/sysdeps.cpp
ArsenArsen: That's not in this PR though.

options/linux/generic/sys-timerfd.cpp
no92: The ensure is getting removed in this PR.

sysdeps/linux/generic/sysdeps.cpp
no92: This is an artifact from before the syscall PR. It gets changed in a later commit touching `sys_sockname`. In order to avoid excessively force-pushing to this PR, I'd leave this commit as is.

sysdeps/managarm/generic/file.cpp
ArsenArsen: Ah, this is the managarm sysdep. Apologies.

options/linux/include/module.h
```diff
@@ -0,0 +1,21 @@
+#ifndef _MODULE_H
+#define _MODULE_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/*
+ * Musl adds them, even tho they aren't specified, but doesn't export them.
```
64, line 9: 'them' -> these, 'tho' -> though

sysdeps/linux/generic/sysdeps.cpp
no92: As far as the (Linux) kernel is concerned, it can handle `sizemask` not being equal to `sizeof sigset_t` (see the `SYSCALL_DEFINE4` for `signalfd4` in signalfd.c). Passing the supported count of signals seems reasonable here. Cross-checking with musl, they do it using NSIG, too.

no92: A fixup has already been pushed changing this to utilize NSIG.

options/linux/generic/sys-prctl-stubs.cpp
no92: Is Linux changing syscall args really a concern? I'd guess they would provide a new, numbered syscall instead.

64: It probably is a concern, since musl has made it a variadic function.

sysdeps/linux/generic/sysdeps.cpp
64: Can't you just push another fixup?

no92: The problem is that a later commit rectifies the issue, which would necessitate a rebase.

no92: I can, however, note this and fix it during the rebase that I'll be doing for squashing the fixups.

options/linux/include/sys/sysinfo.h
no92: `include/uapi/linux/sysinfo.h` does to what amounts to the same.

The man page specifically says:

> Since Linux 2.3.23 (i386) and Linux 2.3.48 (all architectures) the structure is:
> [...]
> `char _f[20-2*sizeof(long)-sizeof(int)]; /* Padding to 64 bytes */`

ElectrodeYT (approved):

abis/mlibc/inotify.h
```diff
@@ -0,0 +1,7 @@
+#ifndef _ABIBITS_INOTIFY_H
+#define _ABIBITS_INOTIFY_H
+
+#define __MLIBC_IN_CLOEXEC 1
+#define __MLIBC_IN_NONBLOCK 2
```
avdgrinten, line 5: Let's not add any more `__MLIBC` ABI constants, the `__MLIBC_O_PATH` etc. ones in `abi-bits/abi.h` were a mistake (by me :smile:).

avdgrinten (changes requested):

options/posix/include/mlibc/posix-sysdeps.hpp
```diff
@@ -147,6 +148,7 @@ int sys_vm_unmap(void *pointer, size_t size);
 [[gnu::weak]] int sys_setsockopt(int fd, int layer, int number,
 		const void *buffer, socklen_t size);
 [[gnu::weak]] int sys_waitpid(pid_t pid, int *status, int flags, pid_t *ret_pid);
+[[gnu::weak]] int sys_wait4(pid_t pid, int *status, int flags, struct rusage *ru, pid_t *ret_pid);
```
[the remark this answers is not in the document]
no92: So should we remove `sys_waitpid` then?

options/linux/include/module.h
```diff
@@ -0,0 +1,21 @@
+#ifndef _MODULE_H
+#define _MODULE_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/*
+ * Musl adds those, even though they aren't specified, but doesn't export them.
```
64, line 9: 'those' should still be 'these'

options/posix/include/mlibc/posix-sysdeps.hpp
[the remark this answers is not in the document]
avdgrinten: Or just add the rusage pointer to sys_waitpid (and let existing impls return ENOSYS if it's non-null).

PR #583, opened 2022-06-30
rtdl: implement RTLD_NEXT

### LINKED ISSUE #577: rtdl: RTLD_NEXT is unimplemented
While trying to use `libasan` on mlibc LFS, we ran into [this](https://github.com/managarm/mlibc/blob/a53280ab5fa0a851b62abcbf77bf0c41707199a1/options/rtdl/generic/main.cpp#L392) ensure, preventing asan from working. Fixing it would allow us to use `libasan`.

64:
Fixes #577. Split off from #448. Blocked on the new RTDL testing framework in #586 so I can add a meaningful test for this.

### DIFF
--- a/options/posix/generic/dlfcn-stubs.cpp
+++ b/options/posix/generic/dlfcn-stubs.cpp
@@ -12,37 +12,29 @@ struct __dlapi_symbol {
 };
 
 extern "C" const char *__dlapi_error();
-extern "C" void *__dlapi_open(const char *, int);
-extern "C" void *__dlapi_resolve(void *, const char *);
+extern "C" void *__dlapi_open(const char *, int, void *);
+extern "C" void *__dlapi_resolve(void *, const char *, void *);
 extern "C" int __dlapi_reverse(const void *, __dlapi_symbol *);
+extern "C" int __dlapi_close(void *);
 
-int dlclose(void *) {
-	mlibc::infoLogger() << "\e[31mmlibc: dlclose() is a no-op\e[39m" << frg::endlog;
-	return 0;
+int dlclose(void *handle) {
+	return __dlapi_close(handle);
 }
 
 char *dlerror(void) {
 	return const_cast<char *>(__dlapi_error());
 }
 
+[[gnu::noinline]]
 void *dlopen(const char *file, int flags) {
-	// The following three flags are glibc extensions.
-	// TODO: Validate the flags.
-	if(flags & RTLD_NOLOAD) {
-		mlibc::infoLogger() << "\e[31mmlibc: dlopen(RTLD_NOLOAD) always fails\e[39m"
-				<< frg::endlog;
-		return nullptr;
-	}
-
-	int unhandled = flags & (RTLD_NODELETE | RTLD_DEEPBIND);
-	if (unhandled)
-		mlibc::infoLogger() << "mlibc: unhandled dlopen flags " << unhandled << frg::endlog;
-
-	return __dlapi_open(file, !(flags & RTLD_GLOBAL));
+	auto ra = __builtin_extract_return_addr(__builtin_return_address(0));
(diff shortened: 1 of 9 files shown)


options/posix/generic/dlfcn-stubs.cpp
```diff
@@ -12,37 +12,27 @@ struct __dlapi_symbol {
...
-		mlibc::infoLogger() << "mlibc: unhandled dlopen flags " << unhandled << frg::endlog;
-
-	return __dlapi_open(file, !(flags & RTLD_GLOBAL));
+	auto ra = __builtin_extract_return_addr(__builtin_return_address(0));
+	return __dlapi_open(file, flags, ra);
 }
 
 void *dlsym(void *__restrict handle, const char *__restrict string) {
-	return __dlapi_resolve(handle, string);
+	auto ra = __builtin_extract_return_addr(__builtin_return_address(0));
```
avdgrinten, line 34: The functions containing this should be marked `[[gnu::noinline]]`.

options/rtdl/generic/linker.cpp
```diff
@@ -262,6 +262,25 @@ SharedObject *ObjectRepository::requestObjectAtPath(frg::string_view path, uint6
...
+	uintptr_t target = reinterpret_cast<uintptr_t>(addr);
+
+	for (auto [name, object] : _nameMap) {
+		// Search all PT_LOAD segments for the specified address.
+		for(size_t j = 0; j < object->phdrCount; j++) {
+			auto phdr = (Elf64_Phdr *)((uintptr_t)object->phdrPointer + j * object->phdrEntrySize);
+			if (phdr->p_type == PT_LOAD) {
+				uintptr_t start = object->baseAddress + phdr->p_vaddr;
+				uintptr_t end = start + phdr->p_memsz;
+				if (start <= target && target <= end)
```
avdgrinten, line 275: `target < end`, right? For example, if start = 0x1000, size = 0x1000, then 0x2000 is already on the next page.

options/rtdl/generic/main.cpp
```diff
@@ -385,16 +402,22 @@ void *__dlapi_open(const char *file, int local) {
...
 	if (handle == RTLD_DEFAULT) {
 		target = Scope::resolveWholeScope(globalScope.get(), string, 0);
+	} else if (handle == RTLD_NEXT) {
+		SharedObject *origin = initialRepository->findCaller(returnAddress);
+		if (!origin) {
+			mlibc::panicLogger() << "rtdl: unable to determine calling object of dlsym "
+				<< "(ra = " << returnAddress << ")" << frg::endlog;
+		}
+
+		target = Scope::resolveNext(globalScope.get(), string, origin);
```
avdgrinten, line 420: The local scope is not handled here (yet)?

64: Yeah, I guess we just use the object's scope when that is implemented

tests/rtdl/rtld_next/meson.build
```diff
@@ -0,0 +1,25 @@
+# Prevent tail calls, as it breaks libc's 'calling object' detection.
+no_tail_calls = '-fno-optimize-sibling-calls'
```
avdgrinten, line 2: Is this needed despite `[[gnu::noinline]]`?

avdgrinten (commented):
LGTM except for the comment below.

tests/rtdl/rtld_next/meson.build
64: Yes:
```c
void *foo() {
    return dlsym(...);
}
```
Even though dlsym isn't inlined, the compiler will generate a tail call here. But that'd cause the return address visible to dlsym to be foo's caller instead, which is not what we want if foo is called from a different object.

PR #586, opened 2022-06-30
tests/rtdl: move to subdir and augment with test DSOs

64:
This PR augments our RTDL testing code by allowing us to build a few test DSOs. We also augment the `dl_iterate_phdr` test accordingly. The idea is that future RTDL changes can add a new subdirectory with a similar `meson.build`.


tests/rtdl/meson.build
```diff
@@ -0,0 +1,39 @@
...
+	test(test_name, exec, suite: 'rtdl', depends: test_depends)
+
+	if build_tests_host_libc and not host_libc_excluded_test_cases.contains(test_name)
+		exec = executable('host-libc-' + test_name, test_name / 'test.c',
+			link_with: test_native_link_with,
+			dependencies: rtlib_deps,
+			build_rpath: test_rpath,
+			override_options: 'b_sanitize=address,undefined',
+			c_args: ['-D_GNU_SOURCE', '-DUSE_HOST_LIBC', '-pthread'],
+			link_args: ['-lresolv', '-ldl', '-pthread', '-Wl,--disable-new-dtags'],
```
qookei, line 34: rtdl tests don't need libresolv or pthread right? Also, what's up with `--disable-new-dtags` (and why isn't it needed when building the mlibc tests)?

qookei (commented):
Seems fine to me, apart from the comment (which isn't a huge blocker either).

tests/rtdl/meson.build
64: `--disable-new-dtags` forces the use of DT_RPATH vs the debian default of DT_RUNPATH. For some reason, dlopen doesn't seem to search DT_RUNPATH on the CI, so this flag makes it work (not really sure why).

qookei: Alright, seems fine then.

qookei: What about the extra deps?

64: Yeah, I'll remove those, they were just copied from the other host tests.

qookei (approved):
Looks good to me

no92 (approved):

64:
Okay, this now works without the need for `--disable-new-dtags`.

avdgrinten:
Very nice! Thanks for fixing this.

PR #589, opened 2022-06-30
sysdeps: Add Vinix

mintsuki:


sysdeps/vinix/generic/entry.cpp
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	mlibc::sys_sigreturn(ret_context, prev_mask);
+
+	__builtin_unreachable();
+}
+
+extern "C" void __mlibc_entry(int (*main_fn)(int argc, char *argv[], char *env[])) {
+	mlibc::sys_sigentry((void *)__mlibc_sigentry);
+
```
64, line 99: Can you add a similar comment here https://github.com/managarm/mlibc/blob/719cfcbc8a3ac3c33a1129b669dea2449c2dbf85/sysdeps/lemon/generic/entry.cpp#L30, just so that other people making their own ports based off this don't forget? Alternatively you can call the function and fix it properly if you prefer.

64:
LGTM apart from the comment provided CI goes green.

options/internal/include/mlibc/internal-sysdeps.hpp
```diff
@@ -35,6 +36,9 @@ int sys_close(int fd);
 int sys_vm_map(void *hint, size_t size, int prot, int flags, int fd, off_t offset, void **window);
 int sys_vm_unmap(void *pointer, size_t size);
 
+int sys_sigentry(void *pointer);
+[[noreturn]] int sys_sigreturn(void *context, sigset_t old_mask);
```
avdgrinten, line 40: Since this is used only on Vinix, it would be better to keep it within the sysdeps.

sysdeps/vinix/generic/entry.cpp
mintsuki: Done

options/internal/include/mlibc/internal-sysdeps.hpp
mintsuki: Done

Dennisbonke (approved):
LGTM

Geertiebear (approved):
LGTM.

PR #591, opened 2022-06-30
options: allow recursive file locking

### LINKED ISSUE #408: ansi: ungetc hangs if lock is already acquired
If a program locks a file or stream with `flockfile()` and then calls `ungetc()` on said file or stream, the program will hang on acquiring the lock in ungetc. This behaviour was observed in gtk3 running on Managarm. A hotfix of `funlockfile()` -> `ungetc()` -> `flockfile()` fixed that issue, pointing again at an issue with our locking implementation used by `ungetc()`.

64:
Fixes #408. We also implement `ftrylockfile`.

### DIFF
--- a/options/ansi/generic/file-io.cpp
+++ b/options/ansi/generic/file-io.cpp
@@ -584,7 +584,7 @@ int fileno_unlocked(FILE *file_base) {
 
 int fileno(FILE *file_base) {
 	auto file = static_cast<mlibc::fd_file *>(file_base);
-	frg::unique_lock<FutexLock> lock(file->_lock);
+	frg::unique_lock lock(file->_lock);
 	return fileno_unlocked(file_base);
 }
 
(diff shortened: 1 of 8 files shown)


64:
Should not be an ABI break, despite internal changes.

options/ansi/generic/file-io.cpp
```diff
@@ -584,7 +584,7 @@ int fileno_unlocked(FILE *file_base) {
 
 int fileno(FILE *file_base) {
 	auto file = static_cast<mlibc::fd_file *>(file_base);
-	frg::unique_lock<FutexLock> lock(file->_lock);
+	frg::unique_lock<RecursiveFutexLock> lock(file->_lock);
```
avdgrinten, line 587: It should actually be possible to drop the template args completely due to CTAD.

PR #463, opened 2022-06-06
Add si_fd to siginfo_t

Dennisbonke:
The abi break part of #458, adds one member to siginfo_t.

Partly fixes #257 
Part of the Chromium on Managarm project.

PR #594, opened 2022-06-30
abis/mlibc: make struct sigaction and siginfo_t match linux ABI

### LINKED ISSUE #257: posix: Review ABI of siginfo_t
siginfo_t has lots of additional members on Linux, review whether we need them.

### LINKED ISSUE #590: abis/mlibc/signal: sa_handler and sa_sigaction should be a union
This matches the Linux ABI and should also ease the implementation of new ports, see e.g #589.

64:
Fixes #257. Fixes #590. Obsoletes #463.

### DIFF
--- a/abis/mlibc/signal.h
+++ b/abis/mlibc/signal.h
@@ -1,6 +1,7 @@
 #ifndef _ABIBITS_SIGNAL_H
 #define _ABIBITS_SIGNAL_H
 
+#include <abi-bits/clock_t.h>
 #include <abi-bits/pid_t.h>
 #include <abi-bits/uid_t.h>
 #include <bits/size_t.h>
--- a/abis/mlibc/signal.h
+++ b/abis/mlibc/signal.h
@@ -11,16 +11,73 @@ union sigval {
 	void *sival_ptr;
 };
 
+// struct taken from musl.
+
 typedef struct {
-	int si_signo;
-	int si_code;
-	int si_errno;
-	pid_t si_pid;
-	uid_t si_uid;
-	void *si_addr;
-	int si_status;
-	union sigval si_value;
+	int si_signo, si_errno, si_code;
+	union {
+		char __pad[128 - 2 * sizeof(int) - sizeof(long)];
+		struct {
+			union {
+				struct {
+					pid_t si_pid;
+					uid_t si_uid;
+				} __piduid;
+				struct {
+					int si_timerid;
+					int si_overrun;
+				} __timer;
+			} __first;
+			union {
+				union sigval si_value;
+				struct {
+					int si_status;
+					clock_t si_utime, si_stime;
+				} __sigchld;
+			} __second;
+		} __si_common;
+		struct {
+			void *si_addr;
+			short si_addr_lsb;
+			union {
+				struct {
+					void *si_lower;
+					void *si_upper;
+				} __addr_bnd;
+				unsigned si_pkey;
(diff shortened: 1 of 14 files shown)


abis/mlibc/signal.h
```diff
@@ -148,11 +206,16 @@ struct sigevent {
 };
 
 struct sigaction {
```
Geertiebear, line 208: So afaict this struct does not actually match the linux kernel's ABI. In the ``sys_sigaction()`` of the linux sysdeps we have this ``struct ksigaction``, which corrects for this mismatch, it would be nice to fix this.

Geertiebear (changes requested):
LGTM except for one comment.

abis/mlibc/signal.h
```diff
@@ -10,16 +11,73 @@ union sigval {
...
+#define si_pkey    __si_fields.__sigfault.__first.si_pkey
+#define si_band    __si_fields.__sigpoll.si_band
+#define si_fd      __si_fields.__sigpoll.si_fd
+#define si_timerid __si_fields.__si_common.__first.__timer.si_timerid
+#define si_overrun __si_fields.__si_common.__first.__timer.si_overrun
+#define si_ptr     si_value.sival_ptr
+#define si_int     si_value.sival_int
+#define si_call_addr __si_fields.__sigsys.si_call_addr
+#define si_syscall __si_fields.__sigsys.si_syscall
+#define si_arch    __si_fields.__sigsys.si_arch
```
avdgrinten, line 80: These defines are incredibly ugly. Can't we use anonymous unions and structs to avoid this mess?

64: Anonymous unions and structs are a C11 feature, right?

abis/mlibc/signal.h
64: Hm, is there a reason why we don't change this directly in the Linux ABI?

abis/mlibc/signal.h
ArsenArsen: Yes

abis/mlibc/signal.h
Geertiebear: Not afaik, no.

64: I don't think we can fix that because `sa_flags` is an `int` in the userspace API but `unsigned long` in the kernel. So we probably need to keep that as-is.

Geertiebear: Ah okay that's annoying.

avdgrinten:
Overall, I'm not convinced that this is a good change. All OSes that currently use this ABI will need to be changed (which OSes are affected?). What do they gain by doing that? Also, what do they gain over simply using `abis/linux` instead?

64:
Well currently our `siginfo_t` is missing a lot of members that userspace expects. I don't think there's a way to avoid that without kernel changes.

64:
I wouldn't be opposed to deleting this in favour of abis/linux either, tbh.

avdgrinten:
Well, it's one issue to change Managarm, but this also affects Ironclad. If only Managarm wants to change, we can easily switch Managarm to `abis/linux` w/o forcing others to update.

64:
That's true: if they want to change to use `abis/linux` (and why wouldn't they, given an ABI break will happen anyway?) then we ought to ping all the sysdeps maintainers to OK this.

PR #596, opened 2022-06-30
abis/mlibc: delete `wait.h` in favour of linux ABI

### LINKED ISSUE #466: Review the ABI of WEXITSTATUS and friends
The current mlibc ABI for these macros looks like this:

```c
#define WEXITSTATUS(x) ((x) & 0x000000FF)
#define WIFCONTINUED(x) ((x) & 0x00000100)
#define WIFEXITED(x) ((x) & 0x00000200)
#define WIFSIGNALED(x) ((x) & 0x00000400)
#define WIFSTOPPED(x) ((x) & 0x00000800)
#define WSTOPSIG(x) (((x) & 0x00FF0000) >> 16)
#define WTERMSIG(x) (((x) & 0xFF000000) >> 24)
#define WCOREDUMP(x) ((x) & WCOREFLAG)
```

There are a few issues:

1. `WEXITSTATUS(0)` is false, which seems to break the assumptions of some programs e.g fish. While the programs themselves should be fixed so as not to assume this, it's probably better if we change this in mlibc because much of the breakage won't be immediately obvious to us.
2. We use separate bits for 'exited normally' and 'killed by signal'. Instead, we should follow Linux and use some bits to store the signal,and use signal == 0 to signify that the program exited normally. This would also fix 1.
3. There's no API for a kernel to actually *set* any of these bits, which leads us to do hacks like https://github.com/managarm/managarm/blob/9374f1e129db96eaa16bbabeb1f23ea9772938ae/posix/subsystem/src/main.cpp#L992-L996. We should probably offer a macro like glibc's `W_EXITCODE(ret, sig)` which just expands to `((ret << 8) | sig)`.

64:
Fixes #466.


abis/linux/wait.h
```diff
@@ -22,4 +22,7 @@
 #define WIFCONTINUED(x) ((x) == 0xffff)
 #define WCOREDUMP(x) ((x) & WCOREFLAG)
 
+// glibc extension, but also useful for the kernel
```
ArsenArsen, line 25: Use C-style comments in public facing headers (yes, I'm aware of the line comment right below it, I'll be replacing them all at some point in the future).

ArsenArsen (changes requested):

abis/linux/wait.h
64: Why don't we merge this and change them later, then?

ArsenArsen: Adding more serves no good other than to increase the size of further patches (hence increasing the grounds of conflict)

64:
Changed as requested.

PR #598, opened 2022-06-30
options/ansi: cache value of env var MLIBC_DEBUG_MALLOC

### LINKED ISSUE #597: malloc: value of env var MLIBC_DEBUG_MALLOC should be cached
At the moment, every malloc / free / realloc call causes a call to `getenv`.

64:
Fixes #597.

### DIFF
--- a/options/ansi/generic/stdlib-stubs.cpp
+++ b/options/ansi/generic/stdlib-stubs.cpp
@@ -421,10 +421,19 @@ size_t wcstombs(char *mb_string, const wchar_t *wc_string, size_t max_size) {
 	return wcsrtombs(mb_string, &wc_string, max_size, 0);
 }
 
+static bool debugMalloc() {
+	static bool debug = [] () -> bool {
+		// TODO: We should use a locking variant of getenv here to guard
+		// against concurrent modifications of the environment.
+		auto value = getenv("MLIBC_DEBUG_MALLOC");
+		return value && !strcmp(value, "1");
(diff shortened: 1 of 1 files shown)


options/ansi/generic/stdlib-stubs.cpp
```diff
@@ -421,10 +421,28 @@ size_t wcstombs(char *mb_string, const wchar_t *wc_string, size_t max_size) {
...
+				__ATOMIC_ACQUIRE, __ATOMIC_ACQUIRE)) {
+		// TODO: We should use a locking variant of getenv here to guard
+		// against concurrent modifications of the environment.
+		auto value = getenv("MLIBC_DEBUG_MALLOC") != nullptr;
+		__atomic_store_n(&debug, value, __ATOMIC_RELAXED);
+		return value;
+	}
+
+	return __atomic_load_n(&debug, __ATOMIC_RELAXED);
+}
```
avdgrinten, line 441: I think this can be vastly simplified to:
```suggestion
static bool debugMalloc() {
	static bool debug = [] () -> bool {
		return getenv("MLIBC_DEBUG_MALLOC");
	}(); // Immediately invoked.
	return debug;
}
```
while also avoiding the race. Note that it would be a bit nicer to also check the value of `MLIBC_DEBUG_MALLOC` (right now, `MLIBC_DEBUG_MALLOC=0` turns on debugging, which is a bit counterintuitive).

avdgrinten (approved):

Issue #530, opened 2022-06-22
options/posix: `pthread_barrierattr_t`, `pthread_mutexattr_t`, `pthread_rwlockattr_t` and `pthread_condattr_t` are incomplete


64:

Dennisbonke:
Wasn't this fixed already?

PR #600, opened 2022-07-01
options/posix: implement pthread_barrier

64:
Progress on #530.

### DIFF
--- a/options/posix/generic/pthread-stubs.cpp
+++ b/options/posix/generic/pthread-stubs.cpp
@@ -1216,28 +1216,112 @@ int pthread_cond_broadcast(pthread_cond_t *cond) {
 // pthread_barrierattr and pthread_barrier functions.
 // ----------------------------------------------------------------------------
 
-int pthread_barrierattr_init(pthread_barrierattr_t *) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+int pthread_barrierattr_init(pthread_barrierattr_t *attr) {
+	attr->__mlibc_pshared = PTHREAD_PROCESS_PRIVATE;
+	return 0;
 }
+
+int pthread_barrierattr_getpshared(const pthread_barrierattr_t *__restrict attr,
+		int *__restrict pshared) {
+	*pshared = attr->__mlibc_pshared;
+	return 0;
+}
+
+int pthread_barrierattr_setpshared(pthread_barrierattr_t *attr, int pshared) {
+	if (pshared != PTHREAD_PROCESS_SHARED && pshared != PTHREAD_PROCESS_PRIVATE)
+		return EINVAL;
+
+	attr->__mlibc_pshared = pshared;
+	return 0;
+}
+
 int pthread_barrierattr_destroy(pthread_barrierattr_t *) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+	return 0;
 }
 
-int pthread_barrier_init(pthread_barrier_t *__restrict, const pthread_barrierattr_t *__restrict,
-		unsigned int) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
-}
-int pthread_barrier_destroy(pthread_barrier_t *) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+int pthread_barrier_init(pthread_barrier_t *__restrict barrier,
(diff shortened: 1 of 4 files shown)


avdgrinten:
I don't think that this implementation is correct. We can't just assume that the barrier is reset after we're woken up from `sys_futex_wait()` (there could be spurious wakeups, for example if a signal is triggered -- although that would return `EINTR`). Also, if we simply check whether `__mlibc_waiting` is reset to zero, we might miss a reset if `__mlibc_waiting` is incremented after the reset but before we can check it.

A strategy that should work is the following: store a sequence number inside the `pthread_barrier_t`. Each thread reads the sequence number before incrementing `__mlibc_waiting`. The last thread that increments `__mlibc_waiting` via CAS (i.e., the thread where `expected + 1== __mlibc_waiting`) increments the sequence number via `__atomic_fetch_add`. We perform the `sys_futex_wait()` on the sequence number, not on `__mlibc_waiting`, until the sequence number increments at least once.

64:
> A strategy that should work is the following: store a sequence number inside the `pthread_barrier_t`. Each thread reads the sequence number before incrementing `__mlibc_waiting`. The last thread that increments `__mlibc_waiting` via CAS (i.e., the thread where `expected + 1== __mlibc_waiting`) increments the sequence number via `__atomic_fetch_add`. We perform the `sys_futex_wait()` on the sequence number, not on `__mlibc_waiting`, until the sequence number increments at least once.

I think that works, but note the following requirement in POSIX:

> [after all threads have returned from `pthread_barrier_wait`] the barrier shall be reset to the state it had as a result of the most recent [pthread_barrier_init()](https://pubs.opengroup.org/onlinepubs/009696899/functions/pthread_barrier_init.html) function that referenced it.

A strict interpretation of this would preclude using a sequence number that gets incremented after every wait has finished. But I don't think any program would be able to meaningfully observe that, so it's probably not worth worrying about.

64:
Okay, I've changed the implementation to the suggested, but you'll have to double check the memory orderings for me.

64:
Hmm, actually I think that implementation doesn't quite work either.

Consider the scenario where the last thread to hit the barrier increments seq and then immediately calls `pthread_barrier_destroy()` followed by `pthread_barrier_init()`. The seq will be reset to zero and I don't think there's a way to avoid that. We probably need to keep track of the number of threads which have entered the barrier wait and the number of threads that have left the barrier wait, and only destroy once the latter count reaches zero.

64:
Now that I think about it, our other locks probably have the exact same bug. Any logic like
```cpp
atomically_unset_lock_bit(&lock->futex_word);
// access lock->foo
mlibc::sys_futex_wake(&lock->futex_word);
```
is vulnerable to this race condition, since between the first two lines, another thread could acquire the lock, unlock it and then immediately destroy + unmap the backing memory and cause the `lock->foo` access to segfault. `sys_futex_wake` doesn't seem vulnerable to this itself, since it just returns EACCESS if the access would trigger a segfault.

avdgrinten:
I don't think that the `pthread_*_destroy()` functions are allowed to race with any other function on the same object.

64:
Well I'm not talking about a race between destroy and a new wait call, I'm talking about the scenario where a barrier is just 'opened' and one thread immediately calls destroy while the others haven't returned from pthread_barrier_wait().

It seems that glibc also had this bug on mutexes and barriers, FWIW

64:
the glibc bug in pthread_mutex_unlock can be found here: https://sourceware.org/bugzilla/show_bug.cgi?id=13690

avdgrinten:
Ah, I see. Yes, seems that musl indeed keeps track of the number of waiters to work around this. I guess that's no *too* bad since it only adds atomic ops on the slow path.

64:
Yeah, I've tried to implement something that works now. But do you agree that the problem is more general? e.g our pthread mutex unlock also suffers from the same issue as glibc did:
https://github.com/managarm/mlibc/blob/022dde31519ab88a8c1ac63d637c82bf1f43805d/options/posix/generic/pthread-stubs.cpp#L866

avdgrinten:
Yes, agreed.

Geertiebear (changes requested):

Geertiebear (approved):
LGTM now. Since the atomics on this are pretty important, maybe we should let @avdgrinten also take a look still.

PR #602, opened 2022-07-01
Small assortment of various LFS changes

### LINKED ISSUE #537: options/posix: mkostemp should ignore O_WRONLY
Since we pass `O_RDWR`, these will OR together to give `O_ACCMODE` which is not what we want. Technically (according to the man page) callers shouldn't pass this flag, but musl at least seems to work around this in the way I describe.

We should also probably move `mkostemp` to `options/glibc` instead.

no92:
Fixes #537.


options/posix/include/bits/posix/timer_t.h
```diff
@@ -0,0 +1,6 @@
+#ifndef _MLIBC_TIMER_T_H
+#define _MLIBC_TIMER_T_H
+
+typedef void * timer_t;
```
avdgrinten, line 4: Why is this a `void *`? What will it point to if POSIX timers are implemented?

no92: On both glibc and musl, `timer_t` is `typedef`ed as a `void *`.

avdgrinten: What does it point to?

no92: On musl, it points to a `intptr_t`. I remember reading that on glibc, it also points to some integer type.

options/linux/include/syscall.h
```diff
@@ -0,0 +1 @@
+#include <sys/syscall.h>
```
64, line 1: Why is this needed? This should probably live in the sysdeps folder since it's specific to the Linux kernel.

64:
The ABI diff fails because sysinfo was (correctly) changed to have C linkage.

options/posix/generic/posix_stdlib.cpp
```diff
@@ -487,3 +488,8 @@ char *canonicalize_file_name(const char *) {
 	__ensure(!"Not implemented");
 	__builtin_unreachable();
 }
+
+int clearenv(void) {
```
64, line 492: This should live with the other environment functions.

options/posix/generic/sys-mman-stubs.cpp
```diff
@@ -23,9 +23,17 @@ int mlock(const void *, size_t) {
 	__builtin_unreachable();
 }
 
-int mlockall(int) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+int mlockall(int flags) {
+	if(!mlibc::sys_mlockall) {
```
64, line 27: This should use the checking macro

options/posix/generic/unistd-stubs.cpp
```diff
@@ -487,7 +487,16 @@ long pathconf(const char *, int name) {
 }
 
 int pause(void) {
-	__ensure(!"Not implemented");
+	if(!mlibc::sys_pause) {
+		MLIBC_MISSING_SYSDEP();
```
64, line 491: Same here.

options/linux/include/syscall.h
no92: This header is present on both glibc and musl, and Dennis ran into some program expecting this to exist. Both of the other libcs just include `<sys/syscall.h>` as well.

64: Ok, but it should still probably go in the Linux sysdeps. I can conceivably imagine programs checking has_include for this header and making decisions based on that, so it's better for it to not be present unless it's implemented correctly.

options/posix/generic/posix_stdlib.cpp
no92: `clearenv` should probably only be included in the posix option, and even there it's kinda doubtful - it hasn't been accepted into the standard as far as I can tell. So I'm only moving this up to be right below `secure_getenv`.

64: The other functions in environment.cpp are behind a posix ifdef already, so IMO it'd make more sense to go there.

options/linux/include/syscall.h
ArsenArsen: Isn't sys/syscall.h also installed as a part of option/linux?

64: There's one there, but I don't think it's used and we ought to remove it.

ArsenArsen: hm? what do you mean?

64: Check options/linux/include/sys/syscall.h and sysdeps/linux/include/sys/syscall.h

64 (approved):

options/posix/generic/posix_stdlib.cpp
```diff
@@ -474,6 +475,11 @@ char *secure_getenv(const char *name) {
 		return getenv(name);
 }
 
+int clearenv(void) {
+	__ensure(!"Not implemented");
```
ArsenArsen, line 479: Why not?

options/posix/generic/sys-time-stubs.cpp
```diff
@@ -68,3 +68,8 @@ int setitimer(int which, const struct itimerval *new_value, struct itimerval *ol
 	}
 	return 0;
 }
+
+int timer_create(clockid_t, struct sigevent *__restrict, timer_t *__restrict) {
+	__ensure(!"Not implemented");
+	__builtin_unreachable();
```
ArsenArsen, line 74: Didn't we have an implementation of this in some other PR (and can we add one if we didn't?)

options/posix/generic/unistd-stubs.cpp
```diff
@@ -487,7 +487,12 @@ long pathconf(const char *, int name) {
 }
 
 int pause(void) {
-	__ensure(!"Not implemented");
+	MLIBC_CHECK_OR_ENOSYS(mlibc::sys_pause, -1);
+	if(int e = mlibc::sys_pause(); e) {
```
ArsenArsen, line 491: Is it possible for `sc_error(e) != e` in case of error? If so, this return value is wrong

sysdeps/linux/generic/sysdeps.cpp
```diff
@@ -893,6 +893,24 @@ int sys_ttyname(int fd, char *buf, size_t size) {
 	return 0;
 }
 
+int sys_pause() {
+#ifdef SYS_pause
+	auto ret = do_syscall(SYS_pause);
+#else
+	auto ret = do_syscall(SYS_ppoll, 0, 0, 0, 0);
+#endif
```
ArsenArsen, line 901: Why?

ArsenArsen (changes requested):

sysdeps/linux/generic/sysdeps.cpp
64: Presumably RISC-V does not have a pause syscall.

no92: Indeed, RISC-V does not have pause.

options/posix/generic/sys-time-stubs.cpp
no92: Not that I would know of or could find. Given that the implementation is non-trivial and I wouldn't know anything that uses it, I'd leave it for a future PR.

ArsenArsen:
BTW, can you add tests for the new functions?

64 (approved):

Dennisbonke (approved):
LGTM

Geertiebear (approved):
LGTM.

PR #603, opened 2022-07-01
options/posix: implement pthread_mutexattr

64:
Progress on #530.

### DIFF
--- a/options/posix/generic/pthread-stubs.cpp
+++ b/options/posix/generic/pthread-stubs.cpp
@@ -862,58 +862,85 @@ int pthread_once(pthread_once_t *once, void (*function) (void)) {
 // pthread_mutexattr functions
 int pthread_mutexattr_init(pthread_mutexattr_t *attr) {
 	SCOPE_TRACE();
-
-	memset(attr, 0, sizeof(pthread_mutexattr_t));
+	attr->__mlibc_type = PTHREAD_MUTEX_DEFAULT;
+	attr->__mlibc_robust = PTHREAD_MUTEX_STALLED;
+	attr->__mlibc_pshared = PTHREAD_PROCESS_PRIVATE;
+	attr->__mlibc_protocol = PTHREAD_PRIO_NONE;
 	return 0;
 }
 
-int pthread_mutexattr_destroy(pthread_mutexattr_t *) {
+int pthread_mutexattr_destroy(pthread_mutexattr_t *attr) {
 	SCOPE_TRACE();
-
+	memset(attr, 0, sizeof(*attr));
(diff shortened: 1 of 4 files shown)


options/posix/generic/pthread-stubs.cpp
```diff
@@ -688,67 +688,95 @@ int pthread_once(pthread_once_t *once, void (*function) (void)) {
...
+	attr->__mlibc_pshared = PTHREAD_PROCESS_PRIVATE;
+	attr->__mlibc_protocol = PTHREAD_PRIO_NONE;
 	return 0;
 }
 
-int pthread_mutexattr_destroy(pthread_mutexattr_t *) {
+int pthread_mutexattr_destroy(pthread_mutexattr_t *attr) {
 	SCOPE_TRACE();
-
+	memset(attr, 0, sizeof *attr);
```
ArsenArsen, line 700: we use function style sizeof

ArsenArsen (changes requested):

options/posix/generic/pthread-stubs.cpp
64: not at all consistently :^)

Geertiebear (approved):
LGTM.

Geertiebear:
Does this complete the mutexattr_t struct?

64:
> Does this complete the mutexattr_t struct?

We have
```c
struct __mlibc_mutex {
	unsigned int __mlibc_state;
	unsigned int __mlibc_recursion;
	unsigned int __mlibc_flags;
};
typedef struct __mlibc_mutex pthread_mutex_t;

struct  __mlibc_mutexattr {
	int __mlibc_type;
	int __mlibc_robust;
	int __mlibc_protocol;
	int __mlibc_pshared;
	int __mlibc_prioceiling;
};
typedef struct __mlibc_mutexattr pthread_mutexattr_t;
```

I guess it's just a question of which fields we can compress into `__mlibc_flags` here (I think all of them except `prioceiling`?)

64:
Oh disregard, yes `mutexattr_t` should be complete now but `mutex_t` is not necessarily (as per the above).

Geertiebear:
Then let's add a required field for prioceiling too, to avoid an ABI break in the future.

64:
> Then let's add a required field for prioceiling too, to avoid an ABI break in the future.

Done. I've also made the getprioceiling / setprioceiling functions return EINVAL (matching musl) since we don't implement that behaviour and it appears to be a pretty heavyweight feature.

Geertiebear (approved):
LGTM now then.

PR #604, opened 2022-07-01
options/posix: implement pthread_condattr

64:
Progress on #530.

### DIFF
--- a/options/posix/generic/pthread-stubs.cpp
+++ b/options/posix/generic/pthread-stubs.cpp
@@ -1069,41 +1069,55 @@ int pthread_mutex_consistent(pthread_mutex_t *) {
 // pthread_condattr and pthread_cond functions.
 // ----------------------------------------------------------------------------
 
-int pthread_condattr_init(pthread_condattr_t *) {
-	SCOPE_TRACE();
-
-	mlibc::infoLogger() << "mlibc: pthread_condattr_init() is not implemented correctly" << frg::endlog;
-	return 0;
-}
-int pthread_condattr_destroy(pthread_condattr_t *) {
-	SCOPE_TRACE();
-
-	mlibc::infoLogger() << "mlibc: pthread_condattr_destroy() is not implemented correctly" << frg::endlog;
+int pthread_condattr_init(pthread_condattr_t *attr) {
+	attr->__mlibc_pshared = PTHREAD_PROCESS_PRIVATE;
+	attr->__mlibc_clock = CLOCK_REALTIME;
 	return 0;
 }
 
-int pthread_condattr_getclock(const pthread_condattr_t *__restrict, clockid_t *__restrict) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
-}
-int pthread_condattr_setclock(pthread_condattr_t *, clockid_t) {
-	SCOPE_TRACE();
-
-	mlibc::infoLogger() << "mlibc: pthread_condattr_setclock() is not implemented correctly" << frg::endlog;
+int pthread_condattr_destroy(pthread_condattr_t *attr) {
+	memset(attr, 0, sizeof(*attr));
 	return 0;
 }
 
-int pthread_condattr_getpshared(const pthread_condattr_t *__restrict, int *__restrict) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
-}
-int pthread_condattr_setpshared(pthread_condattr_t *, int) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+int pthread_condattr_getclock(const pthread_condattr_t *__restrict attr,
+		clockid_t *__restrict clock) {
+	*clock = attr->__mlibc_clock;
+	return 0;
 }
 
-int pthread_cond_init(pthread_cond_t *__restrict cond, const pthread_condattr_t *__restrict) {
+int pthread_condattr_setclock(pthread_condattr_t *attr, clockid_t clock) {
+	if (clock != CLOCK_REALTIME && clock != CLOCK_MONOTONIC
+			&& clock != CLOCK_MONOTONIC_RAW && clock != CLOCK_REALTIME_COARSE
+			&& clock != CLOCK_MONOTONIC_COARSE && clock != CLOCK_BOOTTIME)
+		return EINVAL;
+
+	attr->__mlibc_clock = clock;
+	return 0;
+}
+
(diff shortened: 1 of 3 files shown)


options/posix/generic/pthread-stubs.cpp
```diff
@@ -887,42 +887,54 @@ int pthread_mutex_consistent(pthread_mutex_t *) {
...
 
-int pthread_cond_init(pthread_cond_t *__restrict cond, const pthread_condattr_t *__restrict) {
+int pthread_cond_init(pthread_cond_t *__restrict cond, const pthread_condattr_t *__restrict attr) {
 	SCOPE_TRACE();
 
+	auto clock = attr ? attr->__mlibc_clock : CLOCK_REALTIME;
+	auto pshared = attr ? attr->__mlibc_pshared : PTHREAD_PROCESS_PRIVATE;
+
+	// Since we don't implement these yet, set a flag to error later.
+	cond->__mlibc_flags = clock || pshared;
```
avdgrinten, line 936: How is this going to work if we implement it properly? Can we compress the clock into flags? AFAICT that's not possible since the user might pass the result of `pthread_getcpuclockid()` or similar functions.

64: Hmm, setclock() isn't allowed to accept a CPU-time clock, so that wouldn't work. But perhaps we should add a `clockid_t` field on `pthread_cond_t` just to be safe

64: Yeah, I think that a single unsigned int should be sufficient here. Even if we need to store more clockid_t values, we'd probably fine since we only need 1 bit to store whether the condition variable is process-shared or not.

64:
Holding this until #666 is in, so I can implement setclock properly.

options/posix/generic/pthread-stubs.cpp
64: I've added a clockid_t field on cond_t now.

options/posix/generic/pthread-stubs.cpp
```diff
@@ -1140,7 +1157,7 @@ int pthread_cond_timedwait(pthread_cond_t *__restrict cond, pthread_mutex_t *__r
 			// Note: mlibc::sys_clock_get is available unconditionally.
 			// TODO: allow clock to be configured via a condattr.
```
Geertiebear, line 1158: Remove this comment.

options/posix/include/pthread.h
```diff
@@ -112,13 +112,15 @@ struct __mlibc_mutex {
 typedef struct __mlibc_mutex pthread_mutex_t;
 
 struct  __mlibc_condattr_struct {
-	// TODO: the clock attribute needs to be supported here.
+	int __mlibc_pshared;
+	clockid_t __mlibc_clock;
```
Geertiebear, line 116: Are we sure that this now marks the struct as complete?

options/posix/include/pthread.h
```diff
@@ -112,13 +112,15 @@ struct __mlibc_mutex {
...
 };
 typedef struct __mlibc_condattr_struct pthread_condattr_t;
 
 struct  __mlibc_cond {
-	// TODO: the clock attribute needs to be supported here.
-	int __mlibc_seq;
+	unsigned int __mlibc_seq;
```
Geertiebear, line 121: Same here.

Geertiebear (changes requested):

options/posix/include/pthread.h
64: Yes, there's only set/get clock/pshared.

options/posix/include/pthread.h
64: This should be sufficient to support all functionality yes. But we might want to change our implementation of cond later.

options/posix/generic/pthread-stubs.cpp
```diff
@@ -1069,42 +1069,56 @@ int pthread_mutex_consistent(pthread_mutex_t *) {
...
+	return 0;
 }
-int pthread_condattr_setclock(pthread_condattr_t *, clockid_t) {
-	SCOPE_TRACE();
 
-	mlibc::infoLogger() << "mlibc: pthread_condattr_setclock() is not implemented correctly" << frg::endlog;
+int pthread_condattr_setclock(pthread_condattr_t *attr, clockid_t clock) {
+	if (clock != CLOCK_REALTIME && clock != CLOCK_MONOTONIC
+			&& clock != CLOCK_MONOTONIC_RAW && clock != CLOCK_REALTIME_COARSE
+			&& clock != CLOCK_MONOTONIC_COARSE && clock != CLOCK_BOOTTIME)
```
ArsenArsen, line 1092: dedent this a bit

options/posix/generic/pthread-stubs.cpp
```diff
@@ -1162,9 +1178,9 @@ int pthread_cond_timedwait(pthread_cond_t *__restrict cond, pthread_mutex_t *__r
 				__ensure(timeout.tv_nsec >= 0);
 			}
 
-			e = mlibc::sys_futex_wait(&cond->__mlibc_seq, seq, &timeout);
+			e = mlibc::sys_futex_wait((int *)&cond->__mlibc_seq, seq, &timeout);
 		} else {
-			e = mlibc::sys_futex_wait(&cond->__mlibc_seq, seq, nullptr);
+			e = mlibc::sys_futex_wait((int *)&cond->__mlibc_seq, seq, nullptr);
```
ArsenArsen, line 1183: why were these made?

ArsenArsen (approved):

options/posix/generic/pthread-stubs.cpp
64: I made the seq unsigned for consistency with the other seqs. And really, there's no reason for the seq to be signed to begin with.

ArsenArsen: ah, i missed that, okay. though it might be wise to consider a specific cast instead

64: Ideally that'd be fixed at the same time as the other ones are fixed. But IMO that can wait until we do a big clang-tidy PR or similar cleanup.

Geertiebear (approved):

64:
I'll merge this now, but indentation and cast style can be decided upon and changed later by clang-format / clang-tidy PRs.

Issue #12, opened 2019-08-07
ansi: Fix double close of stdio on exit


avdgrinten:
mlibc always closes `stdin`, `stdout` and `stderr` on exit. This can fail (and, currently, crash mlibc) if the program decided to close those streams itself.

PR #608, opened 2022-07-01
tests/ansi: check that fclose(stdout) doesn't crash on exit

64:
This was fixed earlier, but it doesn't hurt to add a test for this case. See #12.


Geertiebear (approved):
LGTM.

PR #609, opened 2022-07-01
ansi: add RE_DUP_MAX and CHARCLASS_NAME_MAX

64:
Commit taken from the Vinix patches.

### DIFF
--- a/options/ansi/include/limits.h
+++ b/options/ansi/include/limits.h
@@ -32,6 +32,9 @@
 #define LINE_MAX 4096
 #define PIPE_BUF 4096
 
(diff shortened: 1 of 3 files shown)


avdgrinten:
Are these defined by any standard?

64:
They are both regex-related and defined by posix.

Geertiebear (approved):
LGTM then.

ArsenArsen:
One of these is sysconf-able, that should have a matching _SC and respective handling (also, the commit message doesn't explain anything)

ArsenArsen (approved):

PR #618, opened 2022-07-04
options/posix: use rwlock to protect global key array

64:
As per the discussion on Discord.

This allows `pthread_getspecific` to call itself recursively, which may happen under ASan (which hooks `__tls_get_addr` and calls `pthread_getspecific`). It also allows keys to be created and destroyed inside a key destructor.

### DIFF
--- a/options/posix/generic/pthread-stubs.cpp
+++ b/options/posix/generic/pthread-stubs.cpp
@@ -195,7 +195,9 @@ namespace {
 		PTHREAD_KEYS_MAX
 	> key_locals_{};
 
-	FutexLock key_mutex_;
+	// This protects accesses to key_globals_ (not key_locals_).
+	// TODO: Use a type that works with frigg (to avoid manual locking).
+	pthread_rwlock_t key_global_mutex_;
 }
 
 int pthread_exit(void *ret_val) {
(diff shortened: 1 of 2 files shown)


avdgrinten:
What exactly is the issue that is solved here?

qookei:
`pthread_getspecific` ends up calling into __tls_get_addr, and ASan intercepts `__tls_get_addr` calls, and ASan's interceptor calls into `pthread_getspecific` again (if I understood right when discussing this on Discord).

64:
> What exactly is the issue that is solved here?

There are two issues:
1) Before, we didn't use any lock when calling key destructors.
2) When `pthread_getspecific` is called, ASan intercepts `__tls_get_addr`. But the interceptor itself calls `pthread_getspecific` which causes a deadlock (since we acquire a mutex in it). Changing this to use a rwlock works around this issue. AFAICT it's unfortunate that we run into this issue with ASan, and it seems to occur because a) we use a lock in getspecific, unlike other libcs b) our getspecific uses thread_local storage which invokes tls_get_addr, whereas other libcs seem to store key data in the TCB. So we get quite unlucky, but there's also not an obvious workaround on the ASan side.

64:
I also noted on Discord that it's possible to implement `getspecific` / `setspecific` without any locking, and that'd probably be preferable. But it requires some more intrusive changes (adding a list of all TCBs, and looping over them in `pthread_key_delete`), so this solution should do for now.

avdgrinten:
Well, we can certainly avoid using public pthread functions (like the rwlock ones) from `pthread_getspecific` as we have our own locks already. The `__tls_get_addr` issue could be fixed by reserving a pointer in the TCB to point to the array that we need.

64:
We do not have our own rwlock implementation internally, only a normal mutex (FutexLock).

64:
Closing this for now as we decided that it's better to try and make our `pthread_getspecific` not call `__tls_get_addr`.

PR #619, opened 2022-07-04
rtdl: don't load the same SONAME twice

### LINKED ISSUE #585: rtdl: dynamic linker may load libraries more than once
This can easily occur if e.g a library is loaded first by `path/to/libfoo.so` and then `libfoo.so`, where the latter name is resolved via an RPATH.

64:
Fixes #585. Also fixes some hardcoding of paths in the `dl_iterate_phdr` test.


Geertiebear (approved):
LGTM.

PR #620, opened 2022-07-04
options/posix: move local pthread keys into TCB

64:
Followed on from #618. This avoids a call to `__tls_get_addr` inside `pthread_getspecific` / `pthread_setspecific`, which should allow ASan to function. We also add some locking that was missed before.

~~Blocked on #622.~~


Geertiebear (approved):
LGTM.

qookei (approved):

PR #622, opened 2022-07-05
rtdl: zero .tbss and correctly align TCB

64:
Previously, the TCB was not correctly aligned:
* On AArch64 and RISC-V, the assumption `tp + sizeoc(Tcb) == tlsData` would break if sizeof(Tcb) was not a multiple of `tlsMaxAlignment`. This meant that adding members to `Tcb` would cause a breakage.
* On x86_64, we relied on the default alignment of the memory allocator to provide sufficient alignment for the TCB and TLS data (though this caused no breakage in practice).

This also fixes a bug whereby the `.tbss` was not being zeroed on the main thread.

### DIFF
--- a/options/internal/riscv64-include/mlibc/thread.hpp
+++ b/options/internal/riscv64-include/mlibc/thread.hpp
@@ -2,15 +2,16 @@
 
 #include <stdint.h>
 #include <mlibc/tcb.hpp>
+#include <bits/ensure.h>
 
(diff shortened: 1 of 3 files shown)


Geertiebear (commented):
LGTM except for one nitpick comment.

Geertiebear (approved):
LGTM.

PR #624, opened 2022-07-05
options/bsd: Add fstab.h header

Dennisbonke:

### DIFF
--- a/options/bsd/include/fstab.h
+++ b/options/bsd/include/fstab.h
@@ -1,0 +1,23 @@
+#ifndef _FSTAB_H
+#define _FSTAB_H
+
+#define _PATH_FSTAB "/etc/fstab"
+#define FSTAB "/etc/fstab"
+
+#define FSTAB_RW "rw"
+#define FSTAB_RQ "rq"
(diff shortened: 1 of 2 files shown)


options/bsd/include/fstab.h
```diff
@@ -0,0 +1,23 @@
...
+	char *fs_spec;
+	char *fs_file;
+	char *fs_vfstype;
+	char *fs_mntops;
+	const char *fs_type;
+	int fs_freq;
+	int fs_passno;
+};
+
+#endif // _FSTAB_H
```
ArsenArsen, line 23: Don't add C++-style comments in public interfaces.

ArsenArsen (changes requested):
What programs include this header in order to not use getfsent and friends?

Dennisbonke:
> What programs include this header in order to not use getfsent and friends?

`libsanitizers` makes use of this header.

ArsenArsen:
Looks like it's only used to determine the size of this header (couldn't the gcc devs come up with a better way to enumerate the systems structs?)

Geertiebear (approved):
LGTM.

PR #628, opened 2022-07-05
options/ansi: Fix signgam declaration

Dennisbonke:
`libsanitizers` was complaining about `signgam`, and while we did what musl does, this seems to please `libsanitizers`.


Dennisbonke:
> This is more of a hack and not a proper fix. We should investigate why applications weren't able to link against `signgam` in libc rather than doing this. I'm not sure why the `weak_alias` doesn't allow that.

Agree, but I have no clue where to start. Help in this matter would be greatly appreciated.

64 (approved):

PR #632, opened 2022-07-05
options/linux: Add mallopt

Dennisbonke:

### DIFF
--- a/options/linux/generic/malloc.cpp
+++ b/options/linux/generic/malloc.cpp
@@ -10,3 +10,7 @@ void *memalign(size_t, size_t) {
 	__ensure(!"Not implemented");
 	__builtin_unreachable();
 }
+
+int mallopt(int, int) {
+	return 0;
+}
--- a/options/linux/include/malloc.h
+++ b/options/linux/include/malloc.h
@@ -17,6 +17,7 @@ void *memalign(size_t, size_t);
 
 // GNU extension
 size_t malloc_usable_size(void *ptr);
+int mallopt(int param, int value);
 
 #ifdef __cplusplus
 }


64:
What needs this?

options/linux/generic/malloc.cpp
```diff
@@ -10,3 +10,7 @@ void *memalign(size_t, size_t) {
 	__ensure(!"Not implemented");
 	__builtin_unreachable();
 }
+
+int mallopt(int, int) {
+	return 1;
```
ArsenArsen, line 15: Return zero instead. This didn't succeed (it isn't implemented). Funnily enough, errno isn't set by this function, which doesn't even let us mention that it's not implemented via ENOSYS. (Also, wtf? I've never seen this function in actual use)

64:
I agree it doesn't make sense for us to implement this. It should be easy to patch programs which need it.

ArsenArsen:
I don't think it's accurate that this is an implementation detail, though. It's definitely a public API, just one we don't really have a match for.

qookei:
It seems like more than just glibc ever implemented it:
```
CONFORMING TO
       This  function is not specified by POSIX or the C standards.  A similar
       function exists on many System V derivatives, but the range  of  values
       for  param  varies  across systems.  The SVID defined options M_MXFAST,
       M_NLBLKS, M_GRAIN, and M_KEEP, but only the first of  these  is  imple‐
       mented in glibc.
```

Geertiebear:
Maybe just add a warning that this is not implemented.

Geertiebear:
I think we've reached a consensus on discord that we are dropping this change, unless we get a use for it in mlibc. Closing.

PR #634, opened 2022-07-05
options/posix: Overhaul glob.h

Dennisbonke:
The definitions in glob.h (GLOB_APPEND and friends) now match glibc and musl.
Some missing definitions were added.
Struct glob_t was updated to match glibc.


ArsenArsen (changes requested):
This is an abi-break? Also, are these even handled?

Dennisbonke:
> This is an abi-break? Also, are these even handled?

Likely yes, and unsure if they are handled, but `libsanitizers` was complaining, so I added them.

avdgrinten:
Matching glibc or musl should be irrelevant here, these definitions are provided by libc; they are never passed to sysdeps. Why is this change done?

avdgrinten:
That's in `#ifdef SANITIZER_LINUX`. The same file also hardcodes `FILE` etc.

ArsenArsen:
At best, the rest of the glob API should be implemented.

PR #635, opened 2022-07-05
options/posix: Alias some structs to their struct64 counterpart

Dennisbonke:
`libsanitizers` expect `rlimit64`, and we don't provide that one. Alias our `struct rlimit` to `rlimit64` to please `libsanitizers`.

### DIFF
--- a/options/posix/include/sys/resource.h
+++ b/options/posix/include/sys/resource.h
@@ -1,6 +1,7 @@
 #ifndef _SYS_RESOURCE_H
 #define _SYS_RESOURCE_H
 
+#include <limits.h>
 #include <abi-bits/pid_t.h>
 #include <abi-bits/resource.h>
 #include <bits/posix/id_t.h>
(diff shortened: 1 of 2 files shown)


Geertiebear:
Are these implementations correct on all arches?

ArsenArsen:
If we change any places where we assume `sizeof(long) * CHAR_BITS == 64`, yes

Geertiebear:
Okay then for now I'd be much more comfortable merging this with a proper static_assert to ensure that this is actually true. Silently breaking on an arch where this equivalence does not hold is not acceptable.

Geertiebear:
Yeah, that will suffice. That will break when porting to other arches, like x86 for example, where the struct layouts are different.

Dennisbonke:
@Geertiebear fixed your static assert, is it good now?

Dennisbonke:
Closing as we're not going to be porting `libsanitizers` anyway.

PR #636, opened 2022-07-05
abis: Define struct msqid_ds

Dennisbonke:

### DIFF
--- a/abis/linux/msg.h
+++ b/abis/linux/msg.h
@@ -1,0 +1,39 @@
+#ifndef _ABIBITS_MSG_H
+#define _ABIBITS_MSG_H
+
+#include <sys/ipc.h>
+#include <bits/ansi/time_t.h>
+#include <bits/types.h>
+#include <abi-bits/pid_t.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#if defined(__i386__)
+typedef __mlibc_uint64 msglen_t;
+typedef __mlibc_uint64 msgqnum_t;
+#else
+typedef unsigned long msglen_t;
+typedef unsigned long msgqnum_t;
+#endif
+
+struct msqid_ds {
+	struct ipc_perm msg_perm;
+	time_t msg_stime;
+	time_t msg_rtime;
+	time_t msg_ctime;
+	unsigned long msg_cbytes;
+	msgqnum_t msg_qnum;
+	msglen_t msg_qbytes;
+	pid_t msg_lspid;
+	pid_t msg_lrpid;
+	unsigned long __unused[2];
+};
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* _ABIBITS_MSG_H */
+
--- a/options/posix/include/sys/msg.h
+++ b/options/posix/include/sys/msg.h
@@ -1,7 +1,7 @@
 #ifndef _SYS_MSG_H
 #define _SYS_MSG_H
 
-#include <sys/ipc.h>
+#include <abi-bits/msg.h>
 
 #ifdef __cplusplus
 extern "C" {
--- a/sysdeps/aero/include/abi-bits/msg.h
+++ b/sysdeps/aero/include/abi-bits/msg.h
@@ -1,0 +1,1 @@
+../../../../abis/linux/msg.h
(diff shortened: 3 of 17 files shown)


options/posix/include/sys/msg.h
```diff
@@ -2,13 +2,32 @@
...
+	time_t msg_stime;
+	time_t msg_rtime;
+	time_t msg_ctime;
+	unsigned long __msg_cbytes;
+	msgqnum_t msg_qnum;
+	msglen_t msg_qbytes;
+	pid_t msg_lspid;
+	pid_t msg_lrpid;
+	unsigned long __glibc_reserved4;
+	unsigned long __glibc_reserved5;
```
ArsenArsen, line 28: Is this in violation of the GPL?

ArsenArsen (changes requested):
Is this correct? Also, isn't this an abi-bit?

options/posix/include/sys/msg.h
Dennisbonke: Can you copyright struct definitions? No clue, but if it's a problem, we can rename them.

Dennisbonke:
> Is this correct? Also, isn't this an abi-bit?

Don't see why it isn't correct, and the distinction between needs to be abi-bit and can go into the main header remains unclear for me at times. If the mlibc abi matches the linux abi (aka here) then why do the extra work?

ArsenArsen:
Because this doesn't seem to match between Linux architectures

options/posix/include/sys/msg.h
Dennisbonke: Well that's gonna be fun if I need struct definitions and musl doesn't have them. Can't use different names either as programs rely on them (and no, they aren't standard per se). I see no problems copying definitions, however, copying function implementations is indeed not done.

Geertiebear: Why are the reserved fields copied over anyways?

options/posix/include/sys/msg.h
```diff
@@ -2,13 +2,32 @@
...
+typedef unsigned long msgqnum_t;
+
 int msgget(key_t, int);
 
+struct msqid_ds {
+	struct ipc_perm msg_perm;
+	time_t msg_stime;
+	time_t msg_rtime;
+	time_t msg_ctime;
+	unsigned long __msg_cbytes;
```
avdgrinten, line 22: This field is called `msg_cbytes` according to the man page.

options/posix/include/sys/msg.h
avdgrinten: Just rename them to `unsigned long __unused[2]` as musl does.

abis/linux/msg.h
```diff
@@ -0,0 +1,33 @@
...
+#include <sys/ipc.h>
+#include <bits/ansi/time_t.h>
+#include <abi-bits/pid_t.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef unsigned long msglen_t;
+typedef unsigned long msgqnum_t;
```
ArsenArsen, line 13: should we make these fixed-width? how do they differ across the arches we have?

ArsenArsen (commented):
`typedef __syscall_ulong_t msgqnum_t;` is the glibc definition for this. does __syscall_ulong_t always match unsigned long?

abis/linux/msg.h
ArsenArsen: okay, I looked around the logic here a bit, I think these are always `unsigned long int`, but they depend on `__WORDSIZE`, and I'm not sure how we handle that case

Geertiebear: We can copy the logic that depends on __WORDSIZE in a static assert.

Geertiebear:
This PR currently has merge conflicts.

Dennisbonke:
I've fixed the conflicts, can this get another review?

no92:
Rebased onto current master, dropping vinix and adding lyre.

abis/linux/msg.h
Geertiebear: This comment is still not addressed, maybe @ArsenArsen can link to the logic in question so it's easier to copy over?

no92: Looking through glibc behavior, these are always defined to be `unsigned long int`, except for i386, where they are `uint64_t`.

no92:
Fixed the typedefs for `msglen_t` and `msgqnum_t`.

Geertiebear (approved):
LGTM.

PR #638, opened 2022-07-05
options/posix: Partially implement gai_strerror

Dennisbonke:
This is not really how it's supposed to work, but instead of crashing, we now at least get the error code back.


options/posix/generic/netdb-stubs.cpp
```diff
@@ -50,9 +51,10 @@ void freeaddrinfo(struct addrinfo *ptr) {
 	}
 }
 
-const char *gai_strerror(int) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+const char *gai_strerror(int code) {
+	static char buffer[128];
+	snprintf(buffer, sizeof buffer, "Unknown error (%d)", code);
```
ArsenArsen, line 56: `sizeof(...)`, not `sizeof ...`

ArsenArsen (changes requested):
This buffer should be thread-local

ArsenArsen (approved):

64 (approved):

PR #639, opened 2022-07-05
options/posix: Add dlvsym

Dennisbonke:
The implementation is only partial, as the version argument is ignored. Thus, `dlvsym` behaves just like `dlsym`.

### DIFF
--- a/options/posix/generic/dlfcn-stubs.cpp
+++ b/options/posix/generic/dlfcn-stubs.cpp
@@ -37,6 +37,13 @@ void *dlsym(void *__restrict handle, const char *__restrict string) {
 	return __dlapi_resolve(handle, string, ra);
 }
 
+[[gnu::noinline]]
+void *dlvsym(void *__restrict handle, const char *__restrict string, const char *__restrict version) {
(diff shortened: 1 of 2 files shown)


options/posix/generic/dlfcn-stubs.cpp
```diff
@@ -37,6 +37,12 @@ void *dlsym(void *__restrict handle, const char *__restrict string) {
 	return __dlapi_resolve(handle, string, ra);
 }
 
+void *dlvsym(void *__restrict handle, const char *__restrict string, const char *__restrict) {
```
64, line 40: This should be marked `[[gnu::noinline]]`.

options/posix/generic/dlfcn-stubs.cpp
```diff
@@ -37,6 +37,12 @@ void *dlsym(void *__restrict handle, const char *__restrict string) {
...
+	mlibc::infoLogger() << "mlibc: dlvsym ignores version" << frg::endlog;
```
64, line 41: You might as well print the version string too.

64 (changes requested):

PR #641, opened 2022-07-05
sysdeps/linux: Include unistd.h in sys/syscall.h

Dennisbonke:
I got compilation failures if this header is not included, I'm unsure about whether this is correct, so feedback is appreciated.

### DIFF
--- a/sysdeps/linux/include/sys/syscall.h
+++ b/sysdeps/linux/include/sys/syscall.h
@@ -2,5 +2,8 @@
 #define _SYS_SYSCALL_H
 
 #include <abi-bits/arch-syscall.h>
+#if __has_include(<asm/unistd.h>)
+#include <asm/unistd.h>
+#endif
 
 #endif // _SYS_SYSCALL_H


ArsenArsen:
This is not correct.

64:
> This is not correct.

In what way?

Dennisbonke:
> We're supposed to include <asm/unistd.h>, which then includes unistd_64.h on x86_64, when `__X32__` isn't set, or something like that. I don't remember the details. This header also doesn't exist outside of x86_64 64bit.

Makes sense, I'll fix it.

sysdeps/linux/include/sys/syscall.h
```diff
@@ -2,5 +2,8 @@
 #define _SYS_SYSCALL_H
 
 #include <abi-bits/arch-syscall.h>
+#if __has_include("asm/unistd.h")
```
ArsenArsen, line 5: `__has_include(<asm/unistd.h>)`

ArsenArsen (changes requested):

PR #642, opened 2022-07-05
sysdeps/linux: Add sys_yield

Dennisbonke:

### DIFF
--- a/sysdeps/linux/generic/sysdeps.cpp
+++ b/sysdeps/linux/generic/sysdeps.cpp
@@ -451,6 +451,10 @@ int sys_sysinfo(struct sysinfo *info) {
 	return 0;
 }
 
+void sys_yield() {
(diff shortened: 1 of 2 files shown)


ArsenArsen:
Maybe we should adjust this function to match POSIX (return int, never fail on Linux)

64:
> Maybe we should adjust this function to match POSIX (return int, never fail on Linux)

Yes, I agree

avdgrinten:
Why? `void` expresses correctly that the function can't fail.

ArsenArsen:
> Why? `void` expresses correctly that the function can't fail.

Except on systems where it can (permitted by POSIX)

avdgrinten:
I think POSIX is overly pessimistic here; an empty function is a valid implementation for `sched_yield()`.

64 (approved):

PR #643, opened 2022-07-05
options/rtdl: Handle TPOFF64 with reloc->r_addend being not null

Dennisbonke:
The sanitizers from `libsanitizers` can produce symbols that trip this ensure, so we should fix it.


Geertiebear (approved):
LGTM.

PR #644, opened 2022-07-05
options/rtdl: Ignore the first item in __dlapi_iterate_phdr

### LINKED ISSUE #579: rtdl: `dl_iterate_phdr` should bail when the callback returns non-zero, and should report the current executable as the empty string
Needed for ASan.

Dennisbonke:
This matches the behavior of glibc and pleases ASan.

Fixes #579

### DIFF
--- a/options/rtdl/generic/main.cpp
+++ b/options/rtdl/generic/main.cpp
@@ -570,6 +570,12 @@ int __dlapi_iterate_phdr(int (*callback)(struct dl_phdr_info *, size_t, void*), 
 		struct dl_phdr_info info;
 		info.dlpi_addr = object->baseAddress;
 		info.dlpi_name = object->name.data();
+
+		if(object->isMainObject) {
+			info.dlpi_name = "";
+		} else {
+			info.dlpi_name = object->name.data();
+		}
 		info.dlpi_phdr = static_cast<ElfW(Phdr)*>(object->phdrPointer);
 		info.dlpi_phnum = object->phdrCount;
 		info.dlpi_adds = rtsCounter;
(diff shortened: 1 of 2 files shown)


64:
The arguments to `strstr` in `tests/rtdl/dl_iterate_phdr/test.c` need to be swapped for the test to pass.

Dennisbonke:
> The arguments to `strstr` in `tests/rtdl/dl_iterate_phdr/test.c` need to be swapped for the test to pass.

Noted, will fix soon.

avdgrinten:
I don't understand what this commit does.

avdgrinten:
Why do we check for the first object instead of checking for the executable?

64:
> Why do we check for the first object instead of checking for the executable?

They ought to be the same thing, but it's true that would be a better implementation.

options/rtdl/generic/main.cpp
```diff
@@ -548,10 +548,17 @@ int __dlapi_reverse(const void *ptr, __dlapi_symbol *info) {
 extern "C" [[ gnu::visibility("default") ]]
 int __dlapi_iterate_phdr(int (*callback)(struct dl_phdr_info *, size_t, void*), void *data) {
 	int last_return = 0;
+	int amount = 0;
 	for (auto object : globalScope->_objects) {
 		struct dl_phdr_info info;
 		info.dlpi_addr = object->baseAddress;
 		info.dlpi_name = object->name.data();
+
+		if(amount == 0) {
```
64, line 557: Instead of this `amount` variable, we should check `object->isMainObject`

64 (changes requested):

64:
Also merge conflict

64 (approved):

Geertiebear (approved):
LGTM.

Geertiebear (changes requested):
Actually, do we test for this behaviour, and if not, can that be added?

64:
Yeah actually, there's some commented out lines we can re-enable in the test.

64 (approved):

Geertiebear (approved):

PR #646, opened 2022-07-05
options/glibc: Add some more network headers

Dennisbonke:
`libsanitizers` makes use of these headers, and glibc provides them.

### DIFF
--- a/options/glibc/include/net/if_ppp.h
+++ b/options/glibc/include/net/if_ppp.h
@@ -1,0 +1,23 @@
+#ifndef _NET_IF_PPP_H
+#define _NET_IF_PPP_H
+
+#include <mlibc-config.h>
+
+#if defined(__linux__) || defined(__MLIBC_LINUX_OPTION)
+#include <asm/ioctl.h>
+#include <linux/ppp_defs.h>
+
+#define PPPIOCGFLAGS _IOR('t', 90, int)
+#define PPPIOCSFLAGS _IOW('t', 89, int)
+#define PPPIOCGASYNCMAP _IOR('t', 88, int)
+#define PPPIOCSASYNCMAP _IOW('t', 87, int)
+#define PPPIOCGUNIT _IOR('t', 86, int)
+#define PPPIOCSMRU _IOW('t', 82, int)
+#define PPPIOCSMAXCID _IOW('t', 81, int)
+#define PPPIOCGXASYNCMAP _IOR('t', 80, ext_accm)
+#define PPPIOCSXASYNCMAP _IOW('t', 79, ext_accm)
+#define PPPIOCGDEBUG _IOR('t', 65, int)
+#define PPPIOCSDEBUG _IOW('t', 64, int)
+#endif
+
+#endif /* _NET_IF_PPP_H */
--- a/options/glibc/include/netipx/ipx.h
+++ b/options/glibc/include/netipx/ipx.h
@@ -1,0 +1,19 @@
+#ifndef _NETIPX_IPX_H
+#define _NETIPX_IPX_H
+
+#include <mlibc-config.h>
+
+typedef struct ipx_config_data {
+	unsigned char ipxcfg_auto_select_primary;
+	unsigned char ipxcfg_auto_create_interfaces;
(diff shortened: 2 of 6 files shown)


options/glibc/include/net/if_ppp.h
```diff
@@ -0,0 +1,23 @@
...
+#define PPPIOCGASYNCMAP _IOR('t', 88, int)
+#define PPPIOCSASYNCMAP _IOW('t', 87, int)
+#define PPPIOCGUNIT     _IOR('t', 86, int)
+#define PPPIOCSMRU      _IOW('t', 82, int)
+#define PPPIOCSMAXCID   _IOW('t', 81, int)
+#define PPPIOCGXASYNCMAP _IOR('t', 80, ext_accm)
+#define PPPIOCSXASYNCMAP _IOW('t', 79, ext_accm)
+#define PPPIOCGDEBUG    _IOR('t', 65, int)
+#define PPPIOCSDEBUG    _IOW('t', 64, int)
+#endif
```
ArsenArsen, line 21: This is written in mixed style, pick one

options/glibc/include/net/if_ppp.h
```diff
@@ -0,0 +1,23 @@
+#ifndef _NET_IF_PPP_H
+#define _NET_IF_PPP_H
+
+#include <mlibc-config.h>
+
+#if defined __linux__ || defined __MLIBC_LINUX_OPTION
```
ArsenArsen, line 6: `defined(...)` is preferable to `defined ...`

options/glibc/include/netipx/ipx.h
```diff
@@ -0,0 +1,19 @@
...
+
+#if defined __linux__ || defined __MLIBC_LINUX_OPTION
+#include <linux/sockios.h>
+
+#define SIOCAIPXITFCRT (SIOCPROTOPRIVATE)
+#define SIOCAIPXPRISLT (SIOCPROTOPRIVATE + 1)
+#define SIOCIPXCFGDATA (SIOCPROTOPRIVATE + 2)
+#endif
+
+#endif // _NETIPX_IPX_H
```
ArsenArsen, line 19: Don't include C++-style comments in public headers

options/glibc/meson.build
```diff
@@ -43,10 +43,26 @@ if not no_headers
...
+		subdir: 'netipx'
+	)
+	install_headers(
+		'include/netrom/netrom.h',
+		subdir: 'netrom'
+	)
 	install_headers(
 		'include/netinet/in_systm.h',
+		'include/netinet/icmp6.h',
+		'include/netinet/ether.h',
```
ArsenArsen, line 65: Where do these come from?

ArsenArsen (changes requested):
Also, your commit is doubled up for some reason (and the commit messages don't explain anything)

Dennisbonke:
> Also, your commit is doubled up for some reason (and the commit messages don't explain anything)

Not entirely true, they are two separate commits, that I forgot to merge into one. I'll look into a better commit message.

options/glibc/meson.build
Dennisbonke: Likely me messing up a merge, I'll remove them.

options/glibc/include/net/if_ppp.h
Dennisbonke: Will fix.

options/glibc/include/net/if_ppp.h
Dennisbonke: We should write that down somewhere, as that wasn't apparent to me. I'll change it.

options/glibc/include/netipx/ipx.h
Dennisbonke: Makes sense, but we should probably check the entire codebase for that then.

ArsenArsen: Will do at a later date (as mentioned). The only reason I'm pedantic about this is because it came up as a problem when dealing with C89 programs

Geertiebear (approved):
LGTM.

PR #648, opened 2022-07-05
abis: deprecate some headers in the mlibc ABI

### LINKED ISSUE #257: posix: Review ABI of siginfo_t

### LINKED ISSUE #430: termios: Add missing special character defines
We are missing a few special character defines like `VDISCARD`. This requires extending the `NCCS` field in `struct termios` which is an ABI break.

### LINKED ISSUE #466: Review the ABI of WEXITSTATUS and friends

64:
As discussed on Discord. This deprecates some headers in the `mlibc` ABI in favour of the Linux ABI. This partially reverts #580 (a deprecation notice is added instead).

Once we are happy with this PR, we should ping all the sysdep owners to see if they are happy to have their sysdeps moved over to the Linux ABI too, because these changes will require some modifications to their kernels.

Fixes #430, fixes #257, fixes #466.

### DIFF
--- a/ABI_BREAKS.md
+++ b/ABI_BREAKS.md
@@ -6,6 +6,6 @@ This document lists the ABI breaks that were made in each mlibc major version.
 
 - [#452](https://github.com/managarm/mlibc/pull/452): The functions `FD_{CLR,ISSET,SET,ZERO}` were renamed to `__FD_{CLR,ISSET,SET,ZERO}` and replaced by macros to match Wine's assumptions.
 - [#511](https://github.com/managarm/mlibc/pull/511): Musl's regex engine was added, implementing `regcomp` and `regexec`. This required some changes to the `regex_t` struct.
-- [#504](https://github.com/managarm/mlibc/pull/504), [#580](https://github.com/managarm/mlibc/pull/580): In both the mlibc and Linux ABIs, a `domainname` member was added to `struct utsname`, which is a glibc extension.
+- [#504](https://github.com/managarm/mlibc/pull/504): A `domainname` member was added to `struct utsname`, which is a glibc extension.
 - [#311](https://github.com/managarm/mlibc/pull/311): Added all necessary fields in `pthread_attr_t` required for implementing all `pthread_attr` functions.
 - [#652](https://github.com/managarm/mlibc/pull/652): The ABI of `struct statfs` and `struct statvfs` was changed to match Linux. `socklen_t` was also changed from `unsigned long` to `unsigned int`.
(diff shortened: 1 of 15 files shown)


avdgrinten (approved):

Geertiebear (approved):
LGTM.

64:
@Andy-Python-Programmer @mintsuki @streaksu @fido2020 @Menotdan

We're planning on deprecating some parts of the mlibc ABI: the `signal.h`, `wait.h`, `termios.h` and `utsname.h` headers (see #430, #257 and #466 for motivation). Instead we will recommend new ports to use the equivalent headers in the Linux ABI, but switching from one to the other will require some small kernel tweaks:

1. Any signal handling code will need to be updated to use the new members of `sigaction` and `siginfo_t`.
2. The exit code of processes will need to be constructed via the new `W_EXITCODE(ret, signal)` macro.

**Note that these changes will be applied to the abi-break branch, and won't come into effect until the next ABI-breaking release.**

You are of course free to use whichever ABI you want. Please could you indicate (via thumbs up/down reaction or comment) whether you would like your sysdeps to have those headers switched over to the Linux ABI, or not?

64:
> Please could you indicate (via thumbs up/down reaction or comment) whether you would like your sysdeps to have those headers switched over to the Linux ABI, or not? (and if yes, I'll add a commit to this PR doing so)

I'm going to merge this (tomorrow), but the offer is still open.

Andy-Python-Programmer:
Sorry for the late reply!

> We're planning on deprecating some parts of the mlibc ABI: the signal.h, wait.h, termios.h and utsname.h headers (see https://github.com/managarm/mlibc/issues/430, https://github.com/managarm/mlibc/issues/257 and https://github.com/managarm/mlibc/issues/466 for motivation). Instead we will recommend new ports to use the equivalent headers in the Linux ABI

Since its the above mlibc ABI headers are deprecated, updating the Aero sysdeps to use the Linux ABI headers for the above seems a considerable option. Considering that the Linux ABI headers are more stable and wont require many ABI-breaks in future (if any). 

So its a yes for switching the `signal.h`, `wait.h`, `termios.h` and `utsname.h` headers to the equivalent headers in the Linux ABI for Aero.

Thanks!

Geertiebear (commented):
I am going to merge this now, sysdeps have been given enough notice and the ``aero`` sysdeps have been changed.

Issue #614, opened 2022-07-03
options: allow logging and debugging to be configured by environment variables


64:
Some logging, like that in `rtdl`, should be controlled by environment variables (so that we don't have to recompile the program to enable it).

#598 implements this for `malloc` et al., but we ought to do this in a way that is safe w.r.t concurrent modifications of the environment. For example, we could have a global `libcConfig` struct that we initialise in `ld.so` from the environment at program startup.

PR #650, opened 2022-07-06
options/{internal,rtdl}: add GlobalConfig and RtdlConfig

64:
This PR adds a mechanism to pass global program configuration data between `ld.so` and `libc.so`, laying the groundwork for #614.

### DIFF
--- a/options/internal/generic/global-config.cpp
+++ b/options/internal/generic/global-config.cpp
@@ -1,0 +1,27 @@
+#include <stdlib.h>
+#include <string.h>
+#include <mlibc/global-config.hpp>
+
+namespace mlibc {
+
+struct GlobalConfigGuard {
+	GlobalConfigGuard();
+};
+
+GlobalConfigGuard guard;
+
+GlobalConfigGuard::GlobalConfigGuard() {
+	// Force the config to be created during initialization of libc.so.
+	mlibc::globalConfig();
+}
+
+static bool envEnabled(const char *env) {
+	auto value = getenv(env);
+	return value && *value && *value != '0';
+}
+
+GlobalConfig::GlobalConfig() {
+	debugMalloc = envEnabled("MLIBC_DEBUG_MALLOC");
+}
+
+}
--- a/options/internal/include/mlibc/global-config.hpp
+++ b/options/internal/include/mlibc/global-config.hpp
@@ -1,0 +1,19 @@
+#ifndef MLIBC_GLOBAL_CONFIG
+#define MLIBC_GLOBAL_CONFIG
+
+namespace mlibc {
+
+struct GlobalConfig {
+	GlobalConfig();
+	
+	bool debugMalloc;
+};
+
+inline const GlobalConfig &globalConfig() {
+	static GlobalConfig cached;
+	return cached;
+}
+
+}
+
+#endif // MLIBC_GLOBAL_CONFIG
--- a/options/rtdl/generic/main.cpp
+++ b/options/rtdl/generic/main.cpp
@@ -7,6 +7,7 @@
 #include <abi-bits/auxv.h>
 #include <mlibc/debug.hpp>
 #include <mlibc/rtdl-sysdeps.hpp>
+#include <mlibc/rtdl-config.hpp>
 #include <mlibc/rtdl-abi.hpp>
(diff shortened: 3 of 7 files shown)


64:
ABI checker fails because I've removed `__dlapi_secure_required`, but that's in `ld.so` and shouldn't be an issue.

options/internal/include/mlibc/global-config.hpp
```diff
@@ -0,0 +1,18 @@
+#ifndef MLIBC_GLOBAL_CONFIG
+#define MLIBC_GLOBAL_CONFIG
+
+namespace mlibc {
+	struct GlobalConfig {
+		bool secureRequired;
+	};
```
ArsenArsen, line 7: we don't indent in namespaces

options/rtdl/generic/main.cpp
```diff
@@ -28,7 +29,8 @@ extern HIDDEN void *_GLOBAL_OFFSET_TABLE_[];
 extern HIDDEN Elf64_Dyn _DYNAMIC[];
 #endif
 
-bool secureRequired;
+mlibc::GlobalConfig mlibcConfig;
```
ArsenArsen, line 32: might be nicer to use a meyers singleton

ArsenArsen (changes requested):

options/internal/include/mlibc/global-config.hpp
```diff
@@ -0,0 +1,21 @@
...
+
+struct GlobalConfig {
+	GlobalConfig();
+	
+	bool debugMalloc;
+};
+
+inline const GlobalConfig &globalConfig() {
+	static GlobalConfig cached = []() -> GlobalConfig {
+		return GlobalConfig();
```
ArsenArsen, line 14: This isn't necessary (and neither is the comment or the extranious repetition of types, a mere `static GlobalConfig cached{};` would've made this [load on invocation](https://en.cppreference.com/w/cpp/language/storage_duration#Static_local_variables))

options/internal/generic/global-config.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+GlobalConfigGuard::GlobalConfigGuard() {
+	// Force the config to be created during initialization of libc.so.
+	mlibc::globalConfig();
+}
+
+static bool envEnabled(const char *env) {
+	auto value = getenv(env);
+	if (value)
+		return !strcmp(value, "1");
```
ArsenArsen, line 21: It might be nicer to check for any nonzero, nonnull value (so, `return value && *value;`)

options/internal/include/mlibc/global-config.hpp
```diff
@@ -0,0 +1,21 @@
...
+
+namespace mlibc {
+
+struct GlobalConfig {
+	GlobalConfig();
+	
+	bool debugMalloc;
+};
+
+inline const GlobalConfig &globalConfig() {
```
ArsenArsen, line 12: I don't think that will work? does the static member get deduplicated despite being in each TU?

ArsenArsen (changes requested):

options/internal/generic/global-config.cpp
```diff
@@ -0,0 +1,30 @@
...
+};
+
+GlobalConfigGuard guard;
+
+GlobalConfigGuard::GlobalConfigGuard() {
+	// Force the config to be created during initialization of libc.so.
+	mlibc::globalConfig();
+}
+
+static bool envEnabled(const char *env) {
```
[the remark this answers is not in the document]
64: where?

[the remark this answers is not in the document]
ArsenArsen: gah, github unified diff strikes again. disregard.

options/internal/include/mlibc/global-config.hpp
64: Yeah, it works fine. The linker will just pick one:
```
$ nm --demangle libc.so | rg cached
0000000000530390 u guard variable for mlibc::globalConfig()::cached
000000000053038d u mlibc::globalConfig()::cached
```

Geertiebear (approved):
I think this PR look good as it is now, and it seems as though all of @avdgrinten's comments have been covered.

PR #652, opened 2022-07-08
abis: add statfs.h and statvfs.h, pull suseconds_t and socklen_t into abi-bits too

64:


abis/linux/fsfilcnt_t.h
```diff
@@ -0,0 +1,6 @@
+#ifndef _ABIBITS_FSFILCNT_T_H
+#define _ABIBITS_FSFILCNT_T_H
+
+typedef __mlibc_uint64 fsfilcnt_t;
+
+#endif // _ABIBITS_FSFILCNT_T_H
```
ArsenArsen, line 6: `abi-bits` are also public headers (ditto in the other dozen or so cases of line comments)

ArsenArsen (changes requested):
lgtm

abis/mlibc/socket.h
```diff
@@ -1,7 +1,7 @@
 #ifndef _ABIBITS_SOCKET_H
 #define _ABIBITS_SOCKET_H
 
-#include <bits/posix/socklen_t.h>
+#include <abi-bits/socklen_t.h>
```
no92, line 4: This change is, as far as I can tell, not being made in `abis/linux/socket.h`, which seems wrong.

abis/linux/socket.h
```diff
@@ -18,7 +18,7 @@ struct msghdr {
 	size_t msg_iovlen;
 	void *msg_control;
 	size_t msg_controllen;
-	unsigned int msg_flags;
```
ArsenArsen, line 20: unsigned bitfields?

no92 (approved):

Geertiebear (approved):
LGTM.

PR #655, opened 2022-07-08
Additions and fixes for LFS, part 3

no92:


64 (approved):

options/posix/generic/posix_stdlib.cpp
```diff
@@ -217,6 +217,13 @@ char *realpath(const char *path, char *out) {
 
 	// If the path is relative, we have to preprend the working directory.
 	if(path[0] == '/') {
+		if(strlen(path) == 1) {
+			if(!out)
+				out = reinterpret_cast<char *>(getAllocator().allocate(2));
+			strcpy(out, path);
+			return out;
+		}
```
avdgrinten, line 225: I don't think that this is correct. This is supposed to handle `/` inputs, right? However, `//` is equivalent to `/` and that is not handled.

options/glibc/include/sys/ioctl.h
```diff
@@ -25,6 +25,9 @@ int ioctl(int fd, unsigned long request, ...);
 #define SIOCSIFFLAGS 0x8914
 #define SIOCGIFINDEX 0x8933
 
+#define SIOCDEVPRIVATE 0x89F0
+#define SIOCPROTOPRIVATE 0x89E0
```
Dennisbonke, line 29: We sort these by value from what I can tell. Swap this one around with the other one to match that.

options/linux/generic/sys-klog.cpp
```diff
@@ -0,0 +1,16 @@
+#include <errno.h>
+#include <sys/klog.h>
+
+#include <bits/ensure.h>
+
+#include "mlibc/linux-sysdeps.hpp"
```
Dennisbonke, line 6: Why the use of `”` here? We should be able to use `<>` here if I’m not mistaken.

options/linux/include/sys/klog.h
```diff
@@ -0,0 +1,14 @@
...
+extern "C" {
+#endif
+
+int klogctl(int type, char *bufp, int len);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif // _SYS_KLOG_H
```
Dennisbonke, line 14: We shouldn’t use C++ style comments in public headers.

Dennisbonke (changes requested):
Looks good, a few minor nitpicks.

options/linux/generic/sys-klog.cpp
```diff
@@ -0,0 +1,16 @@
...
+
+int klogctl(int type, char *bufp, int len) {
+	MLIBC_CHECK_OR_ENOSYS(mlibc::sys_syslog, -1);
+	int out;
+	if (int e = mlibc::sys_syslog(type, bufp, len, &out); e) {
```
ArsenArsen, line 11: The name of this syscall is an unfortunate thing on Linux. We shouldn't preserve that, and we should just call this sys_klogctl instead.

ArsenArsen (changes requested):

64 (approved):

ArsenArsen (approved):

no92:
Squashed the fixups.

Geertiebear (approved):

PR #657, opened 2022-07-09
options/ansi: fix mbtowc and mbrtowc on null byte

64:


options/ansi/generic/wchar-stubs.cpp
```diff
@@ -83,10 +82,13 @@ size_t mbrtowc(wchar_t *wcp, const char *mbs, size_t mb_limit, mbstate_t *stp) {
 		errno = EILSEQ;
 		return static_cast<size_t>(-1);
 	}else{
-		size_t n = nseq.it - mbs;
-		if(!n) // Null-terminate resulting wide string.
+		if (*mbs) {
+			return nseq.it - mbs;
+		} else {
+			// TODO: Reset shift state.
```
Geertiebear, line 88: Why don't we just do that?

Geertiebear (approved):
LGTM.

PR #662, opened 2022-07-09
options/ansi: add __MLIBC_TRACE macro to aid debugging

64:
Since we find ourselves printf debugging ports a lot (especially when gdb can't be used), this macro will be nice to have.

I *think* this should work under all `-std=` values, since we're in a system header.

### DIFF
--- a/options/ansi/include/stdio.h
+++ b/options/ansi/include/stdio.h
@@ -226,5 +226,12 @@ int fputs_unlocked(const char *, FILE *);
 #	include <bits/posix/posix_stdio.h>
 #endif
 
+/* Macro to aid printf debugging. See tests/ansi/trace.c */
+#define __MLIBC_TRACE(format, ...) ({ \
+			fprintf(stderr, "__MLIBC_TRACE(): %s at " __FILE__ ":%d " format "\n", \
+					__PRETTY_FUNCTION__, __LINE__, ##__VA_ARGS__); \
+			fflush(stderr); \
+		})
+
 #endif // _STDIO_H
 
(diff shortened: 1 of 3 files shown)


ArsenArsen:
> I _think_ this should work under all `-std=` values, since we're in a system header.

not at all, macros aren't impacted by `-std`, and macro expansions are checked at use site (so `-std=c99 -pedantic` would still break on this)

I don't think this belongs in libc, you can easily do something like this:

```c
[i] ~$ cat ~/.config/nvim/minisnip/debugging_trace
#define PORTERS_TRACE(format, ...) ({ \
		fprintf(stderr, "PORTERS TRACER: %s at " __FILE__ ":%d " format "\n", \
				__PRETTY_FUNCTION__, __LINE__, ##__VA_ARGS__); \
		fflush(stderr); \
	})
```

alternatively, you could just dump a header file into `/usr/local/include`, as it comes first in the include path:

```c
[i] ~$ cat /usr/local/include/stdio.h
#include_next <stdio.h>

#ifndef PORTERS_TRACE
#define PORTERS_TRACE(format, ...) ({ \
		fprintf(stderr, "PORTERS TRACER: %s at " __FILE__ ":%d " format "\n", \
				__PRETTY_FUNCTION__, __LINE__, ##__VA_ARGS__); \
		fflush(stderr); \
	})
/* untested */
#endif /* PORTERS_TRACE */
[i] ~$ 
```

avdgrinten:
I also dislike the idea of adding mlibc-specific stuff to the headers. There are too many C extensions already, we should not add our own ones.

PR #665, opened 2022-07-10
abis: move clockid_t.h to abis/ and fix Linux ABI

64:


64:
Blocked on `abidiff` crash.

ArsenArsen:
ABI is almost definitely different; the crash can be ignored for now (I did pass on relevant information upstream)

64:
Yeah, the ABI *should* be different for Linux (clockid_t was wrong before). But since the crash seems to happen when reporting said difference, we can probably just merge this right?

ArsenArsen:
I don't think abidiff is our sole source of truth on whether something is an abi-break ;)

ArsenArsen:
Ah, yeah

PR #666, opened 2022-07-10
options/posix: handle spurious wakeups in pthread_cond_timedwait

64:


options/posix/generic/pthread-stubs.cpp
```diff
@@ -928,13 +928,61 @@ int pthread_cond_wait(pthread_cond_t *__restrict cond, pthread_mutex_t *__restri
...
+			struct timespec timeout;
+			timeout.tv_sec = abstime->tv_sec - now.tv_sec;
+			timeout.tv_nsec = abstime->tv_nsec - now.tv_nsec;
+
+			// Check if abstime has already passed.
+			if (timeout.tv_sec < 0 || (timeout.tv_sec == 0 && timeout.tv_nsec < 0)) {
+				if (pthread_mutex_lock(mutex))
+					__ensure(!"Failed to lock the mutex");
+				return ETIMEDOUT;
+			}
```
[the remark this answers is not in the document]
64: No, though the man page isn't totally clear about it. If I lock after ETIMEDOUT return, glibc hangs (and musl also seems to always lock before returning too, so I'm pretty sure that this is correct).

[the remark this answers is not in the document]
64: nvm, the man page is very clear about it:
> When such timeouts occur, pthread_cond_timedwait() shall nonetheless release and re-acquire the mutex referenced by mutex.

tests/posix/pthread_cond.c
```diff
@@ -1,48 +1,72 @@
...
-	pthread_create(&t2, NULL, &waiting_thread, &arg);
+	pthread_create(&t1, NULL, &worker, NULL);
+	pthread_create(&t2, NULL, &worker, NULL);
 
-	struct timespec wait_time = { .tv_sec = 0, .tv_nsec = 150000000 }; // 150ms
-	nanosleep(&wait_time, NULL);
+	// Wait until the workers have actually entered the cond_wait
+	// before doing a broadcast.
+	while (waiting != 2 || pthread_mutex_trylock(&mtx) == EBUSY)
+		usleep(150000); // 100ms
```
Geertiebear, line 28: Typo in the comment?

options/posix/generic/pthread-stubs.cpp
```diff
@@ -927,15 +927,77 @@ int pthread_cond_wait(pthread_cond_t *__restrict cond, pthread_mutex_t *__restri
...
+			timeout.tv_sec = abstime->tv_sec - now.tv_sec;
+			timeout.tv_nsec = abstime->tv_nsec - now.tv_nsec;
+
+			// Check if abstime has already passed.
+			if (timeout.tv_sec < 0 || (timeout.tv_sec == 0 && timeout.tv_nsec < 0)) {
+				if (pthread_mutex_lock(mutex))
+					__ensure(!"Failed to lock the mutex");
+				return ETIMEDOUT;
+			} else if (timeout.tv_nsec >= nanos_per_second) {
+				timeout.tv_nsec -= nanos_per_second;
```
Geertiebear, line 961: Couldn't ``tv_nsec`` be multiple times larger than ``nanos_per_second``?

tests/posix/pthread_cond.c
```diff
@@ -1,48 +1,72 @@
...
+	int e = pthread_cond_timedwait(&cond, &mtx, &before_now);
+	assert(e == ETIMEDOUT);
+	pthread_mutex_unlock(&mtx);
+
+	long nanos_per_second = 1000000000;
+	struct timespec after_now;
+	assert(!clock_gettime(CLOCK_REALTIME, &after_now));
+	after_now.tv_nsec += nanos_per_second / 10; // 100ms
+	if (after_now.tv_nsec >= nanos_per_second) {
+		after_now.tv_nsec -= nanos_per_second;
```
Geertiebear, line 53: Here too, couldn't ``tv_nsec`` be multiple times larger than ``nanos_per_second``?

options/posix/generic/pthread-stubs.cpp
64: I don't think so, since both abstime.tv_nsec and now.tv_nsec should be less than nanos_per_sec, but I'll add an ensure here why not.

Geertiebear: That makes sense, it would be good to add an ensure in the function at least.

qookei: POSIX says that tv_nsec is in range [0, 1000 million) (and any set/sleep function will fail with EINVAL if it's not for example).

tests/posix/pthread_cond.c
64: Here I think definitely no, assuming clock_gettime works.

Geertiebear (approved):
LGTM.

ArsenArsen (approved):

PR #667, opened 2022-07-11
ci: add fixup checker

ArsenArsen:


avdgrinten:
It would be good to push this to managarm-infrastructure or so.

PR #669, opened 2022-07-12
rtdl: minor fixes

64:
In preparation for #448.


options/rtdl/generic/linker.cpp
```diff
@@ -1107,31 +1107,40 @@ frg::optional<ObjectSymbol> Scope::resolveSymbol(ObjectSymbol r, ResolveFlags fl
...
-	object->linkMap.base = object->baseAddress;
-	object->linkMap.name = object->path.data();
-	object->linkMap.dynv = object->dynamic;
+	__ensure(_linkBfs.size() == 0);
 
-	_linkSet.insert(object, Token{});
+	struct Token {};
+	using Set = frg::hash_map<SharedObject *, Token,
+			frg::hash<SharedObject *>, MemoryAllocator>;
+	Set set{frg::hash<SharedObject *>{}, getAllocator()};
```
avdgrinten, line 1102: Why is this set moved out of the `Loader` object? This adds the (otherwise unnecessary?) restriction that `submitObject` must only be called once.

options/rtdl/generic/linker.cpp
```diff
@@ -1085,13 +1070,13 @@ void Scope::appendObject(SharedObject *object) {
...
-frg::optional<ObjectSymbol> Scope::resolveSymbol(ObjectSymbol r, ResolveFlags flags) {
+frg::optional<ObjectSymbol> Scope::resolveSymbol(frg::string_view string, ResolveFlags flags) {
 	for(size_t i = 0; i < _objects.size(); i++) {
 		if((flags & resolveCopy) && _objects[i]->isMainObject)
 			continue;
 
-		const char *string = (const char *)(r.object()->baseAddress
-				+ r.object()->stringTableOffset + r.symbol()->st_name);
+		// const char *string = (const char *)(r.object()->baseAddress
+		// 		+ r.object()->stringTableOffset + r.symbol()->st_name);
```
avdgrinten, line 1079: ?

64: oops

options/rtdl/generic/linker.cpp
avdgrinten: I guess that would also work.

64: Ok, done.

avdgrinten (approved):

PR #670, opened 2022-07-12
rtdl: pass O_RDONLY to sys_open

cleanbaja:


ArsenArsen:
this won't work with ansi-only builds (they don't have fcntl). I think this is why it was just set to zero before. you'll need to use `__MLIBC_O_RDONLY` from abi.h

64 (approved):

64:
Thanks :)

PR #673, opened 2022-07-14
sysdeps/managarm: hack sys_gettid

64:
This should unbreak the CI.

### DIFF
--- a/sysdeps/managarm/generic/fork-exec.cpp
+++ b/sysdeps/managarm/generic/fork-exec.cpp
@@ -423,6 +423,11 @@ int sys_seteuid(uid_t euid) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


ArsenArsen:
is this really a hack? also, if this function is required, we should probably make it strong

64:
Well, it's a hack because each thread should have a unique TID but the same PID, right?

ArsenArsen:
IIRC, getpid() returns TID currently on managarm, which is what I mean. I could be misremembering, though.

64:
Even if it does, that should be fixed in managarm + a new GET_TID request type should be added for use in `gettid`. In any case, the code here should be changed later.

Geertiebear (approved):

PR #682, opened 2022-07-19
options/internal: add a more strict debug allocator

qookei:
This allocator uses mmap for every allocation, performs explicit checks
on the given arguments, adds a guard page after every allocation (while
placing the allocation as close to it as possible), and, if configured
to do so, will never release virtual addresses to catch a certain
category of use-after-free bugs.

### DIFF
--- a/options/internal/generic/allocator.cpp
+++ b/options/internal/generic/allocator.cpp
@@ -5,6 +5,9 @@
 #include <frg/eternal.hpp>
 #include <mlibc/allocator.hpp>
 #include <mlibc/internal-sysdeps.hpp>
+#include <internal-config.h>
+
+#ifndef MLIBC_DEBUG_ALLOCATOR
 
 // --------------------------------------------------------
 // Globals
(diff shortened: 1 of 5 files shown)


options/internal/generic/allocator.cpp
```diff
@@ -33,3 +36,130 @@ void VirtualAllocator::unmap(uintptr_t address, size_t length) {
...
 
+#else
+
+struct AllocatorMeta {
+	uint64_t allocatedSize;
+	uint64_t pagesSize;
+	frg::array<uint64_t, 4> magic;
+};
+
+constexpr frg::array<uint64_t, 4> __mlibc_allocatorMagic {
```
64, line 47: This doesn't really need to have a double underscore, does it?

options/internal/generic/allocator.cpp
```diff
@@ -33,3 +36,130 @@ void VirtualAllocator::unmap(uintptr_t address, size_t length) {
 	__ensure(!mlibc::sys_anon_free((void *)address, length));
 }
 
+#else
+
+struct AllocatorMeta {
+	uint64_t allocatedSize;
+	uint64_t pagesSize;
```
64, line 43: These should probably be `size_t`

64 (changes requested):
LGTM otherwise

options/internal/generic/allocator.cpp
qookei: Oh yeah this can just go into an anonymous namespace.

64 (approved):

Geertiebear (approved):
LGTM, just remove the fixup commits.

no92 (approved):

PR #683, opened 2022-07-19
rework syscall definitions in linux

ArsenArsen:
potential todo: add a CI pipeline to update this monthly


sysdeps/linux/include-internal/linux/unistd.h
```diff
@@ -0,0 +1 @@
+#include <syscallnos.h>
```
64, line 1: Can't this live in `include/bits`?

sysdeps/linux/update-syscall-list.py
```diff
@@ -0,0 +1,321 @@
...
+logging.basicConfig(level=logging.DEBUG)
+LOGGER = logging.getLogger(__name__)
+
+# list of architecture directories to populate with syscall numbers
+WANTED_ARCHES = ["riscv64", "x86_64"]
+
+argp = argparse.ArgumentParser()
+argp.add_argument("kver", help="Kernel version to download", default="latest")
+
+GENERATE_ARCHES = ["x86_64", "riscv64"]
```
64, line 29: This appears to be unused

64 (changes requested):
Might have some more comment later but the general approach LGTM. What I'd say is that it'd be nice to sort the generated files somehow (either alphabetically or by syscall number) so that they're easier to search + order won't randomly change when the headers are pulled again

sysdeps/linux/include-internal/linux/unistd.h
ArsenArsen: it's not installed

64: Can't we just not install it then? bc having both `include/bits` and `include-internal` seems a bit unnecessary to me

64 (approved):
LGTM, wouldn't mind if the header was moved still but I'm not too fussed about it

Geertiebear (approved):
LGTM, does the script need to be run by users ever or is it only used for development?

ArsenArsen:
not even devs would need to run it normally

PR #684, opened 2022-07-20
Port Linux sysdeps to AArch64

### LINKED ISSUE #570: aarch64: tpidr_el0 should point to 16 bytes before the first tls block
Found by @qookei. This may cause bugs if the linker replaces thread local accesses by direct offsets from the thread pointer.

qookei:
Fixes #570


sysdeps/linux/aarch64/thread_entry.S
```diff
@@ -0,0 +1,25 @@
+.section .text
+.global __mlibc_spawn_thread
+.type __mlibc_spawn_thread, "function"
+__mlibc_spawn_thread:
+	// __mlibc_spawn_thread(flags, stack, pid_out, child_tid, tls)
```
Geertiebear, line 5: Why are these commented out? Why can't we just remove these lines?

Geertiebear (changes requested):
And remove the fixup commits too please.

sysdeps/linux/aarch64/thread_entry.S
qookei: These comments are explaining what we're doing with the argument registers before the clone syscall (similar to the riscv sysdeps).

Geertiebear (approved):

64 (approved):
LGTM modulo the libgcc discussion.

ArsenArsen (approved):
WRT toolchains, we can take a hit even building them, due to the GHA cache (i.e. we could write caches into CI jobs).
It also might be nice to eventually use self-hosted runners for this, which should save us significant time.

64 (approved):

64 (approved):

Geertiebear (approved):

PR #689, opened 2022-07-20
options/posix: Let AF_INET6 in inet_ntop fall through

Dennisbonke:
This makes more sense than hard crashing, we do this in other places too.


ArsenArsen (approved):
not the most descript commit message, but this change is an improvement

Geertiebear (approved):
LGTM.

PR #690, opened 2022-07-20
options/posix: Implement the if_ family

Dennisbonke:
One stub, 3 properly implemented, I forgot what needed it exactly.

### DIFF
--- a/options/posix/generic/net-if-stubs.cpp
+++ b/options/posix/generic/net-if-stubs.cpp
@@ -1,12 +1,15 @@
 #include <errno.h>
+#include <string.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <sys/ioctl.h>
 #include <net/if.h>
 
 #include <bits/ensure.h>
 #include <mlibc/posix-sysdeps.hpp>
 
-void if_freenameindex(struct if_nameindex *) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+void if_freenameindex(struct if_nameindex *idx) {
+	free(idx);
 }
 
 char *if_indextoname(unsigned int index, char *name) {


options/posix/generic/net-if-stubs.cpp
```diff
@@ -1,24 +1,49 @@
...
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+unsigned int if_nametoindex(const char *name) {
+	struct ifreq ifr;
+	int fd, r;
+
+	if((fd = socket(AF_UNIX, SOCK_DGRAM|SOCK_CLOEXEC, 0)) < 0) {
+		return 0;
+	}
+	strncpy(ifr.ifr_name, name, sizeof ifr.ifr_name);
```
ArsenArsen, line 45: function style sizeof

options/posix/generic/net-if-stubs.cpp
```diff
@@ -1,24 +1,49 @@
...
+		}
+		return 0;
+	}
+	return strncpy(name, ifr.ifr_name, IF_NAMESIZE);
 }
 
 struct if_nameindex *if_nameindex(void) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+	errno = ENOSYS;
```
Geertiebear, line 34: We should log an info message here that it is stubbed.

Geertiebear (changes requested):

options/posix/generic/net-if-stubs.cpp
64: Are we sure that the strncpy call behaves correctly here? e.g if `name` is too long, then `ifr.ifr_name` won't be null terminated.

options/posix/generic/net-if-stubs.cpp
```diff
@@ -10,9 +10,24 @@ void if_freenameindex(struct if_nameindex *idx) {
...
+	ifr.ifr_ifindex = index;
+	r = ioctl(fd, SIOCGIFNAME, &ifr);
+	close(fd);
+	if (r < 0) {
+		if(errno == ENODEV) {
+			errno = ENXIO;
+		}
+		return 0;
+	}
+	return strncpy(name, ifr.ifr_name, IF_NAMESIZE);
```
ArsenArsen, line 30: I think this should be a sysdep, this is the Linux implementation, others might not just pick a random socket to run an ioctl on to resolve interface names.

options/posix/generic/net-if-stubs.cpp
```diff
@@ -31,8 +31,9 @@ char *if_indextoname(unsigned int index, char *name) {
 }
 
 struct if_nameindex *if_nameindex(void) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+	mlibc::infoLogger() << "mlibc: if_nametoindex is a stub!" << frg::endlog;
```
ArsenArsen, line 34: like the other one, it might be nice to make this into a sysdep stub

options/posix/generic/net-if-stubs.cpp
```diff
@@ -36,7 +36,15 @@ struct if_nameindex *if_nameindex(void) {
...
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+unsigned int if_nametoindex(const char *name) {
+	struct ifreq ifr;
+	int fd, r;
+
+	if((fd = socket(AF_UNIX, SOCK_DGRAM|SOCK_CLOEXEC, 0)) < 0) {
+		return 0;
+	}
+	strncpy(ifr.ifr_name, name, sizeof(ifr.ifr_name));
```
ArsenArsen, line 46: might be worth memsetting ifreq to 0 and using sizeof - 1 to ensure a null terminator exist, though, whether that matters is OS specific (and should be a sysdep)

no92 (approved):
Other than splitting out the implementations to identical sysdeps for linux and managarm, LGTM.

Dennisbonke:
Rebased and mostly superseded by #782. Kept `if_freeindex` as it was unimplemented, dropped the rest.

Dennisbonke:
After consideration with @no92, closing this PR is better, so closing.

PR #692, opened 2022-07-20
options/posix: Semi implement strtof_l

Dennisbonke:
Weston suddenly started using it, so ignore locale like we always do and call into `strtof()`


ArsenArsen (approved):
IIRC, we don't have locales at all, so this is good enough, though, I'd elaborate in the message more

Geertiebear (approved):

PR #694, opened 2022-07-20
options/posix: Make sendto and recvfrom use a sysdep on linux

Dennisbonke:
Linux has an actual sysdep for `sendto` and `recvfrom`. Use that instead of passing it through to `sendmsg` and `recvmsg`.


ArsenArsen (changes requested):
why are we doing this? sendto is a strict superset of sendmsg; ditto for recvmsg; also, `__linux__` is certainly not the way to test for this. if we're providing subset/superset sysdeps and fallthroughs, we need a concise and coherent system (for instance, picking out the first found sysdep, which could be done by a macro that takes varargs and searches for it, then returns it, or something)

Dennisbonke:
> why are we doing this? sendto is a strict superset of sendmsg; ditto for recvmsg; also, `__linux__` is certainly not the way to test for this. if we're providing subset/superset sysdeps and fallthroughs, we need a concise and coherent system (for instance, picking out the first found sysdep, which could be done by a macro that takes varargs and searches for it, then returns it, or something)

Well I was trying just about everything to fix some bugs, it could well be that this change is indeed unneeded. Afaik nothing hard depends on it so we can drop it if that is preferred.

sysdeps/linux/generic/sysdeps.cpp
```diff
@@ -616,6 +616,22 @@ int sys_listen(int fd, int backlog) {
 	return 0;
 }
 
+ssize_t sys_sendto(int fd, const void *buffer, size_t size, int flags, const struct sockaddr *sock_addr, socklen_t addr_length, ssize_t *length) {
+	auto ret = do_cp_syscall(SYS_sendto, fd, buffer, size, flags, sock_addr, addr_length);
+        if (int e = sc_error(ret); e)
```
64, line 621: indentation is wrong

options/posix/generic/sys-socket-stubs.cpp
```diff
@@ -139,6 +149,7 @@ ssize_t send(int fd, const void *buffer, size_t size, int flags) {
 
 ssize_t sendto(int fd, const void *buffer, size_t size, int flags,
 		const struct sockaddr *sock_addr, socklen_t addr_length) {
+#ifndef __linux__
```
64, line 152: I agree that it'd be better to use sendmsg as a fallback instead of this define

64 (changes requested):

ArsenArsen:
what bug was that? and we don't have an established precedent here, so if we're going down the route of having redundant sysdeps, we should figure out how

Dennisbonke:
> what bug was that? and we don't have an established precedent here, so if we're going down the route of having redundant sysdeps, we should figure out how

Weston hanging, which we for now presumably traced to a pthread issue. I'm more than fine with dropping this change because of that.

Geertiebear:
We've always favoured supersets of sysdeps over redundant ones, and since there was no bug fixed by this change I will close this. Of course if there is a bug involved with these sysdeps, we can reconsider these changes.

PR #697, opened 2022-07-21
options/posix: fix FutexLock deadlock

64:


sysdeps/linux/generic/sysdeps.cpp
```diff
@@ -1011,6 +1011,12 @@ void sys_exit(int status) {
 #define FUTEX_WAIT 0
 #define FUTEX_WAKE 1
 
+int sys_futex_tid() {
```
Geertiebear, line 1014: What is the difference between ``sys_futex_tid()`` and ``sys_gettid()``? Why do we use ``sys_futex_tid()`` in these changes and not ``sys_gettid()``?

64: They should be the same, we really ought to get rid of one of them. But using `sys_futex_tid` means I don't have to mess around with the sysdeps (e.g managarm) which implement `sys_futex_tid` for rtdl already.

64: Maybe we can make a separate PR to deduplicate these.

Geertiebear (approved):

PR #698, opened 2022-07-21
Implement set, get, end functions for passwd/group and tests

AtieP:


options/posix/generic/pwd-stubs.cpp
```diff
@@ -8,16 +8,9 @@
 #include <mlibc/debug.hpp>
 
 namespace {
-	thread_local passwd global_entry;
+	FILE *global_file; // Used by setpwent/getpwent/endpwent.
```
Geertiebear, line 11: It's probably nicer to have this as a static local variable in a wrapper function. Or at least have some ``get_global_file()`` static function that does the opening of the file because right now that code is duplicated 3x. Then also make a function like ``close_global_file()`` for when it needs to be closed.

options/posix/generic/grp-stubs.cpp
```diff
@@ -10,7 +10,7 @@
 #include <mlibc/posix-sysdeps.hpp>
 
 namespace {
-	thread_local group global_entry;
+	FILE *global_file;
```
Geertiebear, line 13: Same as the other comment.

Geertiebear (changes requested):

Geertiebear (approved):
LGTM now.

PR #701, opened 2022-07-21
options/posix: handle EAGAIN in sem_wait

64:


ArsenArsen (approved):
not sure if the error reporting should necessarily be a crash for misc errors, but this is an improvement

64:
> not sure if the error reporting should necessarily be a crash for misc errors, but this is an improvement

I think this is correct for EAGAIN, but EINTR I think we need to return and set errno?

Geertiebear (changes requested):
I think ``EINTR`` we do need to report, it would be good to do it in this PR too.

64:
FIxed this to handle EINTR properly.

Geertiebear (approved):

PR #705, opened 2022-07-23
remove reliance on __linux__, switch to using set10() for mlibc-config.h.in

ArsenArsen:


64 (approved):

Dennisbonke (approved):
With the note that a follow-up PR should be made to make mlibc at least warn during configure if the linux option is enabled, but the linux kernel headers can't be found and the linux-headers option is disabled, this PR looks good to me.

no92 (approved):

Geertiebear (approved):
LGTM.

PR #706, opened 2022-07-23
options/posix: fix race with cancelee in pthread_cancel

### LINKED ISSUE #613: posix: pthread_cancel test fails intermittently on RISC-V Linux
On commit 3459cb4a6f613d61ffe9e82dac0f0111e8555ed4, the CI encountered an intermittent failure in the `pthread_cancel` test on RISC-V (under the Linux sysdeps):
```
  61/144 mlibc:posix / pthread_cancel                 FAIL            1.10s   killed by signal 6 SIGABRT
――――――――――――――――――――――――――――――――――――― ✀  ―――――――――――――――――――――――――――――――――――――
stderr:
In function main, file ../../../src/mlibc/tests/posix/pthread_cancel.c:94: Assertion '!ret' failed!
```
I haven't been able to reproduce this anywhere, so it's possible that it is just a `qemu-user` bug or toolchain bug. It's also possible that the bug is present in the arch/OS-independent code too.

qookei:
It's possible that tgkill fails in pthread_cancel because the cancelee
has in the quit in the time between setting the cancellation trigger
bit, and sending SIGCANCEL.

Fixes #613

### DIFF
--- a/options/posix/generic/pthread-stubs.cpp
+++ b/options/posix/generic/pthread-stubs.cpp
@@ -495,12 +495,17 @@ void pthread_testcancel(void) {
 	}
 }
 int pthread_cancel(pthread_t thread) {
+	if (!mlibc::sys_tgkill) {
+		MLIBC_MISSING_SYSDEP();
+		return ENOSYS;
+	}
+
 	auto tcb = reinterpret_cast<Tcb *>(thread);
 	// Check if the TCB is valid, somewhat..
 	if (tcb->selfPointer != tcb)
 		return ESRCH;
 
-	int old_value = tcb->cancelBits;
+	int old_value = __atomic_load_n(&tcb->cancelBits, __ATOMIC_RELAXED);
 	while (1) {
 		int bitmask = tcbCancelTriggerBit;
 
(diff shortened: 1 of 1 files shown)


ArsenArsen (approved):

avdgrinten:
Doesn't wait()ing for cancelled threads prevent this issue?

qookei:
Even if we add a `wait()` after `tgkill()`, it's already too late in this scenario. The thread we are killing quit on it's own already, before we had a chance to reach `tgkill()`. Adding a `wait()` call could also introduce the following potential deadlock:
1. Thread 1 calls to cancel thread 2.
2. Thread 1 sets `tcbCancelTriggerBit` in thread 2's TCB.
3. Thread 2 quits.
4. Thread 3 spawns with same TID as thread 2.
5. Thread 1 `wait`s for thread 2 (which is already gone), and ends up waiting for thread 3, which may potentially never quit because thread 1 was expected to do something with it.

This is not a problem with the current mechanism (and the proposed fix), because:
1. Thread 1 calls to cancel thread 2.
2. Thread 1 sets `tcbCancelTriggerBit` in thread 2's TCB.
3. Thread 2 quits.
4. Thread 3 spawns with same TID as thread 2 (and a new TCB, with no cancel bits set).
5. Thread 1 `tgkill`s thread 2 (which is already gone), and ends up killing thread 3.
6. Thread 3 goes into the `SIGCANCEL` handler, sees that it (i) wasn't in a cancellation point, (ii) does not have `tcbAsyncCancelBit` set, or (iii) does not have `tcbCancelTriggerBit` set, and as such it ignores the signal.
7. Thread 1 returns out of `pthread_cancel`.

Geertiebear (approved):

PR #707, opened 2022-07-24
options/rtdl: implement `LD_SHOW_AUXV`

no92:


options/rtdl/generic/main.cpp
```diff
@@ -220,6 +236,71 @@ extern "C" void *interpreterMain(uintptr_t *entry_stack) {
 		if(!(*aux))
 			break;
 
+		if(ldShowAuxv) {
+			switch(*aux) {
+#ifdef AT_PHDR
```
64, line 241: These defines are really ugly, but I'm not sure there is a better way. We could perhaps try to unify the ABI headers a bit more.

64 (changes requested):

options/rtdl/generic/main.cpp
ArsenArsen: could also replace AT_* with something like

```c
#define AT_DO_DEFINES \
        at_cb(AT_PHDR, 1) \
        at_cb(AT_PHENT, 2) \
        ...

enum {
#define at_cb(name, val) name = val,
AT_DO_DEFINES
#undef at_cb
};
```

this block would become

```c
#define at_cb(name, val) case name: mlibc::infoLogger() << #name " 0x" ...
AT_DO_DEFINES
#undef at_cb
```

no92: As far as I can tell, possible `a_type` values are defined in the arch ABI spec, so I don't quite see the point of having it be an abi-bit as it is right now.

avdgrinten: Aren't some of the AT_ constants mandatory for ld.so to work? These can be assumed to exist on all ABIs. Other `AT_` constants are Linux extensions and e.g. FreeBSD's `AT_` constants do not match Linux'.

avdgrinten: Also, the constants defined by the ELF ABI supplement can be assumed to exist and have values matching the standard.

64 (approved):

Dennisbonke (approved):
LGTM

Geertiebear (approved):

PR #708, opened 2022-07-24
pastoral sysdeps

ethan4984:


64:
Thanks for the PR. Can you add the pastoral sysdeps to be run on CI? See `.github/ci.yml`.

ArsenArsen:
can you drop the merge commits and reword the commits to align with the rest of the repository?

avdgrinten:
Use of the mlibc ABI is discouraged for new ports. Consider using the Linux ABI instead (but note that this may require changes in your kernel since the definitions of structs / constants need to match).

PR #710, opened 2022-07-28
Fix some CPU_ stuff on linux

Dennisbonke:
Also implements CPU_ZERO for both mlibc and linux.


avdgrinten (commented):
Why is this an ABI break?

This code does not match mlibc coding style *at all*. It is also not documented. Why are the double underscore macros needed? Why are they not prefixed with __MLIBC? Was it copied from somewhere (copyright violation)? Instead of using these strange macros, we should use proper inline functions.

avdgrinten:
The `#ifdef` based on `__MLIBC_LINUX_OPTION` doesn't make sense to me either. Why does this type change based on the enabled options?

Dennisbonke:
Closing as 828 merged a better version of this.

PR #712, opened 2022-07-28
abis: Move __sighandler to abis

Dennisbonke:

### DIFF
--- a/abis/linux/signal.h
+++ b/abis/linux/signal.h
@@ -94,6 +94,9 @@ typedef struct {
 extern "C" {
 #endif
 
+// Argument for signal()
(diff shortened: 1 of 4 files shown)


avdgrinten (commented):
Why is this done? AFAICS no ABI deviates from the current definition anyway.

Dennisbonke:
I tried to remember that, but unfortunately couldn’t figure out why I made this change. It could well be that we deem this unneeded, but I just wanted to try to PR some small stuff before August first and this one was small and easy to do.

Dennisbonke:
I don't remember why we needed this in the first place, and as such, I'm closing it for now.

PR #719, opened 2022-08-05
internal: use libsmarter

64:
Blocked on managarm/libsmarter#4.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -53,11 +53,19 @@ if not headers_only
 	libc_deps += frigg_dep
 	rtdl_deps += frigg_dep
 
+	libsmarter_dep = dependency(
+		'libsmarter',
+		default_options: [],
+		fallback: ['libsmarter', 'libsmarter_dep'],
+	)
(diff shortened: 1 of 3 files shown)


no92 (approved):
LGTM

avdgrinten (commented):
Why are the changes in cxxabi.cpp needed? libsmarter should not call these operators (for example, they do not exist in thor).

64:
libsmarter's make_shared at least seems to call `operator new`.

64:
Ah, okay. `allocate_shared` should do the same thing as `make_shared` but take an explicit allocator. We can just use that, then.

64:
IIRC I wanted to use it for refcounting stuff in ld.so so that we don't e.g leak memory when we destroy threads. This PR needs a bit of updating as per my last comment, though

Geertiebear:
Alright, but then at leas the use-case is clear, thanks.

no92:
Added with #1480.

PR #720, opened 2022-08-05
abis/linux: correct the ABI of some integer types

64:
Verified against musl, see https://github.com/bminor/musl/blob/46d1c7801bb509e1097e8fadbaf359367fa4ef0b/include/alltypes.h.in#L27-L35.


Dennisbonke (approved):
LGTM, might want to backport this into 3.0 too @Geertiebear? Not majorly important but it would be nice considering we're recommending the Linux ABI now.

64:
ABI checker gives a false positive: `ino_t` has been changed from signed to unsigned.

no92 (approved):

PR #721, opened 2022-08-05
options/glibc: simplify basename ABI and implementation

64:
This will need to be backported to the release branch, since it's an ABI break.

### DIFF
--- a/options/ansi/include/string.h
+++ b/options/ansi/include/string.h
@@ -62,27 +62,10 @@ void *memmem(const void *, size_t, const void *, size_t);
 
 /* Handling the basename mess:
  * If <libgen.h> is included *at all*, we use the XPG-defined basename
- * implementation, otherwise, we use the GNU one. Since our ABI previously
- * provided the XPG one under basename, we'll have to diverge from GNU here and
- * provide __mlibc_gnu_basename instead.
+ * implementation, otherwise, we use the GNU one.
  */
 #if defined(__MLIBC_GLIBC_OPTION) && defined(_GNU_SOURCE) && !defined(basename)
-char *__mlibc_gnu_basename_c(const char *path);
-
-# ifdef __cplusplus
-extern "C++" {
-static inline const char *__mlibc_gnu_basename(const char *path) {
-	return __mlibc_gnu_basename_c(path);
-}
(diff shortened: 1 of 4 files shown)


64:
ABI check fails, but that's deliberate.

options/posix/include/libgen.h
```diff
@@ -6,13 +6,8 @@
...
 
-#if defined(basename) && defined(_GNU_SOURCE)
-/* see: ./options/ansi/include/string.h, search for __mlibc_gnu_basename */
-# undef basename
-#endif
-
-char *basename(char *);
-#define basename basename
+char *__mlibc_xpg_basename(char *);
+#define basename __mlibc_xpg_basename
```
ArsenArsen, line 10: this will generate a redefinition warning

options/posix/generic/libgen-stubs.cpp
```diff
@@ -5,7 +5,9 @@
 
 #include <mlibc/debug.hpp>
 
-// Adopted from musl's code.
+// The functions in this file are adapted from musl.
+
+// Note that 'basename' here is defined to be __mlibc_xpg_basename by libgen.h.
```
ArsenArsen, line 10: I'm not sure this is sane

options/ansi/include/string.h
```diff
@@ -62,27 +62,10 @@ void *memmem(const void *, size_t, const void *, size_t);
...
-static inline char *__mlibc_gnu_basename(char *path) {
-	return __mlibc_gnu_basename_c(path);
-}
-}
-# else
-#  define __mlibc_gnu_basename __mlibc_gnu_basename_c
-# endif
-
-#define basename __mlibc_gnu_basename
+char *basename(const char *path);
```
ArsenArsen, line 68: again, this was deliberate (it prevents the `char * x = basename((const char *)"abc");` case.

ArsenArsen (changes requested):
I wouldn't mind removing the overloads if we decide to do that to be consistent with other string functions (though, it can change semantics of c++ programs, which is something to consider before finalizing a decision), but swapping the `__mlibc_gnu_basename_c` and `basename` symbols is a very unnecessary ABI break

PR #723, opened 2022-08-06
options/posix: include sys/select.h in sys/types.h

64:


Dennisbonke (approved):
LGTM

avdgrinten:
Why is this done? That doesn't seem right to me? If we need types from sys/select.h in sys/types.h, we should put them into a `bits/` header instead.

Dennisbonke:
> Why is this done? That doesn't seem right to me? If we need types from sys/select.h in sys/types.h, we should put them into a `bits/` header instead.

Yes, I’d agree, but since glibc includes the entire thing, and I believe musl does too, if the correct feature macro is defined (which, by glibc defaults, it is iirc), then just moving the FD_SET family of macros to a bits header doesn’t solve the entire problem. Applications assume that including sys/types.h exposes pselect and select too. I don’t like it either but this seems to be the glibc way. For context, it seems that 4.3BSD included sys/select.h in sys/types.h, which glibc will do if __USE_MISC is defined as true, which is the case if _DEFAULT_SOURCE is defined as true.

PR #725, opened 2022-08-08
options/posix: move poll(3) from linux option to posix

no92:


qookei (commented):
LGTM, but perhaps it'd be good to remove the `-stubs` infix from `poll-stubs.cpp`, since it doesn't contain any stubs

qookei (approved):

avdgrinten:
poll() is not in POSIX.

qookei:
```
CONFORMING TO
   poll() conforms to POSIX.1-2001 and POSIX.1-2008.       ppoll() is Linux-specific.
```
https://pubs.opengroup.org/onlinepubs/9699919799/functions/poll.html

avdgrinten:
Ah, I was looking for the wrong header.

By the way, shouldn't the poll.h header be moved too?

qookei:
Oh true, that header should be moved as well

no92:
The `poll.h` header is already part of the posix options.

PR #726, opened 2022-08-17
options/ansi: Add basic support for complex arithmetic

nikstrmp:


64:
Thanks for the PR, as mentioned on Discord some testing would be nice to have.

nikstrmp:
> Thanks for the PR, as mentioned on Discord some testing would be nice to have.

Wrote some basic tests, if someone could review them, it would be nice.

nikstrmp:
I'm so sorry, I forgot to push a change that I had made. Thank you for you understanding.

nikstrmp:
I think this is ready for review (for real this time)

avdgrinten:
LGTM but the fix-up commits should be squashed.

options/ansi/generic/complex-stubs.cpp
```diff
@@ -0,0 +1,30 @@
+#include <complex.h>
+
+extern "C" {
```
no92, line 3: This extern seems unnecessary.

options/ansi/meson.build
```diff
@@ -6,6 +6,7 @@ endif
 ansi_sources = files(
 	'generic/stdlib-stubs.cpp',
 	'generic/assert-stubs.cpp',
+	'generic/complex-stubs.cpp',
```
no92, line 9: The `-stubs` suffix should be dropped IMO.

PR #730, opened 2022-08-29
options/posix: Make strtod_l call strtod

Dennisbonke:
The new glib update seems to like strtod_l, so let's make it a call to strtod as to not break 75% of Managarm userspace.


Dennisbonke:
As the CI is about to run to completion and this change should be highly uncontroversial and has been open for over one month, I'm merging it.

PR #734, opened 2022-09-06
utmp implementations from LFS

no92:


options/linux/generic/utmp-stubs.cpp
```diff
@@ -1,27 +1,116 @@
...
+#include <errno.h>
+#include <fcntl.h>
+#include <stdlib.h>
+#include <unistd.h>
+
+#include <bits/ensure.h>
+#include <mlibc/debug.hpp>
+
+/*
+ * The code in this file is largely based on glibc.
```
Geertiebear, line 11: This is a license conflict.

Geertiebear (changes requested):
Unfortunately we can't include GPL'd code in mlibc.

options/linux/generic/utmp-stubs.cpp
no92: The offending commits were dropped.

no92:
Dropped the offending commits and rebased onto master.

Dennisbonke (approved):
I believe that this is good to go now, not as extensive as it once was, but progress none the less and no more license violations! LGTM.

PR #735, opened 2022-09-06
Various function implementations from LFS

no92:


no92:
The ABI checker complaining is intended and should not result in an ABI break.

options/linux/include/sys/swap.h
```diff
@@ -0,0 +1,20 @@
+#ifndef _SYS_SWAP_H
+#define _SYS_SWAP_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define	SWAP_FLAG_PREFER		0x8000
```
Geertiebear, line 8: What happened to the formatting here?

options/glibc/include/sys/user.h
```diff
@@ -1,6 +1,8 @@
 #ifndef _SYS_USER_H
 #define _SYS_USER_H
 
+#include <mlibc/arch-defs.h>
```
avdgrinten, line 4: `mlibc/` headers must not be included from public headers. Use `bits/` instead.

options/linux/include/sys/swap.h
```diff
@@ -0,0 +1,20 @@
...
+#define SWAP_FLAG_DISCARD 0x10000
+
+int swapon(const char *, int);
+int swapoff(const char *);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif // _SYS_SWAP_H
```
Dennisbonke, line 20: We should use C style comments here.

Dennisbonke (changes requested):
LGTM, one comment.

Dennisbonke (approved):
LGTM

PR #737, opened 2022-09-07
IP structs and defines from LFS

no92:


Dennisbonke:
We should check if we aren't upstreaming the same thing twice. See #711 for things that might also be in this PR and the reason for the fact that it isn't merged yet.

no92:
Rebased onto current master.

Dennisbonke (approved):
LGTM

PR #749, opened 2022-09-30
sysdeps/managarm: convert file.cpp to helix_ng

no92:
Depends on managarm/managarm#480.


qookei (commented):
Seems good over all

no92:
Dropped `__raw_map`.

qookei (approved):

PR #750, opened 2022-10-08
Some small fixes for curl

Dennisbonke:
This PR addresses some missing error handling and stubs a setsockopt call that curl makes. Together with managarm/managarm#482 this fixes curl.

### DIFF
--- a/sysdeps/managarm/generic/file.cpp
+++ b/sysdeps/managarm/generic/file.cpp
@@ -1050,18 +1050,22 @@ int sys_poll(struct pollfd *fds, nfds_t count, int timeout, int *num_events) {
 
 	managarm::posix::SvrResponse<MemoryAllocator> resp(getSysdepsAllocator());
 	resp.ParseFromArray(recv_resp.data(), recv_resp.length());
-	__ensure(resp.error() == managarm::posix::Errors::SUCCESS);
-	__ensure(resp.events_size() == count);
+	if(resp.error() == managarm::posix::Errors::ILLEGAL_ARGUMENTS) {
+		return EINVAL;
+	} else {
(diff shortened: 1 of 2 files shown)


sysdeps/managarm/generic/file.cpp
```diff
@@ -1380,18 +1380,22 @@ int sys_poll(struct pollfd *fds, nfds_t count, int timeout, int *num_events) {
 
 	managarm::posix::SvrResponse<MemoryAllocator> resp(getSysdepsAllocator());
 	resp.ParseFromArray(recv_resp->data, recv_resp->length);
-	__ensure(resp.error() == managarm::posix::Errors::SUCCESS);
-	__ensure(resp.events_size() == count);
+	if(resp.error() == managarm::posix::Errors::ILLEGAL_ARGUMENTS) {
+		return EINVAL;
+	} else {
```
ArsenArsen, line 1385: this else isn't necessary, it's only effect is adding an unneeded indentation level (early-return already has the effect of skipping the subsequent code)

ArsenArsen (changes requested):

sysdeps/managarm/generic/file.cpp
Dennisbonke: While true, I like the clearness it gives. It's also not breaking the current code style as a lot of the sysdeps do this too. Personally, I'd keep it as is unless asked to do otherwise.

ArsenArsen: we have always preferred early return, though? also, this adds no clarity, only a bigger patch

64: odd thing to nitpick

no92 (approved):

PR #755, opened 2022-10-12
options/rtdl: Load scope fix

Dennisbonke:
If the local load scope is requested at object load, but said local scope is null, we happily pass a nullptr down which later blows up in our face. To avoid this, I introduced a simple check to see if the local scope is requested, and if so, if the local scope to be used is set to something that's not a nullptr. This fixes `hexchat` on Managarm once more.

### DIFF
--- a/options/rtdl/generic/main.cpp
+++ b/options/rtdl/generic/main.cpp
@@ -460,6 +460,13 @@ void *__dlapi_open(const char *file, int flags, void *returnAddress) {
 			return nullptr;
 		}
 
+		// If the local scope is requested, but the local scope is a nullptr, use the global scope anyway.
+		if(!(flags & RTLD_GLOBAL)) {
+			if(object->localScope == nullptr) {
+				object->localScope = globalScope.get();
+			}
+		}
+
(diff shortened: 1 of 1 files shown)


options/rtdl/generic/main.cpp
```diff
@@ -460,6 +460,13 @@ void *__dlapi_open(const char *file, int flags, void *returnAddress) {
 			return nullptr;
 		}
 
+		// If the local scope is requested, but the local scope is a nullptr, use the global scope anyway.
```
64, line 463: comment is a bit redundant

did you figure out why loadScope is null? is it for ld.so + libc.so + the executable itself?

Dennisbonke: I have no clue why, and if you want I can remove the comment.

no92: I was thinking that this behavior is incorrect, but posix says this about flags:
> If neither RTLD_GLOBAL nor RTLD_LOCAL are specified, then the default behavior is unspecified.

no92 (approved):

PR #757, opened 2022-10-13
rtdl: Use mprotect to remove superfluous permissions

streaksu:
In the name of security.

Tested on Ironclad and Lyre, where it worked, I would appreciate if someone tried it on linux or managarm for completeness

### DIFF
--- a/options/internal/include/mlibc/internal-sysdeps.hpp
+++ b/options/internal/include/mlibc/internal-sysdeps.hpp
@@ -34,6 +34,7 @@ int sys_close(int fd);
 // mlibc assumes that anonymous memory returned by sys_vm_map() is zeroed by the kernel / whatever is behind the sysdeps
 int sys_vm_map(void *hint, size_t size, int prot, int flags, int fd, off_t offset, void **window);
 int sys_vm_unmap(void *pointer, size_t size);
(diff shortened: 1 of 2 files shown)


qookei:
CI passed so it works on Linux on all supported architectures

no92 (approved):

qookei (approved):

ArsenArsen (approved):

avdgrinten:
Why do we need to add `PROT_WRITE` here if the current code doesn't do it?

qookei:
We memset the area after the data from the file with 0, and in the `!MLIBC_MAP_DSO_SEGMENTS` we map with `|PROT_WRITE` since we need to copy data into it. The code to drop perms was added outside of that check so it applies to both cases,

avdgrinten:
Ah, so the old code simply doesn't work if there are read-only segments with file size != mem size.

qookei:
Presumably, although we assume vm_map hands out zeroed memory? So in theory the memset shouldn't be needed.

avdgrinten:
The memset is needed to clear out the tail of the last page that is still part of the file.

qookei:
Ah alright. This PR seems good to merge then.

PR #758, opened 2022-10-18
Several fixes for claws to work

Dennisbonke:
While working on `claws`, we hit a few problems which I fixed. This PR includes an implementation of `fgets_unlocked`, `strxfrm`, a bunch of the `mntent` family of functions and two missing `strftime` specifiers. I also included tests for the missing specifiers and fixed a failing posix_spawn test under debian (no, not everyone has `/bin` symlinked to `/usr/bin`, debian unstable apparently doesn't).

### DIFF
--- a/options/ansi/generic/stdio-stubs.cpp
+++ b/options/ansi/generic/stdio-stubs.cpp
@@ -841,44 +841,9 @@ int fgetc(FILE *stream) {
 }
 
 char *fgets(char *__restrict buffer, size_t max_size, FILE *__restrict stream) {
-	__ensure(max_size > 0);
 	auto file = static_cast<mlibc::abstract_file *>(stream);
 	frg::unique_lock lock(file->_lock);
-	for(size_t i = 0; ; i++) {
-		if (i == max_size - 1) {
-			buffer[i] = 0;
-			return buffer;
-		}
-
-		auto c = fgetc_unlocked(stream);
-
-		// If fgetc() fails, there is either an EOF or an I/O error.
-		if(c == EOF) {
-//			if(ferror(stream)) {
-//				// Technically, we do not have to terminate the buffer in this case;
-//				// do it anyway to avoid UB if apps do not check our result.
-//				buffer[i] = 0;
-//				return nullptr;
-//			}
-
-			// EOF is only an error if no chars have been read yet.
-			//__ensure(feof(stream));
-			if(i) {
-				buffer[i] = 0;
-				return buffer;
-			}else{
-				// In this case, the buffer is not changed.
-				return nullptr;
-			}
-		}else{
-			buffer[i] = c;
-		}
-
-		if(c == '\n') {
-			buffer[i + 1] = 0;
-			return buffer;
-		}
-	}
+    return fgets_unlocked(buffer, max_size, stream);
 }
 
 int fputc_unlocked(int c, FILE *stream) {
@@ -1253,8 +1218,41 @@ size_t fwrite_unlocked(const void *buffer, size_t size, size_t count, FILE *file
 	}
 }
 
-char *fgets_unlocked(char *, int, FILE *) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
-}
+char *fgets_unlocked(char *__restrict buffer, int max_size, FILE *stream) {
+    __ensure(max_size > 0);
+    for(int i = 0; ; i++) {
+        if (i == max_size - 1) {
+            buffer[i] = 0;
+            return buffer;
+        }
 
+        auto c = fgetc_unlocked(stream);
+
+        // If fgetc() fails, there is either an EOF or an I/O error.
+        if(c == EOF) {
+            // if(ferror(stream)) {
+            //  // Technically, we do not have to terminate the buffer in this case;
+            //  // do it anyway to avoid UB if apps do not check our result.
+            //  buffer[i] = 0;
+            //  return nullptr;
+            // }
+
+            // EOF is only an error if no chars have been read yet.
+            //__ensure(feof(stream));
+            if(i) {
+                buffer[i] = 0;
+                return buffer;
+            }else{
+                // In this case, the buffer is not changed.
+                return nullptr;
+            }
+        }else{
+            buffer[i] = c;
+        }
+
+        if(c == '\n') {
+            buffer[i + 1] = 0;
+            return buffer;
+        }
+    }
+}
--- a/options/ansi/generic/stdio-stubs.cpp
+++ b/options/ansi/generic/stdio-stubs.cpp
@@ -234,11 +234,11 @@ int rename(const char *path, const char *new_path) {
 
 int renameat(int olddirfd, const char *old_path, int newdirfd, const char *new_path) {
 	MLIBC_CHECK_OR_ENOSYS(mlibc::sys_renameat, -1);
-    if(int e = mlibc::sys_renameat(olddirfd, old_path, newdirfd, new_path); e) {
-        errno = e;
-        return -1;
-    }
-    return 0;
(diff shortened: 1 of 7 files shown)


no92 (approved):

tests/posix/posix_spawn.c
```diff
@@ -33,6 +33,6 @@ void run_cmd(char *cmd)
 }
 
 int main() {
-    run_cmd("/usr/bin/true");
+    run_cmd("/bin/true");
```
ArsenArsen, line 36: this command isn't the best option here, I fixed this already but forgot to upstream (sorry): https://github.com/ArsenArsen/mlibc/commit/3f224a3204d9151d374ea93c89928122e6e4c0ba

tests/ansi/strftime.c
```diff
@@ -17,5 +17,13 @@ int main() {
 	strftime(timebuf, sizeof timebuf, "%e", &tm);
 	assert(!strcmp(timebuf, result));
 
+	memset(timebuf, 0, sizeof timebuf);
+	strftime(timebuf, sizeof timebuf, "%x", &tm);
+	assert(!strcmp(timebuf, "03/08/21"));
+
+	memset(timebuf, 0, sizeof timebuf);
```
ArsenArsen, line 24: we use fn-style sizeof

options/ansi/generic/stdio-stubs.cpp
```diff
@@ -1257,8 +1257,41 @@ size_t fwrite_unlocked(const void *buffer, size_t size, size_t count, FILE *file
 	}
 }
 
-char *fgets_unlocked(char *, int, FILE *) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
-}
+char *fgets_unlocked(char *__restrict buffer, int max_size, FILE *stream) {
```
ArsenArsen, line 1260: reimplement fgets in terms of this + locking

ArsenArsen (changes requested):

Dennisbonke:
@ArsenArsen good calls on all of those. I'll fix them later today, thanks for pointing them out. I'll indeed apply your commit for fixing the test, assuming it fixes the problem I've been having.

options/ansi/generic/stdio-stubs.cpp
```diff
@@ -1257,8 +1257,41 @@ size_t fwrite_unlocked(const void *buffer, size_t size, size_t count, FILE *file
...
+        if(c == EOF) {
+            // if(ferror(stream)) {
+            //  // Technically, we do not have to terminate the buffer in this case;
+            //  // do it anyway to avoid UB if apps do not check our result.
+            //  buffer[i] = 0;
+            //  return nullptr;
+            // }
+
+            // EOF is only an error if no chars have been read yet.
+            //__ensure(feof(stream));
```
avdgrinten, line 1280: Why is this commented out?

Dennisbonke: I don't know, I copied it from `fgets()` where it is also commented out.

options/internal/generic/locale.cpp
```diff
@@ -73,6 +73,10 @@ char *nl_langinfo(nl_item item) {
 				__ensure(!"ABDAY_* constants don't seem to be contiguous!");
 				__builtin_unreachable();
 		}
+	}else if(item == D_FMT) {
+		return const_cast<char *>("%m/%d/%y");
+	}else if(item == T_FMT) {
+		return const_cast<char *>("%H:%M:%S");
```
ArsenArsen, line 79: This is invalid (you can't edit literals, and it's specified that these should be a static buffer that the caller can edit)

ArsenArsen (changes requested):
Looks like it's possible to add a mntent test suite, since you can specify which fstab to edit, that should be considered also.
I'm not entirely sure about how our printf implementation, so someone else will have to check the validity of that.

options/ansi/generic/stdio-stubs.cpp
```diff
@@ -841,44 +841,9 @@ int fgetc(FILE *stream) {
...
-		}else{
-			buffer[i] = c;
-		}
-
-		if(c == '\n') {
-			buffer[i + 1] = 0;
-			return buffer;
-		}
-	}
+    return fgets_unlocked(buffer, max_size, stream);
```
qookei, line 846: Wrong indentation

options/ansi/generic/stdio-stubs.cpp
```diff
@@ -1257,8 +1222,41 @@ size_t fwrite_unlocked(const void *buffer, size_t size, size_t count, FILE *file
 	}
 }
 
-char *fgets_unlocked(char *, int, FILE *) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
-}
+char *fgets_unlocked(char *__restrict buffer, int max_size, FILE *stream) {
```
qookei, line 1225: Wrong indentation in the whole function

options/ansi/generic/string-stubs.cpp
```diff
@@ -101,9 +101,12 @@ int strncmp(const char *a, const char *b, size_t max_size) {
...
 
-size_t strxfrm(char *__restrict, const char *__restrict, size_t) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+size_t strxfrm(char *__restrict dest, const char *__restrict src, size_t n) {
+	size_t l = strlen(src);
+	if(n > l) {
+		strcpy(dest, src);
+	}
+	return l;
```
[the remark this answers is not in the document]
qookei: I think this should also work for generic UTF-8 collation? Assuming it just goes by codepoint order. But yeah adding a comment would be good.

options/internal/generic/locale.cpp
qookei: This whole function needs to be fixed in that case, as all the above returns also do this pattern.

options/ansi/generic/string-stubs.cpp
qookei, line 109: This should be `strncpy(dest, src, n)`? The man page says:
> The first  n  bytes  of  the  transformed string  are  placed  in  dest.

options/linux/generic/mntent-stubs.cpp
```diff
@@ -1,34 +1,39 @@
...
+	return fprintf(f, "%s\t%s\t%s\t%s\t%d\t%d\n",
+		mnt->mnt_fsname, mnt->mnt_dir, mnt->mnt_type, mnt->mnt_opts,
+		mnt->mnt_freq, mnt->mnt_passno) < 0;
 }
 
-int endmntent(FILE *) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+int endmntent(FILE *f) {
+	if(f) {
```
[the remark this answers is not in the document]
Dennisbonke: We do not.

Dennisbonke:
> Looks like it's possible to add a mntent test suite, since you can specify which fstab to edit, that should be considered also. I'm not entirely sure about how our printf implementation, so someone else will have to check the validity of that.

Bionic only ships tests for the functions that I'm not implementing here, so honestly I'm fine with leaving it as is. I'll address the other comments shortly.

options/ansi/generic/time-stubs.cpp
```diff
@@ -333,6 +339,7 @@ size_t strftime(char *__restrict dest, size_t max_size,
 			break;
 		}
 		default:
+			mlibc::infoLogger() << "mlibc: strftime unknown format type: " << c << frg::endlog;
```
64, line 342: panicLogger? and nuke ensure below

64 (approved):

options/ansi/generic/time-stubs.cpp
Dennisbonke: Fixed

PR #761, opened 2022-10-19
options/internal: fix compilation on clang

no92:
Previously, guarding the different `operator delete`s was done in the name of different mangling, which according to `c++filt` is not the case. To rectify this, both `operator delete`s will now get compiled in, as they do take different arguments.


ArsenArsen:
I'm not a fan of this solution, it involves runtime warnings for something that can definitely be detected at link time (or better). I'll see if there's something better that we can do (e.g. poisoning these functions and telling LD to give up on them).

Also, a better way to catch both GCC and Clang is to define all overloads of these functions, probably (see libsupc++)

ArsenArsen:
okay, `#pragma GCC poison delete` *will* block all instances of delete that follow it, so if we use that in common (internal) headers, that should be okay; however, that doesn't work for new (since it'd conflict with placement new), so I'm looking for further solutions

ArsenArsen:
here's a different fix that also covers other cases (note that I haven't tested it on clang, but I don't see why it wouldn't work)

```patch
From 420d8f2aed8a84a9903820229b29fdf3b5e8c727 Mon Sep 17 00:00:00 2001
From: =?UTF-8?q?Arsen=20Arsenovi=C4=87?= <arsen@aarsen.me>
Date: Wed, 19 Oct 2022 11:40:52 +0200
Subject: [PATCH] options/internal: enable the use of -z defs

note: don't apply this commit without splitting it, there's two changes
here (removing instances of operator delete(...) in deleting destructors
in virtual classes and related hacks, and then enabling defs)

The member operator deletes are defined to prevent the default of
calling libsupc++'s version being called (see the Itanium ABI[1], "The
entries for virtual destructors ...").

[1] https://itanium-cxx-abi.github.io/cxx-abi/abi.html
---
 meson.build                                   |  2 +-
 options/ansi/include/mlibc/file-io.hpp        |  4 ++++
 options/internal/gcc-extra/cxxabi.cpp         | 16 ----------------
 options/internal/generic/charcode.cpp         |  2 ++
 options/internal/include/mlibc/charcode.hpp   |  2 ++
 options/posix/generic/posix_stdio.cpp         |  2 ++
 options/posix/include/mlibc/posix-file-io.hpp |  2 ++
 7 files changed, 13 insertions(+), 17 deletions(-)
 delete mode 100644 options/internal/gcc-extra/cxxabi.cpp

diff --git a/meson.build b/meson.build
index 131415da..bfbbc00c 100644
--- a/meson.build
+++ b/meson.build
@@ -60,6 +60,7 @@ if not headers_only
 	add_project_arguments('-fno-rtti', '-fno-exceptions', language: 'cpp')
 	add_project_link_arguments('-nostdlib', language: ['c', 'cpp'])
+	add_project_link_arguments('-Wl,-z,defs', language: ['c', 'cpp'])
 
 	searchdirs = run_command(c_compiler.cmd_array(), '-print-search-dirs',
 				check: true).stdout()
@@ -231,7 +232,6 @@ internal_sources = [
 	'options/internal/gcc/stack_protector.cpp',
 	'options/internal/gcc/guard-abi.cpp',
 	'options/internal/gcc/initfini.cpp',
-	'options/internal/gcc-extra/cxxabi.cpp',
 	'options/internal' / host_machine.cpu_family() / 'setjmp.S',
 ]
 
diff --git a/options/ansi/include/mlibc/file-io.hpp b/options/ansi/include/mlibc/file-io.hpp
index 15b2c192..85a7c17a 100644
--- a/options/ansi/include/mlibc/file-io.hpp
+++ b/options/ansi/include/mlibc/file-io.hpp
@@ -48,6 +48,8 @@ public:
 	int tell(off_t *current_offset);
 	int seek(off_t offset, int whence);
 
+	void operator delete(void*) { __builtin_trap(); }
+
 protected:
 	virtual int determine_type(stream_type *type) = 0;
 	virtual int determine_bufmode(buffer_mode *mode) = 0;
@@ -85,6 +87,8 @@ struct fd_file : abstract_file {
 
 	static int parse_modestring(const char *mode);
 
+	void operator delete(void*) { __builtin_trap(); }
+
 protected:
 	int determine_type(stream_type *type) override;
 	int determine_bufmode(buffer_mode *mode) override;
diff --git a/options/internal/gcc-extra/cxxabi.cpp b/options/internal/gcc-extra/cxxabi.cpp
deleted file mode 100644
index aa222a83..00000000
--- a/options/internal/gcc-extra/cxxabi.cpp
+++ /dev/null
@@ -1,16 +0,0 @@
-
-#include <bits/ensure.h>
-
-// The cxxabi needs operator delete for *deleting* destructors, i.e., destructors that
-// are called by delete expressions. We never use such expressions in mlibc.
-// Note that G++ complains if we make the operator hidden,
-// thus we use it's mangled name as a workaround.
-#if defined(__clang__)
-	extern "C" [[gnu::visibility("hidden")]] void _ZdlPv() { // operator delete (void *, size_t)
-		__ensure(!"operator delete called! delete expressions cannot be used in mlibc.");
-	}
-#else
-	extern "C" [[gnu::visibility("hidden")]] void _ZdlPvm() { // operator delete (void *, size_t)
-		__ensure(!"operator delete called! delete expressions cannot be used in mlibc.");
-	}
-#endif
\ No newline at end of file
diff --git a/options/internal/generic/charcode.cpp b/options/internal/generic/charcode.cpp
index e09d5cd2..49dc1597 100644
--- a/options/internal/generic/charcode.cpp
+++ b/options/internal/generic/charcode.cpp
@@ -222,6 +222,8 @@ struct polymorphic_charcode_adapter : polymorphic_charcode {
 
 		return charcode_error::null;
 	}
+
+	void operator delete(void*) { __builtin_trap(); }
 };
 
 polymorphic_charcode *current_charcode() {
diff --git a/options/internal/include/mlibc/charcode.hpp b/options/internal/include/mlibc/charcode.hpp
index 67bd03d9..e666145a 100644
--- a/options/internal/include/mlibc/charcode.hpp
+++ b/options/internal/include/mlibc/charcode.hpp
@@ -107,6 +107,8 @@ struct polymorphic_charcode {
 
 	// Whether the encoding has shift states.
 	const bool has_shift_states;
+
+	void operator delete(void*) { __builtin_trap(); }
 };
 
 polymorphic_charcode *current_charcode();
diff --git a/options/posix/generic/posix_stdio.cpp b/options/posix/generic/posix_stdio.cpp
index f38dc658..d705735e 100644
--- a/options/posix/generic/posix_stdio.cpp
+++ b/options/posix/generic/posix_stdio.cpp
@@ -24,6 +24,8 @@ struct popen_file : mlibc::fd_file {
 		_popen_pid = new_pid;
 	}
 
+	void operator delete(void*) { __builtin_trap(); }
+
 private:
 	// Underlying PID in case of popen()
 	pid_t _popen_pid;
diff --git a/options/posix/include/mlibc/posix-file-io.hpp b/options/posix/include/mlibc/posix-file-io.hpp
index dd38fa82..e375b2e6 100644
--- a/options/posix/include/mlibc/posix-file-io.hpp
+++ b/options/posix/include/mlibc/posix-file-io.hpp
@@ -11,6 +11,8 @@ struct mem_file : abstract_file {
 	mem_file(char **ptr, size_t *sizeloc, void (*do_dispose)(abstract_file *) = nullptr);
 
 	int close() override;
+
+	void operator delete(void*) { __builtin_trap(); }
 protected:
 	int determine_type(stream_type *type) override;
 	int determine_bufmode(buffer_mode *mode) override;
-- 
2.38.0

```

Dennisbonke:
I'm curious to find out what @avdgrinten @Geertiebear @qookei think of this. Which approach should we pick.

tests/meson.build
```diff
@@ -148,6 +148,18 @@ use_pie = false
 test_c_args = []
 test_link_args = []
 
+host_test_c_args = []
+
+c_compiler = meson.get_compiler('c')
+
+if c_compiler.get_id() == 'gcc'
+	test_c_args += []
+	test_c_host_test_c_argsargs += []
```
qookei, line 157: argsargs

no92:
This PR is ready now, but the Github UI looks funny due to the fact that the required jobs were changed (by adding the compiler to the matrix).

options/posix/generic/syslog-stubs.cpp
```diff
@@ -106,8 +106,8 @@ static void _vsyslog(int priority, const char *message, va_list ap) {
 	strftime(timebuf, sizeof timebuf, "%b %e %T", &tm);
 
 	pid = (log_opt & LOG_PID) ? getpid() : 0;
-	l = snprintf(buf, sizeof buf, "<%d>%s %n%s%s%.0d%s: ",
-		priority, timebuf, &hlen, log_ident, "[" + !pid, pid, "]" + !pid);
+	l = snprintf(buf, sizeof buf, "<%d>%s %n%s%c%.0d%c: ",
+		priority, timebuf, &hlen, log_ident, ('[' + !pid), pid, (']' + !pid));
```
qookei, line 110: This is wrong, the previous code skipped the `[` and `]` (by adding 1 to skip the character and get an empty string) from the output if `!pid` was true.

options/posix/generic/pthread-stubs.cpp
```diff
@@ -47,7 +47,7 @@ struct ScopeTrace {
 #define SCOPE_TRACE() ScopeTrace(__FILE__, __LINE__, __FUNCTION__)
 
 static constexpr unsigned int mutexRecursive = 1;
-static constexpr unsigned int mutexErrorCheck = 2;
+[[maybe_unused]] static constexpr unsigned int mutexErrorCheck = 2;
```
qookei, line 50: How come this is seen as unused?

options/ansi/generic/wchar-stubs.cpp
```diff
@@ -190,6 +189,8 @@ size_t wcsrtombs(char *mbs, const wchar_t **wcsp, size_t mb_limit, mbstate_t *st
 		return size;
 	}
 
+	mlibc::code_seq<char> nseq{mbs, mbs + mb_limit};
```
qookei, line 192: Does this also need the `frg::min(..., MB_CUR_MAX)` like in `stdlib-stubs.cpp`?

options/internal/generic/threads.cpp
```diff
@@ -93,7 +93,7 @@ static constexpr unsigned int mutex_owner_mask = (static_cast<uint32_t>(1) << 30
 static constexpr unsigned int mutex_waiters_bit = static_cast<uint32_t>(1) << 31;
 
 // Only valid for the internal __mlibc_m mutex of wrlocks.
-static constexpr unsigned int mutex_excl_bit = static_cast<uint32_t>(1) << 30;
+[[maybe_unused]] static constexpr unsigned int mutex_excl_bit = static_cast<uint32_t>(1) << 30;
```
qookei, line 96: I think this shouldn't be here at all? It's also defined in `pthread-stubs.cpp` and used there.

tests/rtld/meson.build
```diff
@@ -14,7 +14,11 @@ rtld_test_cases = [
 	'tls_align',
 ]
 
-foreach test_name : rtld_test_cases 
+host_libc_rtld_nosan_test_cases = [
+	'preinit',
```
qookei, line 18: Does `rtld_next` also need to be in this list? It's native libraries are given `b_sanitize=none`.

tests/ansi/creal-cimag.c
```diff
@@ -18,6 +18,7 @@
     )
 
 int main() {
+#if !(defined(USE_HOST_LIBC) && defined(__clang__))
```
qookei, line 21: What's up with this? Does clang with glibc produce significantly different results?

qookei (commented):
I don't see the original issue (the `operator delete` stuff) addressed anywhere in this PR?

options/posix/generic/pthread-stubs.cpp
qookei: Oh this is similar to `mutex_excl_bit`, but the other way around. It's also defined and used in `threads.cpp`, and IG should be removed from here.

no92:
> I don't see the original issue (the `operator delete` stuff) addressed anywhere in this PR?

That got dropped as I didn't run into that issue any longer.

tests/ansi/creal-cimag.c
no92: The commit message has an explanation for this.

tests/rtld/meson.build
no92: This list only applies `b_sanitize=none` to the host-libc test executable, and is unrelated to the test libraries (that are handles separately anyways).

options/ansi/generic/wchar-stubs.cpp
no92: No, as for `wcsrtombs` the limit is purely given by the `mb_limit` function parameter.

no92:
Addressed the issues raised.

tests/ansi/creal-cimag.c
qookei: Ah, my bad.

qookei (approved):

PR #768, opened 2022-11-08
Various header additions for systemd

Dennisbonke:
HARD BLOCKED ON #766, #773, #776, #823, #824, #825, #826 (all have been merged!)

~~This PR is my internal staging area for systemd. I will cherry-pick commits into smaller chunks, that are actually reviewable, as this is of course not doable.~~

This was the systemd master staging PR, which has been split out in the PRs linked above. I found it fitting to finish with it. This PR adds some defines, adds one struct, and stubs 2 functions that don't need to be exported in a header.

### DIFF
--- a/options/linux/generic/capabilities.cpp
+++ b/options/linux/generic/capabilities.cpp
@@ -1,0 +1,19 @@
+#include <mlibc/debug.hpp>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+int capset(void *, void *) {
+	mlibc::infoLogger() << "mlibc: capset is a no-op!" << frg::endlog;
+	return 0;
+}
+
+int capget(void *, void *) {
+	mlibc::infoLogger() << "mlibc: capget is a no-op!" << frg::endlog;
+	return 0;
+}
+
+#ifdef __cplusplus
+}
+#endif
--- a/abis/mlibc/errno.h
+++ b/abis/mlibc/errno.h
@@ -119,4 +119,5 @@
 #define ECOMM 1113
 #define EBADE 1114
 #define EHWPOISON 1115
+#define EBADRQC 1116
 #endif // _ABIBITS_ERRNO_H
--- a/abis/mlibc/errno.h
+++ b/abis/mlibc/errno.h
@@ -118,4 +118,5 @@
 #define EKEYEXPIRED 1112
 #define ECOMM 1113
 #define EBADE 1114
+#define EHWPOISON 1115
 #endif // _ABIBITS_ERRNO_H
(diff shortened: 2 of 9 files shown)


Dennisbonke:
The last big part of the monster systemd PR is finally here! This PR finishes most of the legwork of getting `systemd` to compile. A few controversial changes have been left out, and some changes have been dropped in favor of work from the mlibc LFS project. This is part 8 of the systemd master PR, finalizing it.

options/linux/generic/capabilities.cpp
```diff
@@ -0,0 +1,15 @@
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+int capset(void *, void *) {
```
no92, line 5: These functions don't seem to be part of libc, at least on my host. They aren't declared anywhere, either.

options/posix/include/net/if.h
```diff
@@ -47,6 +48,22 @@ struct ifreq {
 		char *ifru_data;
 	} ifr_ifru;
 };
+#define ifr_name ifr_ifrn.ifrn_name
```
no92, line 51: These defines are duplicate, see right below.

no92 (commented):
Other than the two complaints, LGTM.

options/linux/generic/capabilities.cpp
ikbenlike: they're part of the linux capabilities API, should be declared in `sys/capability.h`

no92: The headers seems to be part of `libcap` though.

ikbenlike: they would have to be implemented as syscalls I think - implementing them here and providing an empty libcap object seems like the nicest option to me

Dennisbonke: I followed musl here, which does this and then calls the linux syscall.

Dennisbonke: See [this](https://github.com/bminor/musl/blob/7d756e1c04de6eb3f2b3d3e1141a218bb329fcfb/src/linux/cap.c#L8).

Geertiebear: If these are stubbed then they should at least log a message that they are so.

Dennisbonke: As requested, I've added a message.

options/posix/include/net/if.h
Dennisbonke: Dropped the offending commit.

Geertiebear (approved):
LGTM

PR #770, opened 2022-11-14
linux-headers: add linux/fuse.h file

ikbenlike:
This PR adds a linux kernel header necessary for supporting FUSE. As far as I can tell, no other changes would need occur in mlibc to implement FUSE in managarm later on.

### DIFF
--- a/options/linux-headers/include/linux/fuse.h
+++ b/options/linux-headers/include/linux/fuse.h
@@ -1,0 +1,837 @@
+/**
+ * The contents of this file are sourced from a header provided
+ * with a Linux kernel version 5-lts. It was originally distributed
+ * under GPL-2.0 or a BSD-2-clause license.*/
+
+#ifndef _LINUX_FUSE_H
+#define _LINUX_FUSE_H
+
+#include <stdint.h>
+
+/*
+ * Version negotiation:
+ *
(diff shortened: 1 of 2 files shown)


Geertiebear (approved):

ikbenlike:
all checks passed and the changes have been approved - any idea when this could get merged?

Dennisbonke:
I'd love to merge this but I have a few questions. It seems like you just straight up copied the entire file, including comments. I personally got told that's a no no, so awaiting input of @avdgrinten on that. Furthermore, and that is more of a personal nitpick, we can probably do some styling changes and we usually only add what we _really_ need in the Linux headers (unless you need all this stuff then disregard this point, then only the style changes holds).

ikbenlike:
The style could be changed and the comments could be removed if necessary - but the structs themselves and so on are part of the interface as defined by linux, and things like libfuse depend on this interface. This stuff isn't strictly necessary as of yet, but if FUSE support were to get added to managarm this stuff will become necessary. The main reason I PR'ed this is because it's quite simple, and past this there aren't any other changes that would need to happen to mlibc for future FUSE support

Dennisbonke:
Closing as we now use the actual kernel headers.

PR #773, opened 2022-12-26

Dennisbonke:
Batch number 2, again split off from #768. Nothing too big, a few function stubs and mostly kernel headers for the remaining part. Also adds one additional errno constant in the mlibc abi.


options/posix/generic/posix_signal.cpp
```diff
@@ -131,3 +131,8 @@ int sigaltstack(const stack_t *__restrict ss, stack_t *__restrict oss) {
 
 	return 0;
 }
+
+int sigisemptyset(const sigset_t *) {
+	__ensure(!"sigpending() not implemented");
```
qookei, line 136: Wrong message, also this is easy to implement (`return !(*set);` or so)

options/posix/include/bits/posix/posix_signal.h
```diff
@@ -91,6 +91,8 @@ int sigtimedwait(const sigset_t *__restrict set, siginfo_t *__restrict info, con
 int sigwait(const sigset_t *__restrict set, int *__restrict sig);
 int sigwaitinfo(const sigset_t *__restrict set, siginfo_t *__restrict info);
 
+int sigisemptyset(const sigset_t *set);
```
qookei, line 94: This is a GNU extension, maybe wrap in a check for the glibc option?

options/posix/generic/posix_signal.cpp
Dennisbonke: Should be fixed.

options/posix/include/bits/posix/posix_signal.h
Dennisbonke: Good catch, wrapped as requested.

options/linux-headers/include/mtd/mtd-user.h
```diff
@@ -0,0 +1,23 @@
...
+	uint8_t type;
+	uint32_t flags;
+	uint32_t size;
+	uint32_t erasesize;
+	uint32_t writesize;
+	uint32_t oobsize;
+	uint64_t padding;
+};
+
+#define MEMGETINFO _IOR('M', 1, struct mtd_info_user)
```
64, line 19: does this work

64 (approved):

options/linux-headers/include/mtd/mtd-user.h
Dennisbonke: As in, is this define okay? Should be, matches Linux.

PR #779, opened 2023-01-22
Some small things for Chromium

Dennisbonke:
We found some missing items for Chromium while @czapek1337 was building chromium under cbuildrt. This includes some missing signal constants in the Linux ABI. This PR fixes them.

### DIFF
--- a/options/glibc/include/resolv.h
+++ b/options/glibc/include/resolv.h
@@ -29,11 +29,15 @@ int res_init(void);
 /* From musl: Unused; purely for broken apps
  * To avoid an massive struct, only add the items requested. */
 typedef struct __res_state {
+	int retrans;
+	int retry;
 	unsigned long options;
 	int nscount;
 	struct sockaddr_in nsaddr_list[MAXNS];
 	char *dnsrch[MAXDNSRCH + 1];
 	char defdname[256];
+	unsigned ndots:4;
+	unsigned nsort:4;
 	union {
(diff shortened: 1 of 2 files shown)


options/glibc/include/resolv.h
```diff
@@ -29,11 +29,15 @@ int res_init(void);
 /* From musl: Unused; purely for broken apps
  * To avoid an massive struct, only add the items requested. */
 typedef struct __res_state {
+	int retrans;
```
64, line 32: abi break?

Dennisbonke: Technically yes, but we merged additions to this struct before without it being an abi-break. I'll defer to [that discussion](https://github.com/managarm/mlibc/pull/553) for my reasoning to merge this anyway.

ArsenArsen: yeah, it doesn't matter here, it's never used in abi (which is why it shouldn't have ever been added).  it's possible some programs use it on library boundaries, but I don't see that happening, since this tends to be just used through that one pointer

ArsenArsen: presuming something does store values here (but never creates new instances of res_state), which seems the most likely scenario in practice, this would lead to a collision if applied with programs already relying on current location.

there's no need to match anything with an api that there's no intention of implementing, and that has no business existing in the first place other than to serve as a stub

PR #784, opened 2023-01-29
Remove the `linux-headers` option, instead allow for supplying original headers

no92:
The ABI check is expected to fail in this PR, as the master that's being compared against doesn't support external kernel headers.


ArsenArsen:
ISTR that struct ifreq stuff elsewhere, and some discussion about whether it's even an ABI break.  Could you check other PRs for it?

no92:
Added a section in the README explaining how the `linux_kernel_headers` meson option works.

.github/workflows/ci.yml
```diff
@@ -70,12 +70,27 @@ jobs:
...
           - name: Compile sysdeps
             run: |
-                meson \
+                meson setup \
                     "-Dc_args=['-fno-stack-protector', '-U__linux__']" \
                     "-Dcpp_args=['-fno-stack-protector', '-U__linux__']" \
                     "-Dbuild_tests=true" \
                     "-Db_sanitize=undefined" \
-                    --cross-file ci/${{matrix.sysdeps}}.cross-file compile-${{matrix.sysdeps}}
+                    "-Db_sanitize=undefined" \
```
Dennisbonke, line 91: This is double.

Dennisbonke (changes requested):
LGTM but one change

.github/workflows/ci.yml
64, line 91: duplicate line?

64 (approved):

Dennisbonke (approved):
LGTM

PR #741, opened 2022-09-13
glibc: implement getopt_long_only

no92:

PR #787, opened 2023-02-02
options/glibc: make getopt aware of optreset param

cleanbaja:


cleanbaja:
already implemented in #741

PR #793, opened 2023-02-10
Posix wait

ikbenlike:
This PR extends the functionality of `wait` and `waitpid` to also except NULL as argument, since the standard specifies this is possible. It also adds the glibc `error` header, and implements the functions therein.


options/glibc/generic/error.cpp
```diff
@@ -0,0 +1,73 @@
...
+unsigned int error_message_count = 0;
+int error_one_per_line = 0;
+void (*error_print_progname)(void) = NULL;
+
+void error(int status, int errnum, const char *format, ...) {
+	va_list args;
+	va_list copied;
+	va_start(args, format);
+	va_copy(copied, args);
+	va_end(args);
```
ArsenArsen, line 17: I use va_* APIs maybe once a year at this point, so not entirely sure if I'm right, but: isn't the copy here redundant?

qookei: As far as I know a copy is only needed if you want to iterate through the same varargs twice (e.g. when calling 2 functions that take a `va_list` with the same list).

ikbenlike: I think I remember something about needing to copy the list of reusing it, but I'm not 100% sure and this API isn't really my strong suit. If this isn't required, would I simply be able to initialize the list and pass it on to `vfprintf`?

qookei: Yeah if you want to use the same list twice you need to make a copy, but in this case you only use it to pass it on to `vfprintf`, in which case you can just pass the original one through.

ikbenlike: Makes sense; I'll change this later.

ikbenlike: This has been corrected.

tests/glibc/error.py
```diff
@@ -0,0 +1,7 @@
+import subprocess
+import sys
+from pyexpect import expect
+
+completed = subprocess.run([sys.argv[1]], check=True, capture_output=True)
```
ArsenArsen, line 5: this could be just `check_output` instead of run

options/glibc/include/error.h
```diff
@@ -0,0 +1,21 @@
...
+
+extern unsigned int error_message_count;
+extern int error_one_per_line;
+extern void (*error_print_progname)(void);
+	
+#ifdef __cplusplus
+}
+#endif
+	
+#endif // _ERROR_H
```
ArsenArsen, line 21: we can't use these comments in public headers

ArsenArsen (changes requested):

options/glibc/include/error.h
ikbenlike: why not?

qookei: C89 doesn't support single-line comments

tests/glibc/error.py
ikbenlike: This change has been made

options/glibc/include/error.h
ikbenlike: I removed it

ArsenArsen (approved):

options/glibc/include/error.h
Dennisbonke: `/* _ERROR_H */` is preferred.

ikbenlike: This change has been made.

options/glibc/generic/error.cpp
```diff
@@ -0,0 +1,67 @@
...
+void error(int status, int errnum, const char *format, ...) {
+	va_list args;
+	va_start(args, format);
+
+	error_message_count++;
+	
+	fflush(stdout);
+	if(error_print_progname) {
+		error_print_progname();
+	}
```
Dennisbonke, line 21: Please put the `else` on the same line as the closing bracket.

options/glibc/generic/error.cpp
```diff
@@ -0,0 +1,67 @@
...
+	if(!(last_line == linenum && error_one_per_line && !first_call)) {
+		first_call = false;
+		last_line = linenum;
+		error_message_count++;
+
+		fflush(stdout);
+		if(error_print_progname) {
+			error_print_progname();
+		}
+		else {
```
Dennisbonke, line 52: Please put the `else` on the same line as the closing bracket.

options/posix/generic/sys-wait-stubs.cpp
```diff
@@ -14,11 +14,15 @@ int waitid(idtype_t, id_t, siginfo_t *, int) {
...
 pid_t waitpid(pid_t pid, int *status, int flags) {
 	pid_t ret;
+	int tmp_status = 0;
 	MLIBC_CHECK_OR_ENOSYS(mlibc::sys_waitpid, -1);
-	if(int e = mlibc::sys_waitpid(pid, status, flags, NULL, &ret); e) {
+	if(int e = mlibc::sys_waitpid(pid, &tmp_status, flags, NULL, &ret); e) {
 		errno = e;
 		return -1;
 	}
+	if(status){
```
Dennisbonke, line 23: Missing space between `)` and `{`.

Dennisbonke (changes requested):
Code looks good, some style issues were found.

Dennisbonke (approved):
LGTM

PR #794, opened 2023-02-10
options/rtdl: set object->localScope to globalScope for global objects

qookei:
This fixes a crash that if an object was already in the global scope (as a result of for example DT_NEEDED), dlopening it again with RTLD_LOCAL would try to use object->localScope, which did not point to anything, while the object should remain in the global scope instead.

Also added a regression test for this.

Closes #755.


options/rtdl/generic/linker.cpp
```diff
@@ -242,6 +242,8 @@ SharedObject *ObjectRepository::requestObjectWithName(frg::string_view name,
 
 		// TODO: Free this when the scope is no longer needed.
 		localScope = frg::construct<Scope>(getAllocator());
+	} else if (localScope == nullptr) {
```
64, line 245: When will it be nullptr here?

qookei: When called from `dlopen(..., RTLD_GLOBAL)` `localScope` is null and `createScope` is false. Same for `requestObjectAtPath`.

64: Hmm... does it make sense to avoid passing `nullptr` in that case? Rather than enforcing it inside the function

qookei: Actually, IG it'd be better to pass the global scope in from the outside, since that's what the other caller of `requestObjectWithName` does

64: Yeah that's what I mean

64 (approved):

PR #798, opened 2023-02-17
options/ansi: Add missing install line

Dennisbonke:
Likely fixes some build errors we found.


Dennisbonke:
CI approves but vinix is still required even though it has been removed?

PR #802, opened 2023-03-02
lyre: add sysdeps for socketpair syscall

RaidTheWeb:


RaidTheWeb:
it simply happens to be so

ArsenArsen:
sounds fixable

PR #813, opened 2023-03-24
LFS: add posix timers

no92:


Dennisbonke (approved):
LGTM

Dennisbonke:
The abi break is expected, we changed the way we declare signam, but we still declare it.

PR #816, opened 2023-03-29
LFS: ABI fixes

no92:
This possibly needs to be merged into `abi-break` but that needs updating first.


options/linux/include/sys/epoll.h
```diff
@@ -40,7 +43,11 @@ typedef union epoll_data {
 struct epoll_event {
 	uint32_t events;
 	epoll_data_t data;
-};
+}
```
Geertiebear, line 46: Why not just always packed?

no92: My host (glibc, via a test program), musl and bionic seem to agree that this is correct and only necessary for x86_64.

Dennisbonke (approved):
LGTM outside that one comment.

Dennisbonke (approved):
LGTM

PR #812, opened 2023-03-13
ioctls rework

no92:
Blocked on managarm/managarm#519.

PR #818, opened 2023-03-30
sysdeps/managarm: rework ioctls to allow using dedicated messages

no92:
Split off from #812. This is the mlibc part of managarm/managarm#519.


Dennisbonke (approved):
LGTM

Geertiebear (approved):
LGTM as long as the managarm counterpart is also approved.

PR #819, opened 2023-03-31
Additions for `iproute2`

no92:


options/posix/generic/posix_string.cpp
```diff
@@ -132,9 +132,8 @@ char *strcasestr(const char *s, const char *pattern) {
 	return nullptr;
 }
 
-char *strdupa(const char *) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+char *strdupa(const char *str) {
+	return strcpy(reinterpret_cast<char *>(alloca(strlen(str) + 1)), str);
```
ikbenlike, line 136: is this code actually safe? from `man alloca`:
> This temporary space is automatically freed when the function that called alloca() returns to its caller.

In this case I would expect the destination buffer to be freed upon `strdupa` returning - for the sake of completeness you should write a test case specificially engineered to use the stack anyway, such that improper use of `alloca` will be detected by seeing that the destination buffer was corrupted

ikbenlike (changes requested):

options/posix/generic/posix_string.cpp
ArsenArsen: no

ikbenlike:
> Why should we take over the `_GNU_SOURCE` feature test macro? This seems like a glibc specific thing.

would it not be possible for some software to depend on it working as it does? although providing this stuff unconditionally could work as well, I suppose

tests/posix/strdupa.c
```diff
@@ -0,0 +1,14 @@
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
+#endif
+#include <assert.h>
+#include <string.h>
+
+int main() {
+	char test[19] = "Hello mlibc World!";
```
ikbenlike, line 8: would be good to add an expected fail test case that tries to `free` the `strdupa`d buffer

ikbenlike (changes requested):

options/internal/generic/ubsan.cpp
```diff
@@ -238,3 +238,17 @@ void __ubsan_handle_missing_return(UnreachableData *data) {
...
+	SourceLocation loc;
+	SourceLocation attr_loc;
+	int arg_index;
+};
+
+extern "C" [[gnu::visibility("hidden")]]
+void __ubsan_handle_nonnull_arg(NonNullArgData *data) {
+	mlibc::panicLogger()
+		<< LOG_NAME_LOC("null pointer passed to non-null argument", data->loc)
+		<< "argument " <<data->arg_index << " is required to be non-null in "
```
Dennisbonke, line 252: You're missing a space between `<<` and `data->arg_index`.

options/linux/include/bits/linux/linux_sched.h
```diff
@@ -4,6 +4,9 @@
 
 // Glibc extension
 int sched_getcpu(void);
+
+#if defined(_GNU_SOURCE)
 int setns(int fd, int nstype);
+#endif // _GNU_SOURCE
```
Dennisbonke, line 10: We should use C style comments here.

options/posix/include/bits/posix/posix_string.h
```diff
@@ -21,10 +22,23 @@ void *memccpy(void *__restrict dest, const void *__restrict src, int c, size_t n
...
+})
+#define strndupa(x, y) ({ \
+	const char *str = (x); \
+	size_t len = strnlen(str, (y)) + 1; \
+	char *buf = alloca(len); \
+	buf[len - 1] = '\0'; \
+	(char *) memcpy(buf, str, len - 1); \
+})
 void *memrchr(const void *, int, size_t);
+#endif // defined(_GNU_SOURCE)
```
Dennisbonke, line 41: We should use C style comments here.

tests/posix/strdupa.c
no92: That would be UB, though.

Dennisbonke (changes requested):
LGTM outside a few styling issues.

tests/posix/strdupa.c
ikbenlike: yes, hence it's expected to fail

Dennisbonke (approved):
LGTM

PR #824, opened 2023-04-02
More More stubs and header additions from #768 for systemd

Dennisbonke:
Mostly defines, a few new headers and 9 function stubs. Part 5 of the systemd master PR.


no92 (approved):
LGTM

PR #825, opened 2023-04-02

Dennisbonke:
Mostly defines and structs, adds 1 function stub. Part 6 of the systemd master PR.


no92 (approved):
LGTM

PR #826, opened 2023-04-03

Dennisbonke:
A few defines, 2 structs, 1 implemented function and 6 function stubs. Part 7 of the systemd master PR.


no92 (approved):
LGTM

PR #828, opened 2023-04-03
`CPU_*` macros

no92:
Closes #710.

### DIFF
--- a/options/internal/include/bits/cpu_set.h
+++ b/options/internal/include/bits/cpu_set.h
@@ -1,0 +1,13 @@
+#ifndef _MLIBC_INTERNAL_CPU_SET_H
+#define _MLIBC_INTERNAL_CPU_SET_H
+
+typedef unsigned long __cpu_mask;
+
+#define CPU_SETSIZE 1024
+#define __NCPUBITS (8 * sizeof(__cpu_mask))
+
+typedef struct {
+	__cpu_mask __bits[CPU_SETSIZE / __NCPUBITS];
+} cpu_set_t;
+
+#endif /* _MLIBC_INTERNAL_CPU_SET_H */
--- a/options/linux/include/bits/linux/cpu_set.h
+++ b/options/linux/include/bits/linux/cpu_set.h
@@ -13,9 +13,30 @@ extern "C" {
 cpu_set_t *__mlibc_cpu_alloc(int num_cpus);
 size_t __mlibc_cpu_alloc_size(int num_cpus);
 
+void __mlibc_cpu_zero(const size_t setsize, cpu_set_t *set);
+void __mlibc_cpu_set(const int cpu, const size_t setsize, cpu_set_t *set);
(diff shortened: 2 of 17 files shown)


options/posix/include/pthread.h
```diff
@@ -86,8 +86,10 @@ struct __mlibc_threadattr {
 	int __mlibc_inheritsched;
 	struct sched_param __mlibc_schedparam;
 	int __mlibc_schedpolicy;
+#if __MLIBC_LINUX_OPTION
```
Geertiebear, line 89: So, I don't really like this. For multiple reasons, one that this has to be an abi-break now which is kind of ugly. But it's also harmful that ABI now depends on compilation config. I'd much rather see this be included in the struct unconditionally, and that we have a definition for ``cpu_set_t`` internally (this can be in abi-bits maybe?). Then we can export ``cpu_set_t`` and the whole family of functions conditionally, but still have the ABI be constant.

Geertiebear (changes requested):

options/internal/include/bits/cpu_set.h
```diff
@@ -0,0 +1,13 @@
...
+typedef unsigned long __cpu_mask;
+
+#define CPU_SETSIZE 1024
+#define __NCPUBITS (8 * sizeof(__cpu_mask))
+
+typedef struct {
+	__cpu_mask __bits[CPU_SETSIZE / __NCPUBITS];
+} cpu_set_t;
+
+#endif // _MLIBC_INTERNAL_CPU_SET_H
```
Dennisbonke, line 13: We should use C style comments here.

options/linux/include/bits/linux/cpu_set.h
```diff
@@ -0,0 +1,45 @@
...
+#define CPU_ISSET(cpu, set) CPU_ISSET_S(cpu, sizeof(cpu_set_t), set)
+
+#define CPU_COUNT_S(setsize, set) __mlibc_cpu_count((setsize), (set))
+#define CPU_COUNT(set) CPU_COUNT_S(sizeof(cpu_set_t), set)
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif // _LINUX_CPU_SET_H
```
Dennisbonke, line 45: We should use C style comments here.

options/posix/include/pthread.h
```diff
@@ -204,6 +206,7 @@ int pthread_getattr_np(pthread_t, pthread_attr_t *);
 
 int pthread_getaffinity_np(pthread_t thread, size_t cpusetsize, cpu_set_t *cpuset);
 int pthread_setaffinity_np(pthread_t thread, size_t cpusetsize, const cpu_set_t *cpuset);
+#endif // __MLIBC_LINUX_OPTION
```
Dennisbonke, line 209: We should use C style comments here.

Dennisbonke (changes requested):
LGTM outside a few styling issues.

Dennisbonke (approved):
LGTM

PR #829, opened 2023-04-05
build: Add disable_libgcc_dependency meson option

mintsuki:


Geertiebear (approved):
LGTM, what is the usecase for this?

ArsenArsen (changes requested):
logical operators exist, you can use those (as well as a toolchain that isn't broken enough to require this flag)

mintsuki:
I know full well what I am doing. The toolchain is not broken.

mintsuki:
> LGTM, what is the usecase for this?

@Geertiebear So that the build system won't hard depend on a search path available libgcc and I can manually pass a prebuild libgcc.a without it complaining.

mintsuki:
@ArsenArsen I am not sure what the requested changes are supposed to be.

ArsenArsen:
> @ArsenArsen I am not sure what the requested changes are supposed to be.

```
if a
    if b
    endif
endif
```

to

```
if a and b
endif
```

mintsuki:
> > @ArsenArsen I am not sure what the requested changes are supposed to be.
> 
> ```
> if a
>     if b
>     endif
> endif
> ```
> 
> to
> 
> ```
> if a and b
> endif
> ```

@ArsenArsen I'm not sure where you think doing that would be beneficial in this patch.

lukflug:
Since both conditions are negated, one could use De Morgan to further transform the expression.

ArsenArsen (approved):
oh, my bad, I misread the indentation level of the original conditions, there's indeed no `if { if { } }`-type nesting

PR #831, opened 2023-04-06
LFS: additions part 8

no92:


options/linux/include/scsi/sg.h
```diff
@@ -42,6 +47,29 @@ struct sg_io_hdr {
     unsigned int info;
 };
 
+struct sg_scsi_id {
```
ikbenlike, line 50: is this sort of stuff not provided by the linux-headers option?

no92: While we don't have a `linux-headers` option any more, these header files are explicitly not provided by the kernel.

ikbenlike: in that case we would indeed have to provide these ourselves. I think that addresses all my concerns with this PR

ikbenlike (approved):
LGTM

options/glibc/include/bits/glibc/glibc_assert.h
```diff
@@ -0,0 +1,28 @@
...
+#endif
+
+void __assert_fail_perror(int errno, const char *file, unsigned int line,
+		const char *function);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif // MLIBC_GLIBC_ASSERT_H
```
Dennisbonke, line 15: We should use C style comments here.

options/glibc/include/bits/glibc/glibc_assert.h
```diff
@@ -0,0 +1,28 @@
...
+
+#ifdef NDEBUG
+
+#undef assert_perror
+#define assert_perror(ignore) ((void)0)
+
+#else // NDEBUG
```
Dennisbonke, line 22: We should use C style comments here.

options/glibc/include/bits/glibc/glibc_assert.h
```diff
@@ -0,0 +1,28 @@
...
+
+#undef assert_perror
+#define assert_perror(errno) (!(errno) \
+		|| (__assert_fail_perror((errno), __FILE__, __LINE__, __func__), 0))
+
+#endif // NDEBUG
```
Dennisbonke, line 28: We should use C style comments here.

Dennisbonke (changes requested):
LGTM outside 3 styling issues.

Geertiebear (approved):
LGTM

Dennisbonke (approved):
LGTM

PR #832, opened 2023-04-06
LFS: part 9, network

no92:


options/posix/generic/net-if-stubs.cpp
```diff
@@ -1,12 +1,14 @@
...
 
 #include <bits/ensure.h>
+#include <mlibc/debug.hpp>
 #include <mlibc/posix-sysdeps.hpp>
 
-void if_freenameindex(struct if_nameindex *) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+void if_freenameindex(struct if_nameindex *idx) {
+	(void) idx;
```
ikbenlike, line 10: nitpick, but: you can leave the argument unnamed to avoid having to cast it to void

options/posix/generic/net-if-stubs.cpp
```diff
@@ -21,8 +23,8 @@ char *if_indextoname(unsigned int index, char *name) {
 }
 
 struct if_nameindex *if_nameindex(void) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+	errno = ENOSYS;
```
ikbenlike, line 26: log the fact that this is unimplemented

ikbenlike (changes requested):

Dennisbonke (approved):
LGTM

Geertiebear (approved):
LGTM

PR #833, opened 2023-04-06
LFS: part 10, sysdeps

no92:


options/posix/generic/sys-socket-stubs.cpp
```diff
@@ -98,6 +90,7 @@ ssize_t recv(int sockfd, void *__restrict buf, size_t len, int flags) {
 
 ssize_t recvfrom(int sockfd, void *__restrict buf, size_t len, int flags,
 		struct sockaddr *__restrict src_addr, socklen_t *__restrict addrlen) {
+#ifndef __linux__
```
ikbenlike, line 93: should this not take place in the implementation of the sysdep?

options/posix/generic/sys-socket-stubs.cpp
```diff
@@ -140,6 +142,7 @@ ssize_t send(int fd, const void *buffer, size_t size, int flags) {
 
 ssize_t sendto(int fd, const void *buffer, size_t size, int flags,
 		const struct sockaddr *sock_addr, socklen_t addr_length) {
+#ifndef __linux__
```
ikbenlike, line 145: same as with `recvfrom`

ikbenlike (commented):
from `recvfrom(2)`:
> The recv() call is normally used only on a connected socket (see connect(2) and is identical to recvfrom() with a NULL src_addr argument.

I am unsure if mlibc already provides `recv`, however it might be worth it to simplify the implementation of it like this (and if not yet provided, it would be relatively simple to do so)

ikbenlike:
I just checked and `recv` is already implemented as a call to `recvfrom`, and the fallback in case there is no sysdep is now handled more elegantly. personally I can't think of anything else that should be changed in this PR

Dennisbonke (approved):
LGTM

options/posix/include/mlibc/posix-sysdeps.hpp
```diff
@@ -166,7 +168,7 @@ int sys_vm_unmap(void *pointer, size_t size);
 // as an infinite timeout, making sigtimedwait(..., nullptr) equivalent to sigwaitinfo(...)
 [[gnu::weak]] int sys_sigtimedwait(const sigset_t *__restrict set, siginfo_t *__restrict info, const struct timespec *__restrict timeout, int *out_signal);
 [[gnu::weak]] int sys_kill(int, int);
-[[gnu::weak]] int sys_accept(int fd, int *newfd, struct sockaddr *addr_ptr, socklen_t *addr_length);
+[[gnu::weak]] int sys_accept(int fd, int *newfd, struct sockaddr *addr_ptr, socklen_t *addr_length, int flags);
```
Geertiebear, line 171: The change of this sysdep should be propagated to other ports as well. Also make sure that the functionality stays the same (so copy the old ``fcntl()`` trick to the sysdeps as well).

Geertiebear (changes requested):

Geertiebear (approved):
LGTM now.

PR #834, opened 2023-04-06
LFS: part 11, structs

no92:


ikbenlike (approved):
LGTM

Dennisbonke (approved):
LGTM

options/glibc/include/sys/user.h
```diff
@@ -42,4 +42,8 @@ struct user {
 }
 #endif
 
+#define PAGE_SHIFT 12
```
Geertiebear, line 45: What is this and is it platform dependent?

no92: It's apparently a define for gdb, and it is not platform-dependent (in glibc + musl). It does not reflect true page size, as you should use the function for that.

PR #835, opened 2023-04-07
LFS: part 12

no92:


options/ansi/generic/stdlib-stubs.cpp
```diff
@@ -297,14 +297,27 @@ void *bsearch(const void *key, const void *base, size_t count, size_t size,
 
 	return nullptr;
 }
+
+static int qsort_callback(const void *a, const void *b, void *arg) {
+	auto compare = (int (*)(const void *, const void *)) arg;
```
ikbenlike, line 302: would a `reinterpret_cast` here not be cleaner?

options/linux/generic/mntent-stubs.cpp
```diff
@@ -1,16 +1,23 @@
...
 #include <mntent.h>
 #include <stdio.h>
+#include <limits.h>
 #include <string.h>
 #include <bits/ensure.h>
 
+static char *internal_buf;
+static size_t internal_bufsize;
+
+#define SENTINEL (char *)&internal_buf
```
ikbenlike, line 12: I am not a fan of these being global. is there no cleaner way to achieve the same or similar (for example passing in null to `getmntent_r` and checking against that instead)

no92: I have placed `internal_buf(size)` in an anonymous namespace.

options/glibc/include/sys/kd.h
```diff
@@ -0,0 +1,17 @@
...
+#endif
+
+#include <linux/kd.h>
+
+#ifdef __undef_LINUX_TYPES_H
+# undef _LINUX_TYPES_H
+# undef __undef_LINUX_TYPES_H
+#endif
+
+#endif // _SYS_KD_H
```
Dennisbonke, line 17: We should use C style comments here.

options/posix/include/bits/posix/posix_stdio.h
```diff
@@ -29,6 +30,21 @@ int vdprintf(int fd, const char *format, __gnuc_va_list args);
 
 char *fgetln(FILE *, size_t *);
 
+// GNU extensions
```
Dennisbonke, line 33: We should use C style comments here.

Dennisbonke (changes requested):
LGTM outside 2 style comments.

options/ansi/generic/stdlib-stubs.cpp
```diff
@@ -297,14 +297,27 @@ void *bsearch(const void *key, const void *base, size_t count, size_t size,
...
+
+	return compare(a, b);
+}
+
 void qsort(void *base, size_t count, size_t size,
 		int (*compare)(const void *, const void *)) {
+	return qsort_r(base, count, size, qsort_callback, (void *) compare);
+}
+
+void qsort_r(void *base, size_t count, size_t size,
```
Geertiebear, line 312: Could you add a test for qsort?

Geertiebear (changes requested):
In general not a fan of the mtent code but we should probably have some mlibc wide code to do with successive file reading. So nothing to change about that for now. I would like to see a test for ``qsort()``, and it would also be nice if we could have one for the mtent functions (can be done by taking some dummy fstab file and verifying the code against it).

ikbenlike:
I agree with Geert here. A test would be good, but some of the code quality changes are out of the scope of this PR. When that test gets added, LGTM as far as I'm concerned

Dennisbonke (approved):
LGTM

PR #836, opened 2023-04-07
implement `malloc_usable_size`

no92:
Hard-depends on managarm/frigg#59. CI will also fail until that is merged.

### DIFF
--- a/options/glibc/generic/malloc.cpp
+++ b/options/glibc/generic/malloc.cpp
@@ -1,0 +1,6 @@
+#include <bits/glibc/glibc_malloc.h>
+#include <mlibc/allocator.hpp>
+
+size_t malloc_usable_size(void *p) {
+	return getAllocator().get_size(p);
+}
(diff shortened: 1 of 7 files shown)


Geertiebear:
As discussed on discord, this change might not even be desirable, and uses of ``malloc)usable_size()`` should probably be fixed in upstream packages.

no92:
Given as glibc, mlibc and bionic support it and systemd uses it (correctly now, since version 253) I don't see a lot of room for calling this not desirable, especially given that the implementation is trivial and doesn't involve additional overhead.

Dennisbonke (approved):
LGTM

ikbenlike (changes requested):
from my understanding this is a GNU extension, not linux-specific. if that is the case it should probably go in the glibc option

Dennisbonke:
> from my understanding this is a GNU extension, not linux-specific. if that is the case it should probably go in the glibc option

Good catch, the man page seems to agree. We should probably move this to the glibc option indeed.

Dennisbonke (changes requested):
We should move this to the glibc option.

no92:
Moved the function to `options/glibc`.

Geertiebear (approved):
LGTM.

PR #837, opened 2023-04-07
LFS: part 13, implement freopen

no92:


Dennisbonke (approved):
LGTM

options/posix/generic/posix-file-io.cpp
```diff
@@ -15,6 +15,10 @@ int mem_file::close() {
 	return 0;
 }
 
+int mem_file::reopen(const char *, const char *) {
+	return -1;
```
Geertiebear, line 19: This is incorrect, an ``freopen()`` call on a memstream is valid, and should return a ``FILE*`` conforming to the input arguments. It should also flush the memstream and close it.

Geertiebear (changes requested):

options/posix/generic/posix-file-io.cpp
no92: Due to the large rework of both `FILE` and frigg internals needed for this, I have made this a panic for now.

Geertiebear: That's fine, we can add an issue for that and fix it later.

Geertiebear (approved):
LGTM now.

no92:
Done.

PR #843, opened 2023-04-09
abi: Add __MLIBC_ABI_ONLY macro to suppress external references

mintsuki:


Dennisbonke (commented):
LGTM but I'd like a sign off from korona or Geert.

Geertiebear (approved):
LGTM for right now, this might be fixed in a better way by completely separating all ABI definitions from function definitions.

PR #845, opened 2023-04-10
misc: Do not rely on magic macros to modify freestanding header behaviour

mintsuki:
This PR makes it so mlibc does not make assumptions about magic modifier macros to change the behaviour of *non* mlibc-provided freestanding headers.

### DIFF
--- a/options/internal/include/bits/size_t.h
+++ b/options/internal/include/bits/size_t.h
@@ -1,7 +1,6 @@
 #ifndef MLIBC_SIZE_T_H
 #define MLIBC_SIZE_T_H
 
-#define __need_size_t
-#include <stddef.h>
+typedef __SIZE_TYPE__ size_t;
 
 #endif // MLIBC_SIZE_T_H
--- a/options/internal/include/bits/wchar_t.h
+++ b/options/internal/include/bits/wchar_t.h
@@ -2,8 +2,11 @@
 #ifndef MLIBC_WCHAR_T_H
 #define MLIBC_WCHAR_T_H
 
-#define __need_wchar_t
-#include <stddef.h>
+#ifndef __cplusplus
+
+typedef __WCHAR_TYPE__ wchar_t;
+
+#endif
 
 #endif // MLIBC_WCHAR_T_H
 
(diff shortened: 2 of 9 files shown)


options/internal/include/bits/size_t.h
```diff
@@ -1,7 +1,6 @@
 #ifndef MLIBC_SIZE_T_H
 #define MLIBC_SIZE_T_H
 
-#define __need_size_t
-#include <stddef.h>
+typedef __SIZE_TYPE__ size_t;
```
avdgrinten, line 4: Won't think break if somebody includes both this file (indirectly) and GCC's `stddef.h`?

options/internal/include/bits/wchar_t.h
```diff
@@ -2,8 +2,11 @@
 #ifndef MLIBC_WCHAR_T_H
 #define MLIBC_WCHAR_T_H
 
-#define __need_wchar_t
-#include <stddef.h>
+#ifndef __cplusplus
+
+typedef __WCHAR_TYPE__ wchar_t;
+
+#endif
```
avdgrinten, line 9: Same here.

options/internal/include/bits/wint_t.h
```diff
@@ -1,7 +1,6 @@
 #ifndef MLIBC_WINT_T_H
 #define MLIBC_WINT_T_H
 
-#define __need_wint_t
-#include <stddef.h>
+typedef __WINT_TYPE__ wint_t;
```
avdgrinten, line 4: Same here.

options/internal/include/bits/size_t.h
mintsuki: As long as they are the same type, which they should be else `stddef.h` would be broken, it should not break.

options/internal/include/bits/wchar_t.h
mintsuki: Same as above.

options/internal/include/bits/wint_t.h
mintsuki: Same as above.

Dennisbonke (approved):
LGTM

PR #851, opened 2023-04-11
sysdeps/linux: Use the correct ABI for gid_t.h

mintsuki:


Dennisbonke (approved):
LGTM, thanks for the fix!

PR #852, opened 2023-04-11
misc: Remove root /abis directory and move headers to /sysdeps

mintsuki:
Get rid of the root `/abis` directory and move all the ABI headers to `/sysdeps`.

As discussed in the Discord, the `/abis` directory is mostly redundant and the separation is not useful.


sysdeps/dripos/include/abi-bits/signal.h
qookei, line 1: `managarm/include/abi-bits/signal.h` is a symlink to the Linux `signal.h`, not a copy of the mlibc ABI `signal.h`

sysdeps/dripos/include/abi-bits/termios.h
qookei, line 1: `managarm/include/abi-bits/termios.h` is a symlink to the Linux `termios.h`, not the mlibc ABI `termios.h`

sysdeps/dripos/include/abi-bits/wait.h
qookei, line 1: `managarm/include/abi-bits/wait.h` is a symlink to the Linux `wait.h`, not the mlibc ABI `wait.h`

qookei (changes requested):

mintsuki:
@qookei I have moved the mlibc ABI to `/sysdeps/legacy-mlibc-abi` as a compromise, whilst the rest is the same as it was before this force push, save for stuff that you mentioned that should've been fixed.

Dennisbonke:
> I've already voiced concern over this approach on Discord, but will do it here as well for clarity's sake: I don't like the relative symlinking happening inside the sysdep directory to pick ABI's. This is very fragile, and will mean ABI (or now also sysdep) changes will be cumbersome to propagate. 

I second this opinion. For me, it feels wrong to symlink to other sysdeps.

mintsuki:
Since we're just reposting what was said on the Discord now, I will restate how "feels wrong" isn't a technical reason to do or not do something technical.

Geertiebear:
"Feels wrong" in this case has been supported by an actual argument, giving rise to this "wrong" feeling.

PR #853, opened 2023-04-12
C11 threads

no92:


options/internal/generic/threads.cpp
```diff
@@ -0,0 +1,65 @@
+#include <abi-bits/errno.h>
+#include <bits/ensure.h>
+#include <mlibc/all-sysdeps.hpp>
+#include <mlibc/debug.hpp>
+#include <mlibc/threads.hpp>
+#include <mlibc/tcb.hpp>
+#include <pthread.h>
```
ikbenlike: doesn't this mean the internal thread API still depends on pthreads?

ikbenlike (changes requested):

options/ansi/generic/threads.cpp
```diff
@@ -0,0 +1,13 @@
+#include <abi-bits/errno.h>
+#include <mlibc/threads.hpp>
+#include <threads.h>
+
+int thrd_create(thrd_t *thr, thrd_start_t func, void *arg) {
+	int res = mlibc::thread_create(thr, 0, (void * (*)(void *)) func, arg);
```
ikbenlike, line 6: I think casting this and then calling the result is UB, and I think it could also cause problems in cases where `sizeof(int) != sizeof(void*)`. instead I think you could add two arguments corresponding to each of the possible function pointers, or perhaps a union or some such thing. regardless I think this cast should be avoided

ikbenlike (changes requested):

options/internal/generic/threads.cpp
```diff
@@ -0,0 +1,70 @@
...
+#include <mlibc/all-sysdeps.hpp>
+#include <mlibc/debug.hpp>
+#include <mlibc/threads.hpp>
+#include <mlibc/tcb.hpp>
+
+extern "C" Tcb *__rtdl_allocateTcb();
+
+namespace mlibc {
+
+int thread_create(struct __mlibc_thread_data **__restrict thread, const struct __mlibc_threadattr *__restrict attrp, void *entry, void *__restrict user_arg, bool returnsInt) {
```
ikbenlike, line 13: minor nitpick: `returnsInt` doesn't match the style of the other arguments

options/internal/include/mlibc/tcb.hpp
```diff
@@ -79,15 +79,24 @@ namespace mlibc {
 #endif
 }
 
+enum TcbThreadReturnValue {
```
ikbenlike, line 82: make this an enum class instead

options/internal/include/mlibc/tcb.hpp
```diff
@@ -79,15 +79,24 @@ namespace mlibc {
...
 	int tid;
 	int didExit;
-	void *returnValue;
+	union {
+		void * voidPtr;
+		int intVal;
+	} returnValue;
 	uintptr_t stackCanary;
 	int cancelBits;
+	enum TcbThreadReturnValue returnValueType;
```
ikbenlike, line 99: I don't think you need to prefix the name with `enum`

sysdeps/linux/generic/thread.cpp
```diff
@@ -15,10 +15,17 @@ extern "C" void __mlibc_enter_thread(void *entry, void *user_arg) {
 	while(!__atomic_load_n(&tcb->tid, __ATOMIC_RELAXED))
 		mlibc::sys_futex_wait(&tcb->tid, 0, nullptr);
 
-	void *(*func)(void *) = reinterpret_cast<void *(*)(void *)>(entry);
-	auto result = func(user_arg);
+	static_assert(sizeof(void *) >= sizeof(int));
+	static_assert(sizeof(uintptr_t) >= sizeof(int));
```
ikbenlike, line 19: these asserts seem unnecessary

sysdeps/managarm/x86_64/thread.cpp
```diff
@@ -14,12 +14,19 @@ extern "C" void __mlibc_enter_thread(void *entry, void *user_arg, Tcb *tcb) {
 	if(mlibc::sys_tcb_set(tcb))
 		__ensure(!"sys_tcb_set() failed");
 
-	void *(*func)(void *) = reinterpret_cast<void *(*)(void *)>(entry);
-	auto result = func(user_arg);
+	static_assert(sizeof(void *) >= sizeof(int));
+	static_assert(sizeof(uintptr_t) >= sizeof(int));
```
ikbenlike, line 18: same story here

no92:
Given that this PR now supports the parts of C11 threads that `foot` uses, I'd consider it complete for now.

options/internal/include/mlibc/tcb.hpp
```diff
@@ -79,15 +79,24 @@ namespace mlibc {
...
+
 struct Tcb {
 	Tcb *selfPointer;
 	size_t dtvSize;
 	void **dtvPointers;
 	int tid;
 	int didExit;
-	void *returnValue;
+	union {
+		void * voidPtr;
```
ikbenlike, line 94: unnecessary space in `* voidPtr`

sysdeps/aero/generic/thread.cpp
```diff
@@ -17,12 +17,17 @@ extern "C" void __mlibc_enter_thread(void *entry, void *user_arg, Tcb *tcb) {
...
 
-    void *(*func)(void *) = reinterpret_cast<void *(*)(void *)>(entry);
-    auto result = func(user_arg);
+	if(tcb->returnValueType == TcbThreadReturnValue::ThreadReturnsVoidPtr) {
+		void *(*func)(void *) = reinterpret_cast<void *(*)(void *)>(entry);
+		tcb->returnValue.voidPtr = func(user_arg);
+	} else {
+		int (*func)(void *) = reinterpret_cast<int (*)(void *)>(entry);
+		tcb->returnValue.intVal = func(user_arg);
+	}
```
ikbenlike, line 26: this could just be `auto`, currently you express the same thing (the type of the variable) twice. this goes for all the sysdeps

ikbenlike (approved):
LGTM

sysdeps/aero/generic/thread.cpp
```diff
@@ -17,12 +17,17 @@ extern "C" void __mlibc_enter_thread(void *entry, void *user_arg, Tcb *tcb) {
     if (mlibc::sys_tcb_set(tcb))
         __ensure(!"sys_tcb_set() failed");
 
-    void *(*func)(void *) = reinterpret_cast<void *(*)(void *)>(entry);
-    auto result = func(user_arg);
+	if(tcb->returnValueType == TcbThreadReturnValue::ThreadReturnsVoidPtr) {
```
Geertiebear, line 20: Might be nice to create a helper function for this so sysdeps don't have to concern themselves with this particular peculiarity.

ikbenlike: something that returns an `std::variant<void*,int>` or something along those lines?

Geertiebear: No i'd rather see an ``mlibc::invoke_thread_func(tcb, user_arg)``, or even a ``Tcb::invoke_thread_func(user_arg)`` function.

Geertiebear: The result is only used to set the return value field in the TCB.

ikbenlike: ah, true, I hadn't thought of that. in that case I'd agree adding this as a function to `Tcb` is probably the cleanest, especially because this is basically just boilerplate that every sysdep would repeat

no92:
Addressed the comments raised above and rebased on current master.

sysdeps/aero/generic/thread.cpp
```diff
@@ -17,12 +17,10 @@ extern "C" void __mlibc_enter_thread(void *entry, void *user_arg, Tcb *tcb) {
     if (mlibc::sys_tcb_set(tcb))
         __ensure(!"sys_tcb_set() failed");
 
-    void *(*func)(void *) = reinterpret_cast<void *(*)(void *)>(entry);
-    auto result = func(user_arg);
+	tcb->invokeThreadFunc(entry, user_arg);
```
Geertiebear, line 20: Indentation is wrong here.

Geertiebear (approved):
One style issue, but for the rest LGTM. Can be merged when you fix the indent.

PR #854, opened 2023-04-13
`uchar.h` additions for foot

no92:


options/ansi/generic/uchar.cpp
```diff
@@ -0,0 +1,23 @@
+#include <bits/ensure.h>
+#include <uchar.h>
+#include <wchar.h>
+
+size_t c32rtomb(char *, char32_t, mbstate_t *) MLIBC_STUB_BODY
+
+size_t mbrtoc32(char32_t *pc32, const char *pmb, size_t max, mbstate_t *ps) {
```
ikbenlike, line 7: afaik these arguments should all be marked `restrict`

options/ansi/include/string.h
```diff
@@ -51,8 +51,11 @@ void *memset(void *dest, int c, size_t size);
 char *strerror(int errnum);
 size_t strlen(const char *s);
 
-// POSIX extensions.
+#if defined(_BSD_SOURCE) || defined(_GNU_SOURCE)
```
ikbenlike, line 54: should this not instead check if the POSIX option is enabled?

no92: Yeah this should also check if the the header is enabled, but the feature test macros are correct here.

no92:
Addressed the comments raised and rebased onto current `master`.

options/ansi/generic/uchar.cpp
```diff
@@ -0,0 +1,23 @@
+#include <bits/ensure.h>
+#include <uchar.h>
+#include <wchar.h>
+
+size_t c32rtomb(char *, char32_t, mbstate_t *) MLIBC_STUB_BODY
+
+size_t mbrtoc32(char32_t *__restrict pc32, const char *__restrict pmb, size_t max, mbstate_t *__restrict ps) {
```
Geertiebear, line 7: Would be nice to add a test for this.

options/ansi/include/string.h
```diff
@@ -51,8 +51,11 @@ void *memset(void *dest, int c, size_t size);
 char *strerror(int errnum);
 size_t strlen(const char *s);
 
-// POSIX extensions.
+#if __MLIBC_POSIX_OPTION && (defined(_BSD_SOURCE) || defined(_GNU_SOURCE))
```
Dennisbonke, line 54: Not sure if we want to do this, discussion about it in the past were not in favor. I can't merge this before this is solved. Korona needs address this.

Dennisbonke (changes requested):
One issue, and the tests.

options/ansi/include/string.h
no92: Programs can define or not define the feature test macros, which switches whether this is being included here or not.

Geertiebear: If downstream programs require this I'm fine with it. It's a little ugly but doesn't seem like a hill to die on. I'm not aware of the discussion Dennis is referring to, so if someone can find an argument against it from that conversation I'm all ears.

Dennisbonke: [Here](https://discord.com/channels/678608576151027731/678613590646849576/805403505258266654) until the end of that day for the first real discussion I found. Korona preferred patching over including strings.h with the feature test macro. If @avdgrinten signs off on this then I'm happy to merge it and dump a bunch of patches, but without his approval I can't sign off on this (unless I read the discussion in discord in a different way, then I just ask what I've been doing the past two years).

ElectrodeYT: I'm personally in favor of merging this; maintaining patches for ports which would otherwise not need any patches at all is not sustainable in the long run and is also an absolute waste of time. We can not possible break anything by auto-including strings.h in certain situations, and its unlikely to massively affect compile time either.

Sure, this may not be completely standard, but standards are not always followed, and this is such a minute detail with big effects that not merging this because "it's not the standard" when many apps believe it to be the standard is unproductive.

Dennisbonke (approved):
LGTM

PR #859, opened 2023-04-16
misc: Use external freestanding headers implementation

mintsuki:


Geertiebear:
What does this PR fix?

mintsuki:
It is not a fix for anything. It makes mlibc rely on an external, more complete implementation of the freestanding C headers instead of its internal ones which were rather incomplete.

mintsuki:
Refer to `limits.h` for example. In general they are incomplete enough that relying on an external implementation instead of NIH'ing them makes some sense, but it is subject to personal preference.

ikbenlike:
as I mentioned before on discord, I approve of this idea but I do not have the means to verify the correctness of this particular implementation. I would say that the rationale is similar as with the `linux-headers` situation, in that there is no need to maintain a set of headers like this ourselves

Dennisbonke:
Not against this. However, this would break Managarm if I'm not mistaken, can you prepare a PR for bootstrap-managarm too?

mintsuki:
> Not against this. However, this would break Managarm if I'm not mistaken, can you prepare a PR for bootstrap-managarm too?

@Dennisbonke see bootstrap-managarm#232

Geertiebear:
> Refer to limits.h for example. In general they are incomplete enough that relying on an external implementation instead of NIH'ing them makes some sense, but it is subject to personal preference.

I'd prefer then that we just embed the external headers into mlibc then. Are they really so complex and change so often that we should rely on an external repository? Seems better to me to just add the missing parts to mlibc.

PR #860, opened 2023-04-16
internal/essential: Optimise memmove() slightly

mintsuki:


ikbenlike:
these changes look fine, however would an `reinterpret_cast` as also done in `memcpy` and `memset` not be cleaner? not really an issue with your PR, but if you're changing that function anyway you can just as well change that cast while you're at it

mintsuki:
Not sure where you want me to put the `reinterpret_cast`.

ikbenlike:
in place of the C-style casts on line 199 and 200

mintsuki:
Ah, well, that isn't code that I am touching in this PR. If you want that fixed, make a PR that fixes that part specifically instead.

options/internal/generic/essential.cpp
```diff
@@ -199,9 +199,8 @@ void *memmove(void *dest, const void *src, size_t size) {
 	char *dest_bytes = (char *)dest;
 	char *src_bytes = (char *)src;
 	if(dest_bytes < src_bytes) {
-		for(size_t i = 0; i < size; i++)
-			dest_bytes[i] = src_bytes[i];
-	}else{
+		memcpy(dest, src, size);
```
no92, line 202: I don't think that this is correct; `memcpy` does not make any guarantees that the data will be copied forwards and not backwards. This branch however only ensures that a forwards copy is safe, which is why the for loop was in place previously.

no92 (changes requested):
I believe that this commit needs to be reverted due to the first line relying on implementation-specific behavior that might realistically change. The change to the else should be good though.

options/internal/generic/essential.cpp
no92: More specifically, memcopying overlapping regions is UB and is a classic bug that has lead to numerous vulnerabilities.

ikbenlike: but mlibc also provides `memcpy`, so would it not be fine to rely on it's inner workings due to the fact we provide it to begin with?

mintsuki: Yes, this is a non-issue since we also provide memcpy() and we build with `-fno-builtin`

mintsuki:
This commit does not need to be reverted. See resolved conversation.

options/internal/generic/essential.cpp
no92: Yes, we control memcopy, but again we do not guarantee how it operates. For example, if we were to provide Arch-specific optimized versions (e.g. using SIMD) this could reasonably be broken. As-is, this is just incorrect.

mintsuki: > Yes, we control memcopy, but again we do not guarantee how it operates. For example, if we were to provide Arch-specific optimized versions (e.g. using SIMD) this could reasonably be broken. As-is, this is just incorrect.

"This is just incorrect" is an overly harsh statement. As is, it is correct in that it works given our only implementation of `memcpy()` that exists. It could *potentially* become incorrect in the future if our implementation of `memcpy()` changes such that the assumption of copy direction is broken.

Until then, the most that is needed is a comment stating how `memmove()` depends on the copy direction of `memcpy()` and to be careful in case the latter is changed.

mintsuki: This is not a "technically correct" or "technically incorrect" a-ha gotcha discussion given *we* are the libc so even my usage of the word "assumption" is not really correct, since we *know* how `memcpy()` works by virtue of the fact that *we are* the ones implementing it.

PR #862, opened 2023-04-25
Improve meson.build

hanhlinux:
Library name should be without `lib`

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -328,7 +328,7 @@ if not headers_only
 	if not disable_libgcc_dependency
 		libgcc = meson.get_compiler('c').find_library('gcc', required: false)
 
-		compiler_rt_name = 'clang_rt.builtins-' + host_machine.cpu_family()
+		compiler_rt_name = 'libclang_rt.builtins-' + host_machine.cpu_family()
 		compiler_rt = meson.get_compiler('c').find_library(compiler_rt_name, required: false)
 
 		if libgcc.found()


Geertiebear (approved):
Good catch, thanks.

options/posix/generic/nl_types.cpp
```diff
@@ -0,0 +1,16 @@
+#include <nl_types.h>
+
+int catclose(nl_catd catd) 
+{
+	return 0;
```
Geertiebear, line 5: When stubbing please add a warning message that these don't do anything.

sysdeps/linux/include/bits/syscall.h
```diff
@@ -21,12 +21,14 @@ __sc_word_t __do_syscall7(long, __sc_word_t, __sc_word_t, __sc_word_t, __sc_word
 		__sc_word_t, __sc_word_t, __sc_word_t);
 long __do_syscall_ret(unsigned long);
 
+// We will need better syscall() macros in C++ because they cannot compiled with
+// libc++ 
+
+#warning C++ syscall() macros are disabled because of incompatibility with libc++
```
Geertiebear, line 27: This is a regression and I'd rather not merge that. Could you post the actual error when compiling ``libc++``?

options/posix/include/nl_types.h
```diff
@@ -1,6 +1,22 @@
...
+extern "C" {
+#endif
 
+#define NL_SETD 1
+#define NL_CAT_LOCALE 1
 
-#endif // NL_TYPES_H 
\ No newline at end of file
+typedef int nl_item;
+typedef void *nl_catd;
```
Geertiebear, line 12: I'd rather not make this a void pointer, we should add a struct and make this a pointer to the struct. Since this interface uses opaque types we can then easily change the internal struct later.

Geertiebear (changes requested):

options/posix/include/nl_types.h
hanhlinux: Oh I forgot to limit changes, Im doing some local changes. Sorry about that.

hanhlinux: Lemme open another pull request.

PR #863, opened 2023-05-01
unshim cxx, add crt[in].o to managarm sysdeps in place of the gcc ones

ArsenArsen:


ArsenArsen:
oh, right.  the ci will need gcc 13 for ubuntu.  i'll figure out how to make those packages later..

meson.build
```diff
@@ -90,6 +93,56 @@ if not headers_only
...
+	endforeach
+
+	relevant_started = false
+
+	foreach line : cpp_output
+		if relevant_started
+			if not line.startswith(' ')
+				break
+			endif
+			debug('maybe relevant', line)
```
Geertiebear, line 132: Should remove the debug prints

Geertiebear (changes requested):

meson.build
Geertiebear: Oh fair enough, I just assumed that if it works then these aren't necessary.

sysdeps/managarm/meson.build
```diff
@@ -115,25 +115,27 @@ if not no_headers
...
-		command: c_compiler.cmd_array() + ['-c', '-o', '@OUTPUT@', '@INPUT@'],
-		input: crt_src,
-		output: 'crt0.o',
-		install: true,
-		install_dir: get_option('libdir')
-	)
-
-	if host_machine.cpu_family() == 'x86_64'
-		crt_pie = custom_target('Scrt1',
+	crtstuff = ['crt0']
```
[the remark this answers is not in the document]
ArsenArsen: open to better names

no92 (approved):
LGTM

Dennisbonke (approved):
LGTM

ArsenArsen:
no, it will not do that, but I haven't done a merge yet since I haven't updated the CI to contain new toolchains

ArsenArsen:
should be done

.github/workflows/ci.yml
```diff
@@ -6,21 +6,20 @@ jobs:
     build-mlibc:
         strategy:
             matrix:
-                arch: [x86_64, riscv64, aarch64]
+                arch: [x86_64, aarch64]
```
Geertiebear, line 9: Why is ``riscv`` removed?

Geertiebear (commented):
Does this mean one needs to install build a cross compiler to build ``mlibc`` now?

.github/workflows/ci.yml
ArsenArsen: the port isn't complete enough to configure a cross compiler

ArsenArsen:
nothing changed in that regard, my guess is that old hacks still work

Geertiebear:
We never really used any hacks that I'm aware of, and I'm asking since every sysdep now has a cross compiler file that is not the system compiler, implying that we can no longer compile mlibc with the system compiler.

ArsenArsen:
using a mistargeted compiler should still work as before.  the current configuration just gets us GCC 13 with libstdc++ 13 in CI

.github/workflows/ci.yml
ArsenArsen: unsure, but this is the build log:
```
239.5 In file included from ../../../tool-src/libgcc/unwind-dw2.c:410:
239.5 ./md-unwind-support.h: In function 'riscv_fallback_frame_state':
239.5 ./md-unwind-support.h:67:6: warning: assignment to 'struct sigcontext *' from incompatible pointer type 'mcontext_t *' [-Wincompatible-pointer-types]
239.5    67 |   sc = &rt_->uc.uc_mcontext;
239.5       |      ^
239.5 ./md-unwind-support.h:77:53: error: invalid use of undefined type 'struct sigcontext'
239.5    77 |       fs->regs.reg[i].loc.offset = (_Unwind_Ptr) &sc->gregs[i] - new_cfa;
239.5       |                                                     ^~
239.5 ./md-unwind-support.h:84:21: error: invalid use of undefined type 'struct sigcontext'
239.5    84 |     (_Unwind_Ptr) sc->gregs[0] - new_cfa;
239.5       |                     ^~
239.5 make[2]: *** [../../../tool-src/libgcc/static-object.mk:17: unwind-dw2.o] Error 1
```

... slightly mangled by docker.  to reproduce it, you can use the script in mlibc-crossers by just adding a riscv target

Geertiebear: That seems lie an easy fix, seems like we're just missing a ``struct sigcontext``.

Geertiebear:
But using a mistargeted compiler required cxxshim. Is cxxshim still supported?

.github/workflows/ci.yml
ArsenArsen: sure, but I have no knowledge of riscv64, nor a way to test it.

Geertiebear: Right, but I'm also not going to merge a regression on the CI. I'll submit a PR to your branch later to fix this.

no92: This should get fixed once #945 is merged.

Dennisbonke (approved):
LGTM

PR #873, opened 2023-05-24
options/ansi: Stub missing C11 threads functions

Dennisbonke:
Fixes #853


Geertiebear (approved):
LGTM but obviously compile errors should be fixed.

PR #875, opened 2023-06-08
Implement `fmemopen`

### LINKED ISSUE #527: posix: fmemopen() unimplemented

no92:
Supersedes #540.
Fixes #527.

### DIFF
--- a/options/posix/generic/posix-file-io.cpp
+++ b/options/posix/generic/posix-file-io.cpp
@@ -71,8 +71,7 @@ int mem_file::io_seek(off_t offset, int whence, off_t *new_offset) {
 			*new_offset = _pos;
 			break;
 		default:
-			__ensure(!"Unknown whence value passed!");
-			__builtin_unreachable();
+			return EINVAL;
 	}
 	return 0;
 }
--- a/options/posix/generic/posix-file-io.cpp
+++ b/options/posix/generic/posix-file-io.cpp
@@ -94,6 +94,100 @@ void memstream_mem_file::_update_ptrs() {
 	*_sizeloc = _buf.size() - 1;
 }
 
+fmemopen_mem_file::fmemopen_mem_file(void *in_buf, size_t size, int flags, void (*do_dispose)(abstract_file *))
+: mem_file{flags, do_dispose}, _inBuffer{in_buf}, _inBufferSize{size} {
+	if(!_inBuffer) {
+		_inBuffer = getAllocator().allocate(size);
+		_needsDeallocation = true;
+	}
+
+	if(_flags & O_APPEND) {
+		// the initial seek-size for append is zero if buf was NULL, or the first '\0' found, or the size
+		_max_size = (_needsDeallocation) ? 0 : strnlen(reinterpret_cast<char *>(_inBuffer), _inBufferSize);
+		_pos = _max_size;
+	} else if((_flags & O_WRONLY || _flags & O_RDWR) && _flags & O_CREAT && _flags & O_TRUNC) {
+		// modes: "w", "w+"
+		_max_size = 0;
+	} else {
+		_max_size = size;
+	}
+}
+
+int fmemopen_mem_file::close() {
+	if(_needsDeallocation) {
+		getAllocator().free(_inBuffer);
+	}
+
+	return 0;
+}
+
+int fmemopen_mem_file::io_read(char *buffer, size_t max_size, size_t *actual_size) {
+	if ((_pos >= 0 && _pos >= _max_size) || !max_size) {
+		*actual_size = 0;
+		return 0;
+	}
+
+	size_t bytes_read = std::min(size_t(_max_size - _pos), max_size);
+	memcpy(buffer, _buffer().data() + _pos, bytes_read);
+	_pos += bytes_read;
+	*actual_size = bytes_read;
+	return 0;
+}
+
+int fmemopen_mem_file::io_write(const char *buffer, size_t max_size, size_t *actual_size) {
+	off_t bytes_write = std::min(static_cast<size_t>(_buffer_size() - _pos), max_size);
+	memcpy(_buffer().data() + _pos, buffer, bytes_write);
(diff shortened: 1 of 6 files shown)


no92:
The failing assert will not be done on glibc due to (incorrectly) failing there. NetBSD's libc tests also guard against this in the same way. My implementation is adhering to POSIX, and thus correct; musl and the BSDs agree with me here. This should be ready for review now.

ikbenlike (approved):
LGTM

Geertiebear (commented):
Haven't taken a good look at this yet, will do later, but just a nitpick already when skimming it over: the comments aren't in the same style :) In general we follow that comments should ``// Look like this.``, if you could fix that in the relevant places that would be really nice.

no92:
Fixed the comments.

options/posix/generic/posix-file-io.cpp
```diff
@@ -6,12 +6,39 @@
...
+: abstract_file{do_dispose}, _inBuffer{in_buf}, _inBufferSize{size}, _flags{flags},
+	_allowResizing{false} {
+	if(!_inBuffer) {
+		_inBuffer = getAllocator().allocate(size);
+		_needsDeallocation = true;
+	}
+
+	if(_flags & O_APPEND) {
+		// the initial seek-size for append is zero if buf was NULL, or the first '\0' found, or the size
+		_max_size = (_needsDeallocation) ? 0 : strnlen(reinterpret_cast<char *>(_inBuffer), _inBufferSize);
```
Geertiebear, line 25: While this is correct, there is missing functionality here, which is that append should also set the position of the stream to the first null byte. This functionality is also not tested in the tests afaict, which is why that hasn't been caught.

tests/posix/fmemopen.c
```diff
@@ -0,0 +1,119 @@
...
+	f = fmemopen(NULL, BUFFER0_SIZE, "a");
+	assert(f);
+
+	assert(!fseek(f, 0, SEEK_END));
+	pos = ftell(f);
+	assert(pos == 0);
+
+	fclose(f);
+
+	// test seek with mode "a" and a buffer containing a '\0'
```
Geertiebear, line 56: Missing test case: testing the initial state of the buffer as mandated by the spec.

tests/posix/fmemopen.c
```diff
@@ -0,0 +1,119 @@
...
+	// > use of fseek() or fseeko() on the stream with SEEK_END shall seek relative to this size.
+	// > For modes r and r+ the size shall be set to the value given by the size argument.
+#if !defined(__GLIBC__)
+	assert(pos == BUFFER0_SIZE);
+#endif
+
+	fclose(f);
+
+	// test seek with mode "w"
+	f = fmemopen(NULL, BUFFER0_SIZE, "w");
```
Geertiebear, line 37: Missing test case: testing the initial conditions of the buffer (the implementation should place a null byte at the first byte).

options/posix/include/mlibc/posix-file-io.hpp
```diff
@@ -21,13 +23,42 @@ struct mem_file : abstract_file {
 	int io_seek(off_t offset, int whence, off_t *new_offset) override;
 private:
 	void _update_ptrs();
+
+	// Convenience function to obtain the correct buffer
+	frg::span<char> _buffer() {
```
Geertiebear, line 28: Both of these functions could and should be marked ``const``.

Geertiebear (changes requested):
In general, I'm not a fan of this code. It seems that the ``_allowResizing`` flag was used to cram two different classes into one implementation, which makes for a very confusing read of the code. In this case it would be appropriate to have two different classes, and possibly a common superclass that abstracts some of the common behaviour away. Right now (what should be) virtual calls are emulated using the ``_allowResizing`` flag.

In summary, please add missing test cases, since there were several untested cases I've noticed. And separate the two different implementations using inheritance, this will clean up the implementation and hopefully reduce the likelihood of bugs.

tests/posix/fmemopen.c
Geertiebear: Yeah that's fair. I think we don't have to test for this.

Geertiebear (commented):
I feel like ``io_seek()`` and ``io_read()`` can be fully abstracted by just making ``_buffer_size()`` a protected virtual function in the base class. ``io_write`` is unfortunately a little bit trickier it seems since there is a special case for ``fmemopen()``.

options/posix/include/mlibc/posix-file-io.hpp
Geertiebear: You can still mark the function as ``const`` just not the return value

options/posix/generic/posix-file-io.cpp
```diff
@@ -30,58 +21,160 @@ int mem_file::determine_bufmode(buffer_mode *mode) {
...
-void mem_file::_update_ptrs() {
+void memstream_mem_file::_update_ptrs() {
 	*_bufloc = _buf.data();
 	*_sizeloc = _buf.size() - 1;
 }
 
+fmemopen_mem_file::fmemopen_mem_file(void *in_buf, size_t size, int flags, void (*do_dispose)(abstract_file *))
+: mem_file{flags, do_dispose}, _inBuffer{in_buf}, _inBufferSize{size} {
+	if(!_inBuffer) {
+		_inBuffer = getAllocator().allocate(size);
```
Geertiebear, line 100: Do we properly handle the case when ``size`` is 0? On glibc this will cause the first read to return EOF, but im not sure that happens here (don't we just crash?) This behavior should obviously be tested ofc.

no92: We correctly implement it, although we could also return EINVAL and still be compliant with POSIX. I added a test case.

no92: Fun side note: GHA's glibc actually returns EINVAL with size 0.

PR #877, opened 2023-06-10
`sysconf` changes

no92:


options/posix/generic/unistd-stubs.cpp
```diff
@@ -677,49 +677,65 @@ void sync(void) {
...
+			return 256;
 		case _SC_PHYS_PAGES:
-			// TODO: actually return a proper value for _SC_PHYS_PAGES
-			mlibc::infoLogger() << "\e[31mmlibc: sysconf(_SC_PHYS_PAGES) returns arbitrary value 1024\e[39m" << frg::endlog;
+			mlibc::infoLogger() << "\e[31mmlibc: sysconf(_SC_PHYS_PAGES) returns fallback value 1024\e[39m" << frg::endlog;
 			return 1024;
 		case _SC_NPROCESSORS_ONLN:
-			// TODO: actually return a proper value for _SC_NPROCESSORS_ONLN
-			mlibc::infoLogger() << "\e[31mmlibc: sysconf(_SC_NPROCESSORS_ONLN) unconditionally returns 1\e[39m" << frg::endlog;
+			mlibc::infoLogger() << "\e[31mmlibc: sysconf(_SC_NPROCESSORS_ONLN) returns fallback value 1\e[39m" << frg::endlog;
```
ElectrodeYT, line 710: Realistically, this (and others) probably need some kind of sysdep callback; although _SC_NPROCESSORS_ONLN could probably be fetched by get_nprocs(), if we implement that function.

sysdeps/linux/generic/sysdeps.cpp
```diff
@@ -1349,6 +1351,54 @@ int sys_fstatvfs(int fd, struct statvfs *out) {
...
+			if(int e = sys_getrlimit(RLIMIT_NOFILE, &ru); e) {
+				return e;
+			}
+			*ret = (ru.rlim_cur == RLIM_INFINITY) ? -1 : ru.rlim_cur;
+			break;
+		}
+		case _SC_NPROCESSORS_ONLN: {
+			cpu_set_t set;
+			CPU_ZERO(&set);
+			if(int e = sys_getaffinity(0, sizeof(set), &set); e) {
```
ElectrodeYT, line 1367: Why do we do it like this, though? This looks a lot like it could be in unistd-stubs, not in the Linux sysdeps.

ElectrodeYT (changes requested):

sysdeps/linux/generic/sysdeps.cpp
no92: Not every sysdep might want to implement this, as this is a Linux-specific function.

options/posix/generic/unistd-stubs.cpp
Geertiebear: I mean that's already added in this PR. The warning messages are a good enough solution.

Geertiebear:
This LGTM, thanks!

qookei (approved):

PR #878, opened 2023-06-19
define _GNU_SOURCE in posix_string.cpp

moodyhunter:
...otherwise `memrchr` will be mangled as a C++ function (_Z7memrchrPKvim)

### DIFF
--- a/options/posix/generic/posix_string.cpp
+++ b/options/posix/generic/posix_string.cpp
@@ -1,3 +1,7 @@
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
+#endif
+
 #include <bits/ensure.h>
 #include <stdlib.h>
 #include <string.h>


no92:
I don't seem to be able to reproduce this locally.

moodyhunter:
It's neither defined in any header file nor the compiler cmdline, idk where else a macro can be defined :(

options/posix/generic/posix_string.cpp
```diff
@@ -1,3 +1,5 @@
+#define _GNU_SOURCE
```
no92, line 1: Wrapping this in an `#ifndef _GNU_SOURCE` might solve the problem of compilers predefining this macro.

moodyhunter: Good idea, will do

no92 (approved):
LGTM now.

Geertiebear:
If the commits could be squashed then LGTM!

moodyhunter:
I have just squashed the two commits :)

PR #884, opened 2023-07-28
Changes for managarm cancelable syscalls

Geertiebear:
Since the changes are incomplete and a little unstable, marking it as draft right now.

### DIFF
--- a/sysdeps/managarm/generic/file.cpp
+++ b/sysdeps/managarm/generic/file.cpp
@@ -1748,6 +1748,10 @@ int sys_read(int fd, void *data, size_t max_size, ssize_t *bytes_read) {
 	if (!handle)
 		return EBADF;
 
+	HelHandle cancel_handle;
+	HEL_CHECK(helCreateOneshotEvent(&cancel_handle));
+	helix::UniqueDescriptor cancel_event{cancel_handle};
+
 	managarm::fs::CntRequest<MemoryAllocator> req(getSysdepsAllocator());
 	req.set_req_type(managarm::fs::CntReqType::READ);
 	req.set_fd(fd);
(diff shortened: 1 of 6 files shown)


Geertiebear:
Yeah there are quite some debug logs left currently :sweat_smile:

sysdeps/managarm/generic/file.cpp
```diff
@@ -1,3 +1,4 @@
+#include "mlibc/debug.hpp"
```
ElectrodeYT, line 1: Is this LSP? In any case, this should be replaced with `<>`

ElectrodeYT (approved):
LGTM

I do have a question, since this is not instantly obvious to me: would this code work without the changes in managarm, or are we looking at circular dependencies here? To my knowledge, merging in the managarm side first would result in some weird behavior.

sysdeps/managarm/include/mlibc/posix-pipe.hpp
```diff
@@ -278,12 +298,39 @@ auto exchangeMsgsSync(HelHandle descriptor, Args &&...args) {
...
+	setCurrentRequestEvent(event);
+	HEL_CHECK(
+	    helSubmitAsync(descriptor, actions.data(), actions.size(), globalQueue.getQueue(), 0, 0)
+	);
+
+	auto element = globalQueue.dequeueSingle(false);
+	if (!element) {
+		element = globalQueue.dequeueSingle();
+		__ensure(element);
+	}
```
avdgrinten, line 325: This is the only call where we pass `false` to `dequeueSingle()` (AFAICT) and we immediately call into `dequeueSingle()` again. So what is the difference compared to just doing `dequeueSingle(true)` here?

sysdeps/managarm/generic/fork-exec.cpp
```diff
@@ -148,7 +148,10 @@ int sys_sleep(time_t *secs, long *nanos) {
 	));
 
 	auto element = globalQueue.dequeueSingle();
-	auto result = parseSimple(element);
+	if (!element) {
+		return EINTR;
+	}
+	auto result = parseSimple(*element);
```
avdgrinten, line 169: We have this change for sleep but AFAICT `sleep()` is still not cancelable since we don't integrate it with the cancellation event. We could drop support for `sleep()` in the initial PR.

sysdeps/managarm/include/mlibc/posix-pipe.hpp
avdgrinten: Note we use protocol errors to signify cancellation (e.g., in `read()` you check for `managarm::fs::Errors::INTERRUPTED`), so maybe the `ignoreCancel = false` case is not necessary (and we can always re-try if `helFutexWait` is interrupted)?

sysdeps/managarm/generic/file.cpp
```diff
@@ -1628,12 +1629,14 @@ int sys_mknodat(int dirfd, const char *path, int mode, int dev) {
 }
 
 int sys_read(int fd, void *data, size_t max_size, ssize_t *bytes_read) {
-	SignalGuard sguard;
```
avdgrinten, line 1631: Why do we need to do this?

This prevents a signal to be raised, but it should not prevent POSIX from interrupting the thread. I think we still want to keep this.

sysdeps/managarm/include/mlibc/posix-pipe.hpp
Geertiebear: Iirc I did this because some threads were getting some spurious interrupts, but now that I'm thinking about it, that re-entering dequeueSingle should never be an issue in that case.

sysdeps/managarm/generic/fork-exec.cpp
no92: Given the changes done to the return type of `deqeueSingle`, handling the error seems preferable to `__ensure(element)`.

PR #889, opened 2023-08-03
socket: implement shutdown(2)

Andy-Python-Programmer:


options/posix/generic/sys-socket-stubs.cpp
```diff
@@ -187,9 +187,9 @@ int setsockopt(int fd, int layer, int number,
 	return mlibc::sys_setsockopt(fd, layer, number, buffer, size);
 }
 
-int shutdown(int, int) {
-	mlibc::infoLogger() << "mlibc: shutdown() is a no-op!" << frg::endlog;
-	return 0;
+int shutdown(int sockfd, int how) {
+	MLIBC_CHECK_OR_ENOSYS(mlibc::sys_shutdown, -1);
+	return mlibc::sys_shutdown(sockfd, how);
```
no92, line 192: This is incorrect; to quote the man page,

> On success, zero is returned.  On error, -1 is returned, and errno is set to indicate the error.

You can copy the implementation for this from e.g. starting at line 172.

no92: Also, please change this to use the `auto sysdep = MLIBC_CHECK_OR_ENOSYS(...)` idiom, as done in e.g. `net-if-stubs`.

no92 (changes requested):

options/posix/generic/sys-socket-stubs.cpp
Andy-Python-Programmer: Fixed.

ikbenlike (approved):
LGTM

Dennisbonke (commented):
LGTM but could you also add the Linux sysdep for it while you're at it? Always nice to improve the Linux port and it should be just a few lines.

PR #891, opened 2023-08-05
FUSE (WIP)

ikbenlike:
This PR is the mlibc counterpart of [managarm#533](https://github.com/managarm/managarm/pull/533). None of the changes in this PR are necessary for FUSE functionality, however they aid in porting libfuse to managarm.


tests/meson.build
```diff
@@ -95,6 +95,8 @@ all_test_cases = [
 	'glibc/error_expect_fail',
 	'glibc/error',
 	'glibc/error_at_line',
+    'glibc/strerrorname',
+    'glibc/strerrordesc',
```
qookei, line 99: Wrong indentation

options/ansi/generic/string-stubs.cpp
```diff
@@ -474,6 +487,130 @@ void *mempcpy(void *dest, const void *src, size_t len) {
 }
 
 // GNU extensions.
+const char *strerrorname_np(int e) {
+	const char *s;
+	switch(e) {
+	case EAGAIN: s = "EAGAIN"; break;
```
qookei, line 493: I'd prefer this to be done with a macro to avoid typos (there are several in the strings in this function), like so:
```cpp
	switch(e){
#define X(x) case x: s = #x; break;
	X(EAGAIN);
	X(EACCES);
	// ...
#undef X
	default:
	// ...
	}
```

sysdeps/managarm/generic/mount.cpp
```diff
@@ -12,13 +12,16 @@
...
-		const char *fstype, unsigned long, const void *) {
+		const char *fstype, unsigned long, const void *data) {
 	SignalGuard sguard;
 
 	managarm::posix::MountRequest<MemoryAllocator> req(getSysdepsAllocator());
 	req.set_path(frg::string<MemoryAllocator>(getSysdepsAllocator(), source));
 	req.set_target_path(frg::string<MemoryAllocator>(getSysdepsAllocator(), target));
 	req.set_fs_type(frg::string<MemoryAllocator>(getSysdepsAllocator(), fstype));
+	if(data){
+		req.set_arguments(frg::string<MemoryAllocator>(getSysdepsAllocator(), (char*)data));
```
qookei, line 23: Is the extra data always a string? (Is it just the mount options as specified by `-o` to `mount(8)`?)

qookei (changes requested):

sysdeps/managarm/generic/mount.cpp
ikbenlike: I'm not aware of any cases in which this data isn't a string, but I don't think it's is required to be a string. This is mainly a temporary measure until I can think of a better way to approach this.

tests/meson.build
ikbenlike: Fixed.

options/ansi/generic/string-stubs.cpp
ikbenlike: Fixed.

sysdeps/managarm/generic/mount.cpp
ikbenlike: `mount(2)` on my system says:

> The  data argument is interpreted by the different filesystems. Typically it is a string of comma-separated options understood by this filesystem.  See mount(8) for details of the options available for each filesystem type.

I think that, practically speaking, we would be fine by just interpreting this as a string in almost any occasion. However, I don't think it's technically correct.

qookei: That seems fine to me then. Since the man page directs you to `mount(8)` then it'd be difficult to specify binary data (maybe check `mount` source code?)

ikbenlike: I might check the source later, but it's not very high-priority for me. I was thinking about implementing some sort of fs-specific argument decoding dispatch mechanism, but I'm not sure how I'd go about doing this and it'd probably require the sysdep to be aware of every possible filesystem - which would be impractical at best

options/linux/include/sys/mount.h
```diff
@@ -40,6 +40,11 @@ extern "C" {
 
 #ifndef __MLIBC_ABI_ONLY
 
+#define MNT_FORCE 1
```
Dennisbonke, line 43: Double define

options/linux/include/sys/mount.h
```diff
@@ -40,6 +40,11 @@ extern "C" {
...
+#define MNT_DETACH (1<<1)
```
Dennisbonke, line 44: Styling is wrong (see the constants defined earlier in the file.

options/linux/include/sys/mount.h
```diff
@@ -40,6 +40,11 @@ extern "C" {
...
+#define MNT_EXPIRE (1<<2)
```
Dennisbonke, line 45: We should move these defines outside the MLIBC_ABI_ONLY define. Also please make sure these new defines match Linux. Better yet, we should really move these to an abi-bit header, as I'm pretty sure that at least some of the constants in this file don't match Linux.

options/ansi/generic/string-stubs.cpp
```diff
@@ -474,6 +491,132 @@ void *mempcpy(void *dest, const void *src, size_t len) {
 }
 
 // GNU extensions.
+const char *strerrorname_np(int e) {
+	const char *s;
+#define X(x) case x: s = #x; break;
+	switch(e) {
+	X(EAGAIN)
+	X(EACCES)
```
Dennisbonke, line 499: Shouldn't this be indented one more?

Dennisbonke (changes requested):
Few items to fix

options/ansi/generic/string-stubs.cpp
ikbenlike: The case statements I replaced were on the same indentation level - but I can indent it a level more of desired

sysdeps/managarm/generic/mount.cpp
Geertiebear: I'm just throwing around ideas here, but isn't it also possible to have a ``pushMemory`` action similarly to how we have a ``pushDescriptor`` action that pushes an arbitrary piece of memory across process boundaries? Or maybe a mechanism which turns a piece of memory into a descriptor? IIRC that would also help at other points in the managarm sysdeps.

PR #892, opened 2023-08-09
Wine changes

Qwinci:
Changes required for wine to compile.


no92 (approved):
LGTM

options/rtdl/generic/linker.cpp
```diff
@@ -683,7 +683,7 @@ void ObjectRepository::_parseDynamic(SharedObject *object) {
 		case DT_TLSDESC_PLT: case DT_TLSDESC_GOT:
 			break;
 		default:
-			mlibc::panicLogger() << "Unexpected dynamic entry "
+			mlibc::infoLogger() << "Unexpected dynamic entry "
```
Geertiebear, line 686: I'd rather not change this, we should just add the missing dynamic entry into the code and add logging to say we don't handle it properly.

Geertiebear (changes requested):

options/rtdl/generic/linker.cpp
Qwinci: They seem to be values in the os-specific area that are specific to wine .so dlls, so maybe it should just log the unknown os specific entries but otherwise ignore them?

options/rtdl/generic/linker.cpp
```diff
@@ -683,8 +683,10 @@ void ObjectRepository::_parseDynamic(SharedObject *object) {
 		case DT_TLSDESC_PLT: case DT_TLSDESC_GOT:
 			break;
 		default:
```
Geertiebear, line 685: If you could please add a comment here as well explaining why we have this if statement here, that would be great. I just know someone is going to take a look at this code in the future and be extremely confused, and I'd like to avoid that.

Geertiebear (approved):

PR #893, opened 2023-08-09
options/ansi: make UB truly undefined in strxfrm

no92:


Geertiebear (commented):
LGTM, this follows the man page now. Are there tests for this function yet? Would be nice to add some.

no92:
Added a test.

PR #894, opened 2023-08-09
options/ansi: prevent multiplication overflows in calloc

### LINKED ISSUE #790: calloc() does not check for overflow
Currently, `calloc()` is defined in `stdlib-shims.cpp` as:
```c
void *calloc(size_t count, size_t size) {
	// TODO: this could be done more efficient if the OS gives us already zero'd pages
	void *ptr = malloc(count * size);
	if(!ptr)
		return nullptr;
	memset(ptr, 0, count * size);
	return ptr;
}
```

If `count` × `size` is greater than `SIZE_MAX`, then this will allocate a block of memory smaller than `count` × `size` bytes. This is contrary to what ISO C requires:

> The `calloc` function allocates space for an array of `nmemb` objects, each of whose size is `size`. The space is initialized to all bits zero.

The function should instead return a null pointer.

no92:
This fixes #790.

### DIFF
--- a/options/ansi/generic/stdlib-stubs.cpp
+++ b/options/ansi/generic/stdlib-stubs.cpp
@@ -138,6 +138,13 @@ void *aligned_alloc(size_t alignment, size_t size) {
 
 }
 void *calloc(size_t count, size_t size) {
+	// we want to ensure that count*size > SIZE_MAX doesn't happen
+	// to prevent overflowing, we divide both sides of the inequality by size and check with that
+	if(size && count > (SIZE_MAX / size)) {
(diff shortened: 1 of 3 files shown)


ikbenlike (approved):
LGTM

Geertiebear (commented):
This might be really nitpicky, but could you add a test for this? This is exactly the kind of stuff that is easy to test for!

no92:
Adding this test wasn't quite as easy as believed, as asan catches the overflow; I had to disable asan on host-libc for this test as even `__attribute__((no_sanitize("address")))` didn't disable this check.

PR #896, opened 2023-08-11
Latest misc LFS changes

no92:


options/rtdl/generic/linker.cpp
```diff
@@ -22,6 +22,12 @@ constexpr bool logBaseAddresses = false;
 constexpr bool logRpath = false;
 constexpr bool eagerBinding = true;
 
+namespace {
+
+bool rpathWarned = false;
```
Geertiebear, line 27: Is this change really necessary?

Geertiebear:
What led you to making the ``_save_pos()`` change? It LGTM I'm just curious.

options/rtdl/generic/linker.cpp
no92: It massively spams the log when running some applications, so warning once seemed sufficient.

no92:
I don't recall the details as it's been a few weeks but IIRC it was seeking on sysfs or proc files in managarm.

options/rtdl/generic/linker.cpp
Geertiebear: Alright makes sense. I would prefer if it wasn't a global variable though, putting it as a static variable in the case scope should work. Also put a comment explaining that this is both temporary and to reduce log-spam for some applications.

Geertiebear (approved):
LGTM

PR #897, opened 2023-08-12
sysdeps: Fix sp assignment in sys_prepare_stack

Qwinci:
Change sp assignment in sys_prepare_stack in case `*stack` is not null to `*stack + *stack_size`, because *stack points to the bottom of the stack and not top. As it is currently the writes later down in this function would be out of bounds in that case.


Geertiebear (approved):
LGTM, thanks for fixing this mistake! (woops lol)

Geertiebear:
Could you please squash the sysdep commits into the actual commit that changed the sysdep?

Qwinci:
> Could you please squash the sysdep commits into the actual commit that changed the sysdep?

Done. There is the managarm sysdep fix commit that was made before the signature change, should it be kept or do I squash it too?

Geertiebear:
That can be kept, it has nothing to do with the change in the sysdep signature.

PR #902, opened 2023-08-18
sysdeps/managarm: Make epoll_create accept O_CLOEXEC too

Dennisbonke:
Node.js is using O_CLOEXEC, and on Linux that's the same, but on Managarm it is not.


Geertiebear (approved):
LGTM but this should be removed once the ABI is made constant, or it's fixed upstream.

PR #905, opened 2023-08-18
options/elf: Match ELF64_ST_* to their ELF32_ST_* counterparts

Dennisbonke:
Potential abi-break.


Geertiebear:
Luckily this isn't an ABI break, so this LGTM.

PR #908, opened 2023-08-22
Some Linux kernel build fixes

Dennisbonke:
Mostly elf.h definitions.

### DIFF
--- a/options/elf/include/elf.h
+++ b/options/elf/include/elf.h
@@ -19,8 +19,9 @@ extern "C" {
 #define SHF_WRITE 1
 #define SHF_ALLOC 2
 #define SHF_EXECINSTR 4
-#define SHF_STRINGS 16
-#define SHF_TLS 512
+#define SHF_STRINGS 32
+#define SHF_INFO_LINK 64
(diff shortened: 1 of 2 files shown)


options/elf/include/elf.h
```diff
@@ -20,6 +20,7 @@ extern "C" {
 #define SHF_ALLOC 2
 #define SHF_EXECINSTR 4
 #define SHF_STRINGS 16
+#define SHF_INFO_LINK 32
```
no92, line 23: AFAICT this should be defined to 64, not 32.

no92: Oh and SHF_STRINGS should be 32, not 16.

no92: And SHF_TLS should be 1024.

no92 (changes requested):

options/elf/include/elf.h
Dennisbonke: Addressed as requested.

no92 (approved):
LGTM

Geertiebear (approved):
LGTM.

PR #909, opened 2023-08-22
ABI-only CI

ikbenlike:
Add a CI workflow to check if including files with the `__MLIBC_ABI_ONLY` macro defined correctly includes only the ABI.


Geertiebear:
This currently fails the CI

ikbenlike:
> This currently fails the CI

Yes, however, I am unsure if this is a "correct" failure or if there is a fault in my methodology.

Geertiebear:
That seems like something to figure out :wink:

PR #911, opened 2023-08-24
abis/linux: Fix the signage of uid_t and gid_t

Dennisbonke:


no92 (approved):
LGTM

Geertiebear (approved):
LGTM, this is officially an ABI break, so should we maybe merge it to abi-break instead?

PR #913, opened 2023-08-25
options/glibc: Add utmpx paths

Dennisbonke:
options/linux: Define more utmpx stuff

### DIFF
--- a/options/linux/generic/utmpx.cpp
+++ b/options/linux/generic/utmpx.cpp
@@ -2,31 +2,44 @@
 #include <stddef.h>
 #include <errno.h>
 #include <utmpx.h>
+#include <mlibc/debug.hpp>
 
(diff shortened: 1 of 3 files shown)


no92 (approved):
LGTM

options/linux/generic/utmpx.cpp
```diff
@@ -30,3 +30,8 @@ int utmpxname(const char *) {
 	errno = ENOSYS;
 	return -1;
 }
+
+struct utmpx *getutxid(const struct utmpx *) {
+	// return NULL as musl does
```
Geertiebear, line 35: Maybe add an info log here that we haven't implemented this, as it is a genuinly useful interface.

Geertiebear (changes requested):

Geertiebear (approved):

PR #915, opened 2023-08-29
Architecture-generic rtdl rework

no92:
This also brings support for x86.


options/elf/include/elf.h
```diff
@@ -236,19 +236,43 @@ enum {
...
+	R_AARCH64_TLS_DTPREL64 = 1028,
+	R_AARCH64_TLS_DTPMOD64 = 1029,
+	R_AARCH64_TLS_TPREL64 = 1030,
+	R_AARCH64_TLSDESC = 1031,
+	R_AARCH64_IRELATIVE = 1032,
 };
 
+#define R_AARCH64_TLS_DTPREL R_AARCH64_TLS_DTPREL64
+#define R_AARCH64_TLS_DTPMOD R_AARCH64_TLS_DTPMOD64
+#define R_AARCH64_TLS_TPREL R_AARCH64_TLS_TPREL64
```
ikbenlike, line 274: Why do these defines need to exist?

options/rtdl/aarch64/elf.hpp
```diff
@@ -0,0 +1,35 @@
...
+using elf_ehdr = Elf64_Ehdr;
+using elf_phdr = Elf64_Phdr;
+using elf_dyn = Elf64_Dyn;
+using elf_rel = Elf64_Rel;
+using elf_rela = Elf64_Rela;
+using elf_sym = Elf64_Sym;
+using elf_addr = Elf64_Addr;
+
+using elf_info = Elf64_Xword;
+using elf_addend = Elf64_Sxword;
```
ikbenlike, line 16: Would making the code generic in this manner not prevent a 64-bit loader from loading 32-bit executables? Or is this not something we support at all?

options/rtdl/generic/linker.cpp
```diff
@@ -669,7 +680,11 @@ void ObjectRepository::_parseDynamic(SharedObject *object) {
 			object->preInitArraySize = dynamic->d_un.d_val;
 			break;
 		case DT_DEBUG:
+#if ELF_CLASS == 32
+			dynamic->d_un.d_val = reinterpret_cast<Elf32_Word>(&globalDebugInterface);
+#elif ELF_CLASS == 64
 			dynamic->d_un.d_val = reinterpret_cast<Elf64_Xword>(&globalDebugInterface);
+#endif
```
ikbenlike, line 687: Could this not use the `elf_` types? Or does this not work with the `using` types?

options/rtdl/generic/linker.cpp
```diff
@@ -1680,48 +1623,62 @@ void Loader::_processLazyRelocations(SharedObject *object) {
...
+	size_t rel_size = (*object->lazyExplicitAddend) ? sizeof(elf_rela) : sizeof(elf_rel);
+
+	for(size_t offset = 0; offset < object->lazyTableSize; offset += rel_size) {
+#if ELF_CLASS == 32
+		Elf32_Word type;
+		Elf32_Word symbol_index;
+#elif ELF_CLASS == 64
+		Elf64_Xword type;
+		Elf64_Xword symbol_index;
+#endif
```
ikbenlike, line 1636: Same remark about the `elf_` types here.

ikbenlike (changes requested):

options/elf/include/elf.h
no92: They exist on Linux and some programs seem to expect these to be in place.

options/rtdl/aarch64/elf.hpp
no92: This is equivalent to the macro magic in place with other libc implementations. For loading 32-bit executables on e.g. x86_64, I'd assume the 32-bit ld.so would be requested.

options/rtdl/generic/linker.cpp
no92: I'm not sure how this exactly works, so I didn't define a type for this - it would only have this one use anyways.

ikbenlike: Even if it would only have this use, I feel like it would make the code more consistent. I'm however not familiar enough with C++ to say if it would work or not.

ElectrodeYT (approved):
LGTM

options/rtdl/generic/linker.cpp
```diff
@@ -1427,217 +1436,131 @@ void Loader::_scheduleInit(SharedObject *object) {
...
+		if constexpr (tlsAboveTp) {
+			off += tls_offset - sizeof(Tcb);
+		} else {
+			off += tls_offset;
 		}
+
+		rel.relocate(off);
 	} break;
-	case R_RISCV_NONE:
-		// Work around https://sourceware.org/bugzilla/show_bug.cgi?id=24673.
```
qookei, line 1626: What about this workaround?

options/rtdl/aarch64/elf.hpp
```diff
@@ -0,0 +1,35 @@
...
+#define R_RELATIVE R_AARCH64_RELATIVE
+#define R_IRELATIVE R_AARCH64_IRELATIVE
+// #define R_OFFSET
+#define R_COPY R_AARCH64_COPY
+#define R_TLS_DTPMOD R_AARCH64_TLS_DTPMOD
+#define R_TLS_DTPREL R_AARCH64_TLS_DTPREL
+#define R_TLS_TPREL R_AARCH64_TLS_TPREL
+#define R_TLSDESC R_AARCH64_TLSDESC
+
+#define TP_TCB_OFFSET (16)
```
qookei, line 35: This define (and equivalents for other architectures) is unused?

options/rtdl/generic/linker.cpp
no92: This is now handled by `R_NONE`.

qookei: Oh must've missed it, my bad. Perhaps keep the comment saying that it works around a binutils bug on RV?

no92: The issue is still linked where it's relevant (TEXTREL handling), so I removed this here as ignoring R_NONE-type relocations is correct behavior anyways and not really relevant to handling the binutils bug outside of that.

options/rtdl/aarch64/elf.hpp
no92: Good catch, this was supposed to replace the magic number in `tcb.hpp`, which now it does.

options/rtdl/generic/linker.cpp
64: Is it actually correct to ignore none-type relocs? Feels like we might want to at least warn on those

64 (approved):
LGTM, slight preference to use C++ abstractions rather than macros where possible, but not a big deal

options/rtdl/generic/linker.cpp
no92: Every libc I checked just ignores them and the ABI specs all just say "perform no relocation", no more, no less.

64: It probably says that for forward compatability reasons, but IMHO it could be good to warn because it could indicate a problem with the toolchain or linker

Dennisbonke (approved):
LGTM

Geertiebear:
mfw i had a bunch of comments in the making

Dennisbonke:
> mfw i had a bunch of comments in the making

Please do post them. We might still want to fix them.

PR #916, opened 2023-09-24
options/rtdl: return previous behavior of not ensuring symbol resolution

no92:
Previously, we didn't assert this at least on some architectures, including x86_64. This apparently breaks `libmbus.so`.


Dennisbonke (approved):
LGTM

PR #921, opened 2023-10-07
ansi: add EIEIO to strerror()

ElectrodeYT:


no92 (approved):
LGTM

options/ansi/generic/string-stubs.cpp
```diff
@@ -455,6 +455,11 @@ char *strerror(int e) {
 	case ENOTUNIQ: s = "Name not unique on network (ENOTUNIQ)"; break;
 	case ERESTART: s = "Interrupted system call should be restarted (ERESTART)"; break;
 	case EUSERS: s = "Too many users (EUSERS)"; break;
+
+	#ifdef EIEIO
```
Dennisbonke, line 459: Formatting issue, the #ifdef should not be indented here. Same for the #endif.

Dennisbonke (changes requested):
Small formatting issue

options/ansi/generic/string-stubs.cpp
ElectrodeYT: Done

Dennisbonke (approved):
LGTM

PR #925, opened 2023-10-08
options/linux: set window size in openpty

ElectrodeYT:


options/linux/generic/pty-stubs.cpp
```diff
@@ -38,6 +34,13 @@ int openpty(int *mfd, int *sfd, char *name, const struct termios *ios, const str
 		goto fail;
 	}
 
+	if(win) {
+		if(int ioctl_e, e = mlibc::sys_ioctl(ptmx_fd, TIOCSWINSZ, (void*)win, &ioctl_e); e || ioctl_e) {
+			errno = e;
+			goto fail;
```
no92, line 40: I'm not sure whether closing the ptmx is the right move here.

options/linux/generic/pty-stubs.cpp
```diff
@@ -15,10 +15,6 @@
 int openpty(int *mfd, int *sfd, char *name, const struct termios *ios, const struct winsize *win) {
 	__ensure(!ios);
```
no92, line 16: Might be nice to also implement this. It should probably be placed alongside the TIOCSWINSZ.
```cpp
if(ios)
    tcsetattr(ptmx_fd, TCSAFLUSH, ios);
```

qookei (approved):

PR #930, opened 2023-10-16
options/ansi: threads.h should define thread_local for C

netbsduser:
C++ defines `thread_local` as a keyword but in ISO C it isn't and must be a preprocessor definition. (See [ISO C11 draft spec](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf), section 6.2.4 - `_Thread_local` is listed as a storage-class specifier; section 7.26.1 - `thread_local` is listed as a macro defined by `<threads.h>` which expands to `_Thread_local`).

The lack of this breaks the build of e.g. [ObjFW](https://github.com/ObjFW/ObjFW).


Geertiebear (approved):
LGTM.

Dennisbonke (approved):
LGTM

PR #932, opened 2023-10-22
termios fixes

no92:


Dennisbonke (changes requested):
Haven't checked the values but I support the effort. However, blocking on a bootstrap PR to bump the world as this breaks Linux abi.

Dennisbonke (approved):
LGTM

PR #933, opened 2023-10-23
options/posix: Fix function prototype of pthread_exit

Dennisbonke:
It's currently returning an int, and spec says it should be noreturn void. Draft as I'm not quite sure if this is now correct.

### DIFF
--- a/options/posix/generic/pthread-stubs.cpp
+++ b/options/posix/generic/pthread-stubs.cpp
@@ -354,11 +354,18 @@ namespace {
 	FutexLock key_mutex_;
 }
 
-int pthread_exit(void *ret_val) {
+namespace mlibc {
+	__attribute__ ((__noreturn__)) void do_exit() {
+		sys_thread_exit();
+		__builtin_unreachable();
+	}
+}
+
+__attribute__ ((__noreturn__)) void pthread_exit(void *ret_val) {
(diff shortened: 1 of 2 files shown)


options/posix/generic/pthread-stubs.cpp
```diff
@@ -354,11 +354,11 @@ namespace {
 	FutexLock key_mutex_;
 }
 
-int pthread_exit(void *ret_val) {
+[[noreturn]] void pthread_exit(void *ret_val) {
 	auto self = mlibc::get_current_tcb();
 
 	if (__atomic_load_n(&self->cancelBits, __ATOMIC_RELAXED) & tcbExitingBit)
-		return 0; // We are already exiting
+		goto exit;
```
Geertiebear, line 361: Should have mentioned in this discord, but we can avoid a goto here pretty easily. Make a static ``mlibc::do_exit`` function and call it from here and at the bottom of this function.

Geertiebear (dismissed):

options/posix/generic/pthread-stubs.cpp
Dennisbonke: Fixed, should be good to go now.

no92 (approved):
LGTM

qookei (approved):

PR #936, opened 2023-10-23
Java additions

Qwinci:


abis/linux/ioctls.h
```diff
@@ -9,5 +9,9 @@
 #define SIOCGIFINDEX 0x8933
 #define SIOCATMARK 0x8905
 #define SIOCGIFHWADDR 0x8927
+#define SIOCGIFBRDADDR 0x8919
+#define SIOCGIFNETMASK 0x891B
+
+#define FIONREAD 0x541B
```
Dennisbonke, line 15: This is already defined in `options/glibc/include/sys/ioctl.h`

abis/linux/input.h
```diff
@@ -0,0 +1,14 @@
+#ifndef _ABIBITS_INPUT_sH
+#define _ABIBITS_INPUT_sH
+
+#define BTN_MOUSE 0x110
```
Dennisbonke, line 4: We should not ship this, this is part of `input-event-codes.h` of the linux kernel headers. Programs should grab the values from there. Mlibc code in the linux option can make use of those headers too.

options/linux/include/linux/input.h
```diff
@@ -0,0 +1,14 @@
+#ifndef _LINUX_INPUT_H
+#define _LINUX_INPUT_H
+
+#define BTN_MOUSE 0x110
```
Dennisbonke, line 4: Same here, we should not ship these headers.

options/posix/generic/search.cpp
```diff
@@ -149,3 +149,22 @@ void *lfind(const void *key, const void *base, size_t *nelp,
 	__ensure(!"Not implemented");
 	__builtin_unreachable();
 }
+
+#if __MLIBC_GLIBC_OPTION
```
Dennisbonke, line 153: We should move this to a file under the glibc option.

options/posix/include/search.h
```diff
@@ -15,6 +15,25 @@ typedef enum {
   leaf
 } VISIT;
 
+typedef enum {
```
Dennisbonke, line 18: Same for this here, we can include the glibc header in here if the glibc option is enabled.

options/linux/meson.build
```diff
@@ -87,6 +87,7 @@ if not no_headers
 	)
 	install_headers(
 		'include/linux/libc-compat.h',
+		'include/linux/input.h'
```
Dennisbonke, line 90: While not needed (see other comments), you are missing a comma here, failing CI. Just a note for future PRs.

Dennisbonke (changes requested):
A quick look found some issues. I haven't gone in depth on the structs (but they look correct) or on the `sem_trywait()` implementation yet.

Qwinci:
I changed those, installing the partial linux/input.h header also created build issues.

options/posix/generic/semaphore-stubs.cpp
```diff
@@ -98,7 +98,20 @@ int sem_unlink(const char *) {
 	__builtin_unreachable();
 }
 
-int sem_trywait(sem_t *) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+int sem_trywait(sem_t *sem) {
```
qookei, line 101: This impl looks fine, but I feel like it could be reordered like so:
```cpp
while (true) {
    auto state = load(ctr);

    if (!(state & waiters_mask)) return EAGAIN;

    auto desired = state - 1;
    if (cmpxchg(ctr, state, desired)) return 0;
}

File: options/glibc/generic/search.cpp [hunk 8]
```diff
@@ -0,0 +1,17 @@
...
+
+int hcreate_r(size_t, struct hsearch_data *) {
+	__ensure(!"Not implemented");
+	__builtin_unreachable();
+}
+
+void hdestroy_r(struct hsearch_data *) {
+	__ensure(!"Not implemented");
+	__builtin_unreachable();
+}
```
qookei, line 17: Why not actually implement these? Also perhaps the actual implementations should be shared (in `options/internal`?), since AFAICS the non-reentrant versions are just wrappers around these, except that they pass a pointer to some global instance of `struct hsearch_data`.

I propose making out `struct hsearch_data` only contain one pointer to an internal heap-allocated object, which would contain perhaps a `frg::hash_map`? That way we can also extend the struct as needed without ABI breakage.

Qwinci: Yeah thats a good point. Ill do that.

no92 (approved):
LGTM

options/posix/include/unistd.h
```diff
@@ -192,6 +192,8 @@ extern "C" {
 #define _SC_RTSIG_MAX 79
 #define _SC_SIGQUEUE_MAX 80
 
+#define _SC_IOV_MAX 60
```
qookei, line 195: This overlaps our `_SC_CPUTIME`, just assign the next available value (81)

options/glibc/include/netinet/icmp6.h
```diff
@@ -14,6 +16,29 @@ extern "C" {
 #define ND_OPT_RTR_ADV_INTERVAL 7
 #define ND_OPT_HOME_AGENT_INFO 8
 
+struct icmp6_hdr {
```
qookei, line 19: Why is this duplicated here? It's already in the POSIX header?

qookei: As in, can't we include the POSIX header to bring it in?

qookei (approved):
Looks good now. This should be fine to merge into master instead of abi-break, the detected break is only about renaming a struct.

PR #941, opened 2023-10-29
options/glibc: Add the hsearch_r family of functions

Dennisbonke:
Lifted from #936 and then implemented using musl. Needed for `procps`.


no92:
Already added by #1063.

PR #944, opened 2023-11-04
Fix assert noreturn on clang

Qwinci:
Atleast on clang [[noreturn]] is not the same as `__attribute__((__noreturn__))` used in the header and it fails to compile because of that.


Dennisbonke (approved):
LGTM

PR #950, opened 2023-11-19
misc: Remove include-fixed include path

Qwinci:
Adding the include-fixed include directory breaks the build if the directory actually has some system headers. For an example on Gentoo there is pthread.h there and when building mlibc it gets included instead of the mlibc pthread.h causing the build to fail.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -85,11 +85,6 @@ if not headers_only
 		error('could not find compiler-specific header directory')
 	endif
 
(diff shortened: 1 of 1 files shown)


no92 (approved):
LGTM as I fundamentally agree, however I would want to see whether we can get a full world rebuild with this and roll back if needed.

Originally, this seems to have been added by @avdgrinten, maybe he can enlighten us on why this was added to begin with.

PR #951, opened 2023-11-19
options/rtdl: Implement R_IRELATIVE relocation

Qwinci:


no92 (changes requested):
This would either need guarding by `#ifdef`s or something similar as this has quirks for aarch64 and RISC-V.

Qwinci:
> This would either need guarding by `#ifdef`s or something similar as this has quirks for aarch64 and RISC-V.

I converted the late relocation handling to use the Relocation type and moved R_IRELATIVE handling to there because they are supposed to be processed after all other relocations. Also added `#ifdef`s for x86_{x64} and aarch64 (though that implementation doesn't properly pass in the hwcaps as the first argument and always just passes a zero).

no92 (approved):
LGTM

Dennisbonke (approved):
LGTM

PR #952, opened 2023-11-20
sysdeps/managarm: remove CLOCK_MONOTONIC_RAW warning

ElectrodeYT:
helGetClock is a raw counter, so warning here makes no sense.


no92 (approved):
LGTM

PR #958, opened 2023-12-11
termios-stub.cpp: make sure _GNU_SOURCE is defined

moodyhunter:
need to get CBAUD, etc.


Geertiebear:
Not entirely sure if I understand the issue that this solves, could you elaborate?

moodyhunter:
Hi, a similar one is at https://github.com/managarm/mlibc/pull/878

A compiler may not define _GNU_SOURCE so we have to define it to get CBAUD defined.

no92 (approved):
LGTM

moodyhunter:
thanks :)

PR #961, opened 2023-12-21
options/rtdl: Fix alignment overhead computation for the TLS allocation

### LINKED ISSUE #960: Certain combinations of TLS segments lead to misaligned TLS pointers
When the sizes of TLS blocks add up to a number misaligned by 8 from 16, the TCB pointer ends up being similarily misaligned.
This leads to the TLS pointer for a library with a TLS alignment of 16 to be misaligned by 8, which causes a crash if the compiler generates code assuming the alignment is respected (like using aligned SIMD load instructions).

Example situation:
```
rtdl: TLS of libpixman-1.so.0 mapped to 0xfffffffffffffe80, size: 384, alignment: 16
rtdl: TLS of libjpeg.so.8 mapped to 0xfffffffffffffe78, size: 8, alignment: 4
rtdl: TLS of libc.so mapped to 0xfffffffffffffc70, size: 520, alignment: 8
rtdl: TLS of libGLdispatch.so.0 mapped to 0xfffffffffffffc68, size: 8, alignment: 8
rtdl: TLS of libstdc++.so.6 mapped to 0xfffffffffffffc48, size: 32, alignment: 8
...
rtdl: tcb allocated at 0x7fffff9bfbf8, size = 0x80
rtdl: tls allocated at 0x7fffff9bf800, size = 0x3f8
```

TCB ends up at `0x7fffff9bfbf8`, and as such accessing the TLS block for pixman yields `0x7fffff9bfbf8 + 0xfffffffffffffe80 = 0x7fffff9bfa78`, which is not aligned like the library requested.

qookei:
The alignment overhead is used to properly align the TCB, which on x86_64 needs to be aligned to the max TLS block alignment so that accessing blocks before it yields aligned addresses.

Fixes #960.

### DIFF
--- a/tests/rtdl/tls_align/test.c
+++ b/tests/rtdl/tls_align/test.c
@@ -1,0 +1,10 @@
+#include <assert.h>
+#include <stdint.h>
+
+extern _Thread_local char foo_thread_local[];
+extern _Thread_local char bar_thread_local[];
+
+int main() {
+	assert(!((uintptr_t)foo_thread_local & (16 - 1)));
+	assert(!((uintptr_t)bar_thread_local & (8 - 1)));
+}
--- a/options/rtdl/generic/linker.cpp
+++ b/options/rtdl/generic/linker.cpp
@@ -885,7 +885,7 @@ Tcb *allocateTcb() {
 
 	// To make sure that both the TCB and TLS data are sufficiently aligned, allocate
 	// slightly more than necessary and adjust alignment afterwards.
-	size_t alignOverhead = frg::min(alignof(Tcb), tlsMaxAlignment);
+	size_t alignOverhead = frg::max(alignof(Tcb), tlsMaxAlignment);
 	size_t allocSize = tlsInitialSize + sizeof(Tcb) + alignOverhead;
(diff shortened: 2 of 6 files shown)


ElectrodeYT (approved):
LGTM

We should merge this ASAP, my understanding (correct me if im wrong) is that new images do not boot to weston due to this bug.

tests/rtdl/tls_align/test.c
```diff
@@ -0,0 +1,10 @@
+#include <assert.h>
+#include <stdint.h>
+
+extern _Thread_local char foo_thread_local[];
+extern _Thread_local char bar_thread_local[];
+
+int main() {
+	assert(!((uintptr_t)foo_thread_local & (16 - 1)));
+	assert(!((uintptr_t)bar_thread_local & (8 - 1)));
```
ElectrodeYT, line 9: Where is this defined?

qookei: Good catch, forgot to add `libbar.c` to the commit.

ElectrodeYT (approved):

PR #967, opened 2024-01-23
rtdl: Add support for LD_LIBRARY_PATH

48cf:
I'm not sure how correct the implementation is but it was good enough for what we needed it for.


qookei (commented):
We should probably only parse `LD_LIBRARY_PATH` once instead of reparsing it for every library. Also, we could reuse the existing `libdirs` for this. (turn it into a global `frg::small_vector` that holds the default paths + ones dynamically added from `LD_LIBRARY_PATH`?)

qookei (approved):

no92 (approved):
LGTM

PR #968, opened 2024-01-23
meson: Don't use libstdc++ headers if cxxshim is available

48cf:
I've had issues building a host-mlibc and I fixed it by providing cxxshim and I needed this patch for cxxshim headers to be used.


no92 (approved):
LGTM

PR #969, opened 2024-01-23
sysdeps/managarm: Add missing call to __dlapi_enter

48cf:
This causes an ABI break, needed for upcoming gobject-introspection stuff


qookei (approved):

no92 (approved):
LGTM

PR #978, opened 2024-02-03
Small header fixes

no92:
These fixes came about when generating ASTs for every header, where previously unseen issues popped up with missing includes et al.


no92:
I had a brief glance over this with @avdgrinten, should be good according to him.

qookei (approved):

PR #981, opened 2024-02-07
ABI fixes for Rust

no92:


qookei:
RTLD constands are not ABI we care about when it comes to being compatible (musl vs glibc is also incompatible in this regard)

no92:
Closing in favor of #989.

PR #982, opened 2024-02-08
sysdeps/managarm: Fix AArch64 crt

Qwinci:
The entry point was changed to include the sp parameter but the AArch64 crts weren't updated.

### DIFF
--- a/sysdeps/managarm/aarch64/crt-src/Scrt1.S
+++ b/sysdeps/managarm/aarch64/crt-src/Scrt1.S
@@ -1,7 +1,8 @@
 .section .text
 .global _start
 _start:
-	adr x0, main
+	mov x0, sp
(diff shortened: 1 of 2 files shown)


qookei:
This needs a world revbump in Managarm, although not sure if we want to do that for only AArch64? (I think all CI pkgs get rebuilt on mondays anyway, so it wouldn't be broken for too long)

Dennisbonke:
> This needs a world revbump in Managarm, although not sure if we want to do that for only AArch64? (I think all CI pkgs get rebuilt on mondays anyway, so it wouldn't be broken for too long)

That is correct, but as AArcch64 is rolling, it doesn't matter. The entire world is rebuild on sunday anyway indeed.

qookei (approved):

PR #986, opened 2024-02-12
abis/linux: Make EIEIO conform to the maximum value that Linux reports

Dennisbonke:
This fixes a breakage in `elogind`. Technically an ABI break but not used outside the Managarm sysdeps.

### DIFF
--- a/abis/linux/errno.h
+++ b/abis/linux/errno.h
@@ -138,6 +138,6 @@
 
 
 // This is mlibc-specific.
-#define EIEIO           1524152434
+#define EIEIO           4095
 
 #endif // _ABIBITS_ERRNO_H


avdgrinten:
Can we re-use an existing POSIX error in place of EIEIO?

Is error code 4095 used for anything else in Linux?

ElectrodeYT (approved):
As discussed on Discord, there is discussion to be had on `EIEIO` :meme:

Given that this apparently can cause crashes, however, I think it would probably be a good idea to merge this somewhat quickly, in order to at least solve those crashes.

As I noted on Discord, and given the fact that my PR implemented EIEIO in the first place, I think that having a seperate managarm-specific (or, given that fact that its only really implemented in microkernels as far as I am aware, microkernel specific) error to indicate a internal error in a service that is seperate from POSIX is a good idea.

If it ever happens that we don't end up using this (much), we can obviously reconsider this later, however given that this error is also suitable for use in, for example, more advanced or untested device drivers, I believe keeping `EIEIO` is the correct call.

no92 (approved):
LGTM

PR #989, opened 2024-02-13
Generate rust bindings from mlibc headers

no92:
This PR adds a script to generate valid, working rust-libc bindings. It parses (installed) mlibc headers using libclang, and then outputs rust bindings. A configuration file is supplied to list stuff that needs to be ignored, either due to being broken, unnecessary or already provided in `unix/mod.rs` or `unix/linux_like/mod.rs`.


mintsuki:
Please split this PR up.

mintsuki (changes requested):
Please split up the PR into at least 2: one for the scripts and one for the rest of the mlibc bug fixes.

scripts/rust-libc-config.yml
64, line 1: Probably too awful to do now. But ideally there would be some comments explaining why certain things are ignored.

options/posix/include/dlfcn.h
```diff
@@ -3,12 +3,12 @@
 #define _DLFCN_H
 
 #define RTLD_LOCAL 0
-#define RTLD_NOW 1
-#define RTLD_GLOBAL 2
+#define RTLD_LAZY 1
+#define RTLD_NOW 2
```
64, line 7: This is an ABI break?

scripts/rust-libc.py
```diff
@@ -0,0 +1,699 @@
...
+						tc = Type(c)
+						assert(tc.is_valid())
+						name = str(tc)
+						if c.is_anonymous():
+							name = Type.cursor_name(c)
+						if Type.cursor_name(cursor) in config['force_struct_member_type']:
+							info = config['force_struct_member_type'][Type.cursor_name(cursor)]
+							if list(filter(lambda x: x['name'] == c.displayname, info)):
+								detail = next(filter(lambda x: x['name'] == c.displayname, info))
+								if "type" in detail:
```
64, line 356: bruv this is too indented

scripts/rust-libc.py
```diff
@@ -0,0 +1,699 @@
...
+	dry_run = args.dry_run
+
+	colorama.just_fix_windows_console()
+
+	with io.open("scripts/rust-libc-config.yml", "r") as f:
+		config = yaml.load(f, yaml.CSafeLoader)
+
+	path = pathlib.Path(args.path)
+
+	emit("// This file is autogenerated!")
```
64, line 560: Perhaps this header could be read from a separate file, it's a bit easier to read that way + syntax highlighting etc?

scripts/rust-libc.py
```diff
@@ -0,0 +1,699 @@
...
+    fn fmt(&self, f: &mut ::fmt::Formatter) -> ::fmt::Result {
+        f.debug_struct("ifreq")
+            .finish()
+    }
+}""")
+
+	state = State()
+
+	if not args.file:
+		for header in path.rglob("*.h"):
```
64, line 691: It could be nice to sort the headers here, so that the contents are output in a fully deterministic order

options/posix/generic/posix_time.cpp
```diff
@@ -30,3 +36,8 @@ int futimes(int, const struct timeval[2]) {
 	__ensure(!"Not implemented");
 	__builtin_unreachable();
 }
+
+int lutimes(const char *, const struct timeval[2]) {
```
qookei, line 40: `lutimes` is the same as `utimes` but with `AT_SYMLINK_NOFOLLOW` passed as the flags for `sys_utimensat`

scripts/rust-libc.py
```diff
@@ -0,0 +1,699 @@
...
+
+def log_err(prefix, msg):
+	global errors_emitted
+
+	print(f"{colorama.Fore.RED}{prefix}{colorama.Style.RESET_ALL}: {msg}")
+	errors_emitted += 1
+
+def emit(msg):
+	if not dry_run:
+		print(msg)
```
64, line 43: Why not just output to stdout instead of having dry_run etc? Then user can redirect where they want, and errors logged to stderr

no92:
Rebased onto current master.

options/ansi/include/stdio.h
```diff
@@ -151,7 +151,7 @@ int vasprintf(char **, const char *, __builtin_va_list);
 // [C11-7.21.7] Character input/output functions
 
 int fgetc(FILE *stream);
-char *fgets(char *__restrict buffer, size_t max_size, FILE *__restrict stream);
+char *fgets(char *__restrict buffer, int max_size, FILE *__restrict stream);
```
Qwinci, line 154: Should also change the function signature in stdio.cpp and also in `gets` use INT_MAX instead of SIZE_MAX.

Geertiebear (approved):
Aside from the messy python code I don't see a reason why this can't be merged right now. We could always clean up the script if there's ever motivation to.

scripts/rust-libc.py
```diff
@@ -0,0 +1,581 @@
...
+		type_iter = pointee
+
+		while type_iter.kind == TypeKind.POINTER:
+			prefix += "*mut "
+			type_iter = type_iter.get_pointee()
+
+		t = type_iter.spelling.removeprefix("const ")
+		tokens = t.split(' ')
+
+		match tokens[0]:
```
ArsenArsen, line 85: you can match multiple values at once, nesting matches is nearly always wrong

scripts/rust-libc.py
```diff
@@ -0,0 +1,581 @@
...
+			return f"r#{name}"
+		return name
+
+	def cursor_name(c: Cursor):
+		d = c.type.get_declaration()
+		if d and d.is_anonymous():
+			return f"anon_{pathlib.Path(str(d.location.file)).stem}_line{d.location.line}"
+		return Type.escape_name(c.displayname)
+
+class State:
```
ArsenArsen, line 206: can this be a dataclass?

scripts/rust-libc.py
```diff
@@ -0,0 +1,581 @@
...
+	def __init__(self):
+		self.functions = []
+		self.macros = []
+		self.types = []
+		self.structs = []
+		self.variables = []
+
+class RustBindingGenerator:
```
ArsenArsen, line 214: ditto

scripts/rust-libc.py
```diff
@@ -0,0 +1,581 @@
...
+				if gen[i].kind == TokenKind.PUNCTUATION and gen[i].spelling in ("(", ")"):
+					if not (i == 0 or i == (len(gen) - 1)):
+						tokens.append(gen[i].spelling)
+					i += 1
+				elif gen[i].kind == TokenKind.PUNCTUATION and gen[i].spelling == "-":
+					is_unsigned = False
+					i += 1
+				elif gen[i].kind in (TokenKind.LITERAL, TokenKind.IDENTIFIER, TokenKind.PUNCTUATION):
+					spelling = gen[i].spelling
+					if spelling.endswith("ULL") and is_num(spelling[:-3]):
```
ArsenArsen, line 252: this can definitely be simpler

ArsenArsen (changes requested):
this does not appear to be `black`-formatted, which, IIRC, we agreed to use for all `.py` code?

(overall, generating more code like this is quite a good thing, +1)

scripts/rust-libc.py
no92: I don't see how, as this is simple and quite clear

no92:
Updated the script to be formatted with `black` and addressed some comments.

scripts/rust-libc.py
ArsenArsen: actually, even better, this is an array

```py
match tokens:
    case ["char", *_]:
        return f"{prefix}::c_char"
    case ["struct", x, *_] if x in config["force_local_types"]:
        return prefix + x
    case ["struct", x, *_]:
        return f"{prefix}::{x}"
    case ["int", *_]:
...
```

note that many of the `*_` are there only because your original code doesn't length-check, so I eagerly ignore other entries.  if you know sizes in advance, you don't need to do that:

```py
>>> def f(x):  # bad formatting, repl code
...     match x:
...             case ["foo", *_]:
...                     print("foo")
...             case ["bar"]:
...                     print("bar")
...             case _:
...                     print("gone")
... 
>>> f(["foo", 1, 2, 3])
foo
>>> f(["foo"])
foo
>>> f(["bar"])
bar
>>> f(["bar", 1])
gone
>>> 
```

no92: Applied.

Geertiebear (approved):
Aesthetically the script looks much better now :upside_down_face: Do still have issues with the code quality but as I mentioned before I'm not of the opinion that should be a blocker for now.

scripts/rust-libc-config.yml
```diff
@@ -0,0 +1,2028 @@
...
+    - "B576000"
+    - "B921600"
+    - "B1000000"
+    - "B1152000"
+    - "B1500000"
+    - "B2000000"
+    - "B2500000"
+    - "B3000000"
+    - "B3500000"
+    - "B4000000"
```
Dennisbonke, line 2028: Missing a newline at the end of file. We should fix that at the next opportunity but it's not a blocker for me.

Dennisbonke (approved):
LGTM

PR #991, opened 2024-02-15
Add .note.GNU-stack to crt files of linux x86 sysdep

FedorLap2006:
Mark stack in crti and crtn of linux sysdep non-executable.
This also makes the behavior consistent with other crt files.


mintsuki (approved):

no92 (approved):
LGTM

PR #992, opened 2024-02-16
misc: Depend on 'cshim', if available, instead of host freestanding headers

mintsuki:


Qwinci:
LGTM other than the cshim wrap is missing.

mintsuki:
Yeah, it didn't apply because it is gitignored, oops. Fixed.

options/ansi/include/limits.h
```diff
@@ -3,6 +3,10 @@
 
 #define CHAR_BIT 8
 
+#ifndef MB_LEN_MAX
+# define MB_LEN_MAX 1
```
Qwinci, line 7: Shouldn't this be 4 for UTF-8? I am not sure if we actually implement any of the conversions functions properly yet but when we do this should likely be 4.

Qwinci (changes requested):

options/ansi/include/limits.h
mintsuki: On Linux/glibc it appears to be 16... not sure.

Qwinci: On musl it is 4, maybe it being bigger on glibc has something to do with locales.

qookei (approved):

no92 (approved):
LGTM, this builds managarm x86_64.

PR #994, opened 2024-02-17
x86: 64-bit time

FedorLap2006:
Currently `time_t` is `long`, which is fine 64 bit arches, since `long` is always 8 bytes wide.
However, on x86 `long` is 4 bytes wide.

This PR changes `time_t` to always be 8 bytes wide, by making it `__mlibc_int64`.

### DIFF
--- a/options/internal/include/bits/field-padding.h
+++ b/options/internal/include/bits/field-padding.h
@@ -1,15 +1,9 @@
 #ifndef MLIBC_FIELD_PADDING_H
 #define MLIBC_FIELD_PADDING_H
 
-#ifdef __GNUC__
-
-#define __MLIBC_FIELD_PADDED(T, F, PT) \
-	PT : (sizeof(PT)-sizeof(T))*8*(BYTE_ORDER == BIG_ENDIAN); \
+#define __MLIBC_FIELD_PADDED(T, AT, F) \
+	AT : (sizeof(AT)-sizeof(T))*8*(__BYTE_ORDER__ == __ORDER_BIG_ENDIAN__); \
 	T F; \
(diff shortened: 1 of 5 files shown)


FedorLap2006:
After a bit of research, I've found out this is not so trivial, as this would break linux ABI on x86.

sysdeps/linux/generic/sysdeps.cpp
```diff
@@ -763,9 +780,42 @@ int sys_setpriority(int which, id_t who, int prio) {
...
+  
+
+  if (old_value) {
+    old_value->it_interval.tv_sec = old_raw[0];
+    old_value->it_interval.tv_usec = old_raw[1];
+    old_value->it_value.tv_sec = old_raw[2];
+    old_value->it_value.tv_usec = old_raw[3];
+  }
+
+#endif
```
qookei, line 818: Wrong indentation

options/internal/include/bits/field-padding.h
```diff
@@ -0,0 +1,15 @@
+#ifndef MLIBC_FIELD_PADDING_H
+#define MLIBC_FIELD_PADDING_H
+
+#ifdef __GNUC__
+
+#define __MLIBC_FIELD_PADDED(T, F, PT) \
+  PT : (sizeof(PT)-sizeof(T))*8*(__BYTE_ORDER__ == __ORDER_BIG_ENDIAN__); \
+  T F; \
+  PT : (sizeof(PT)-sizeof(T))*8*(__BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__)
```
qookei, line 9: Wrong indentation

marv7000:
This will be fixed by https://github.com/managarm/mlibc/pull/1823

PR #996, opened 2024-02-17
options/glibc: Don't overwrite posix icmp6.h

Qwinci:
Currently if the glibc option is enabled it overwrites the posix icmp6.h header with the extension header.


qookei (approved):

no92 (approved):
LGTM

PR #1003, opened 2024-02-18
options/rtdl: Implement _dl_find_object

Qwinci:


qookei (commented):
Maybe this should be gated behind the glibc option? (or _GNU_SOURCE in the header at the very least)

options/posix/include/dlfcn.h
```diff
@@ -37,11 +39,26 @@ typedef struct {
 	void *dli_saddr;
 } Dl_info;
 
+struct dl_find_object {
+	unsigned long long dlfo_flags;
+	void *dlfo_map_start;
+	void *dlfo_map_end;
+	struct link_map *dlfo_link_map;
+	void *dlfo_eh_frame;
+	unsigned long long __dlfo_unused[7];
```
no92, line 48: glibc seems to have two optional members (hidden behind DLFO_STRUCT_HAS_EH_DBASE and DLFO_STRUCT_HAS_EH_COUNT respectively), but those don't get used on my system apparently.

no92 (approved):
LGTM

Qwinci:
This should be good now, but #1008 should be merged in first so that the function/struct in dlfcn.h get put inside the ifdef block as I removed that from this pr.

options/posix/include/dlfcn.h
```diff
@@ -44,11 +44,45 @@ typedef struct {
...
+#define DLFO_EH_SEGMENT_TYPE PT_GNU_EH_FRAME
+#endif
+
+struct dl_find_object {
+	unsigned long long dlfo_flags;
+	void *dlfo_map_start;
+	void *dlfo_map_end;
+	struct link_map *dlfo_link_map;
+	void *dlfo_eh_frame;
+#if DLFO_STRUCT_HAS_EH_DBASE
```
Geertiebear, line 67: Comments on the endifs please.

options/rtdl/generic/main.cpp
```diff
@@ -814,6 +814,53 @@ void __dlapi_enter(uintptr_t *entry_stack) {
...
+				continue;
+			}
+			end_addr = frg::max(end_addr, phdr->p_vaddr + phdr->p_memsz);
+		}
+
+		result->dlfo_flags = 0;
+		result->dlfo_map_start = (void *)object->baseAddress;
+		result->dlfo_map_end = (void *)(object->baseAddress + end_addr);
+
+// TODO: fill these fields with proper values
```
Geertiebear, line 848: Usually TODO's about incomplete functionality are logged.

options/rtdl/generic/main.cpp
```diff
@@ -814,6 +814,53 @@ void __dlapi_enter(uintptr_t *entry_stack) {
 #endif
 }
 
+#if __MLIBC_GLIBC_OPTION
+
+extern "C" [[gnu::visibility("default")]] int _dl_find_object(void *address, dl_find_object *result) {
+	for(size_t i = 0; i < initialRepository->loadedObjects.size(); i++) {
```
Geertiebear, line 820: Since the iterator variable is not used it would be more idiomatic to use a range based for loop here.

Geertiebear (approved):

PR #1004, opened 2024-02-18
misc: Drop our version of stdint.h, wrap limits.h instead of reimplementing

mintsuki:


Qwinci:
The libc does need to provide `stdint.h` as the gcc stdint.h includes it when not compiling in freestanding mode.

mintsuki:
> The libc does need to provide `stdint.h` as the gcc stdint.h includes it when not compiling in freestanding mode.

@Qwinci this is fixed by changing the gcc `config.gcc` file's port section's `use_gcc_stdint` option from `wrap` to `provide`.

64:
+1 for removing a shitton of code, I like this

Do downstream users need to fiddle their GCC config in order for this to work?

mintsuki:
Yeah, but it's just as simple as I said above.

ElectrodeYT (approved):
LGTM

options/internal/include/bits/types.h
```diff
@@ -1,319 +0,0 @@
...
-	__MLIBC_CHECK_TYPE(__mlibc_int_fast8,  __INT_FAST8_TYPE__);
-	__MLIBC_CHECK_TYPE(__mlibc_int_fast16, __INT_FAST16_TYPE__);
-	__MLIBC_CHECK_TYPE(__mlibc_int_fast32, __INT_FAST32_TYPE__);
-	__MLIBC_CHECK_TYPE(__mlibc_int_fast64, __INT_FAST64_TYPE__);
-
-	__MLIBC_CHECK_TYPE(__mlibc_uint_fast8,  __UINT_FAST8_TYPE__);
-	__MLIBC_CHECK_TYPE(__mlibc_uint_fast16, __UINT_FAST16_TYPE__);
-	__MLIBC_CHECK_TYPE(__mlibc_uint_fast32, __UINT_FAST32_TYPE__);
-	__MLIBC_CHECK_TYPE(__mlibc_uint_fast64, __UINT_FAST64_TYPE__);
-#endif
```
avdgrinten, line 317: Has this been fixed in LLVM? If these asserts still fail, we can't apply the commit. Otherwise, we get an inconsistent ABI and can't link clang-compiled code together with GCC-compiled code.

Qwinci:
LGTM

options/internal/include/bits/types.h
mintsuki: This has been sorted out by the new updates, thank you for pointing it out.

no92 (approved):
LGTM as long as it works with unpatched gcc.

mintsuki:
It does.

mintsuki:
Closing because this is getting too hairy, sadly. The current solution seems to be the most optimal given the unholy circumstances.

PR #1007, opened 2024-02-19
Signature fix for linux syscall0 and wctype_l/iswctype_l

Qwinci:


no92 (approved):
LGTM but needs to be PR'ed against `abi-break`.

Geertiebear (approved):
LGTM

PR #1011, opened 2024-02-19
options/rtdl: Return LinkerError instead of nullptr on error

Qwinci:
I forgot to update this case in the dl non-elf pr.


qookei (approved):

Dennisbonke (approved):
LGTM

PR #1018, opened 2024-02-24
options/rtdl: Implement tlsdesc for x86_64

Qwinci:
I am not sure if the dynamic tlsdesc assembly is right at all, so someone should check it.

### DIFF
--- a/options/rtdl/generic/linker.cpp
+++ b/options/rtdl/generic/linker.cpp
@@ -1736,8 +1736,8 @@ void Loader::_processStaticRelocations(SharedObject *object) {
 	}
 }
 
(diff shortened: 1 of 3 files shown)


no92 (approved):
LGTM

options/rtdl/generic/linker.cpp
```diff
@@ -1810,8 +1810,8 @@ void Loader::_processLazyRelocations(SharedObject *object) {
 		}
 #endif
 // TODO: TLSDESC relocations aren't aarch64 specific
```
Geertiebear, line 1812: Update the comment

mintsuki (approved):

PR #1021, opened 2024-03-03
rtdl: make default library search paths configurable at build time

64:
One scenario where overriding the default search paths is useful is when mlibc is used as a 'secondary' libc on a system. Absent this patch, mlibc will look for libraries in `/lib` and `/usr/lib`, but we want it to look in `/some/mlibc/sysroot/lib` instead.

Note that this is only needed if dynamic linking is used. To my knowledge, `musl` and `glibc` do not support a similar feature, requiring local patching instead.

### DIFF
--- a/options/rtdl/generic/main.cpp
+++ b/options/rtdl/generic/main.cpp
@@ -48,8 +48,8 @@ frg::manual_box<Scope> globalScope;
 
 frg::manual_box<RuntimeTlsMap> runtimeTlsMap;
 
-// We use a small vector of size 4 to avoid memory allocation for the default library paths
-frg::manual_box<frg::small_vector<frg::string_view, 4, MemoryAllocator>> libraryPaths;
+// We use a small vector to avoid memory allocation for the default library paths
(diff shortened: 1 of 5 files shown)


options/rtdl/generic/main.cpp
```diff
@@ -340,11 +340,21 @@ extern "C" void *interpreterMain(uintptr_t *entry_stack) {
...
+		const char *path = paths + i;
+
+		size_t j = 0;
+		for (const char *p = path; *p; p++) {
+			if (*p == '\n')
+				break;
+			j++;
+		}
+
+		libraryPaths->push_back(frg::string_view{path, j});
```
qookei, line 355: Why not do something like this?
```cpp
for(auto path : parseList(MLIBC_DEFAULT_LIBRARY_PATHS, "\n"))
	libraryPaths->push_back(path);
```

qookei (approved):

mintsuki (approved):

PR #1022, opened 2024-03-03
options/elf: Fix aarch64 dtpmod/dtprel values

Qwinci:


Qwinci:
Originally the aarch64 specification specified those like they were but then gnu changed them to be the other way around so thats what eg. binutils generates.

qookei (approved):
Looks good, FTR here's the LLVM changes for this https://reviews.llvm.org/D65459

mintsuki (approved):

PR #1027, opened 2024-03-04
options/rtld: RTDL -> RTLD

mintsuki:


qookei (approved):
Looks good, although there is a merge conflict. The ABI break is for an internal symbol and can be ignored.

Geertiebear (approved):

PR #1031, opened 2024-03-05
Implement `RLIMIT_NOFILE` and `_SC_OPEN_MAX` for managarm

no92:


qookei (approved):

Geertiebear (approved):

sysdeps/managarm/generic/file.cpp
```diff
@@ -2522,5 +2522,35 @@ int sys_unlockpt(int fd) {
 	return 0;
 }
 
+int sys_getrlimit(int resource, struct rlimit *limit) {
+	switch(resource) {
+		case RLIMIT_NOFILE:
+			/* TODO: change this once we support more than 512 */
+			limit->rlim_cur = 512;
```
Geertiebear, line 2529: Actually, on second thought, wouldn't it be a better idea to make a define/constexpr for this so that it's consistent between ``sys_getrlimit`` and the fd table?

qookei: That'd probably be better, the 512 constant is also hard-coded in `getHandleForFd`. OTOH this is just `page size / sizeof(HelHandle)` so it's unlikely to change, except for when we allow for the FD table to be dynamically-sized.

PR #1033, opened 2024-03-07
options/rtdl: Fix tls in tlsAboveTp case

Qwinci:
Store the tls offset relative to the end of the tcb instead of the start of the tp in the tlsAboveTp case and take into account possible padding between the tls space and the tp in relocations.


Qwinci:
This is ready to review now.

qookei (approved):

mintsuki (approved):

PR #1043, opened 2024-03-27
options/internal: use provided path in file_window

FedorLap2006:
Use provided `path` parameter, instead of hardcoded `/etc/localtime` in constructor of `file_window`.


no92 (approved):
LGTM

PR #1044, opened 2024-03-27
ci: add job to test rust bindings generation

no92:


.github/workflows/rust-libc.yml
```diff
@@ -0,0 +1,50 @@
...
+    check-bindings:
+        name: Verify bindings generation
+        runs-on: ubuntu-20.04
+        container:
+          image: ghcr.io/managarm/mlibc-crossers:latest
+        steps:
+          - name: Install prerequisites
+            run: |
+                apt-get update
+                apt-get install -y ninja-build qemu-user \
```
64, line 15: qemu-user is not needed

64 (approved):
This just tests that the script doesn't error, right?

no92:
> This just tests that the script doesn't error, right?

Exactly.

ElectrodeYT (approved):

PR #1047, opened 2024-03-30
improvements for rust 1.77.1

no92:


no92:
Ready for review.

options/linux/generic/sys-statx.cpp
```diff
@@ -3,9 +3,40 @@
 #include <bits/ensure.h>
 
 #include <mlibc/debug.hpp>
+#include <mlibc/posix-sysdeps.hpp>
 #include <mlibc/linux-sysdeps.hpp>
 
 int statx(int dirfd, const char *pathname, int flags, unsigned int mask, struct statx *statxbuf) {
+	if(!mlibc::sys_statx) {
```
64, line 10: Genuine question, does it make more sense for sysdeps to implement this fallback at their own risk? Seems like it could lead to unintended consequences if we add it here

options/posix/generic/sys-resource-stubs.cpp
```diff
@@ -42,6 +42,8 @@ int getrlimit(int resource, struct rlimit *limit) {
 	return 0;
 }
 
+[[gnu::alias("getrlimit")]] int getrlimit64(int resource, struct rlimit *limit);
```
64, line 45: Should it be behind an ifdef for 32 bit platforms?

options/linux/generic/sys-statx.cpp
64: According to the man page we should set stx_mask appropriately, then it would be an acceptable catch-all fallback

64 (changes requested):

options/posix/generic/sys-resource-stubs.cpp
no92: No, this is intended to be present on 32-bit platforms as well.

options/linux/generic/sys-statx.cpp
```diff
@@ -3,9 +3,40 @@
...
+		statxbuf->stx_btime.tv_nsec = statbuf.st_mtim.tv_nsec;
+		statxbuf->stx_ctime.tv_sec = statbuf.st_ctim.tv_sec;
+		statxbuf->stx_ctime.tv_nsec = statbuf.st_ctim.tv_nsec;
+		statxbuf->stx_mtime.tv_sec = statbuf.st_mtim.tv_sec;
+		statxbuf->stx_mtime.tv_nsec = statbuf.st_mtim.tv_nsec;
+
+		return 0;
+	}
+
 	MLIBC_CHECK_OR_ENOSYS(mlibc::sys_statx, -1);
```
64, line 40: This line is now redundant

64 (approved):

PR #1049, opened 2024-04-02
sysdeps: Lyre/Ironclad: call __dlapi_enter() on entry

mintsuki:


qookei (approved):
Keep in mind this is an ABI break and every executable needs to be relinked

mintsuki:
> Keep in mind this is an ABI break and every executable needs to be relinked

Yeah that's okay, we don't care about that atm.

PR #1053, opened 2024-04-07
fix(abis/linux): remove redefinition of `S_IEXEC`

Andy-Python-Programmer:
First defined: https://github.com/managarm/mlibc/blob/266ec427d78fa3deae7c288aaffc3709dbdb2903/abis/linux/stat.h#L29
Redefined in the same file but a few lines below: https://github.com/managarm/mlibc/blob/266ec427d78fa3deae7c288aaffc3709dbdb2903/abis/linux/stat.h#L44


Dennisbonke (approved):
LGTM

PR #1054, opened 2024-04-13
options/internal: use memcpy when memmove arguments don't overlap

64:


options/internal/generic/essential.cpp
```diff
@@ -189,14 +189,21 @@ void *memset(void *dest, int byte, size_t count) {
 // --------------------------------------------------------------------------------------
 
 void *memmove(void *dest, const void *src, size_t size) {
-	char *dest_bytes = (char *)dest;
-	char *src_bytes = (char *)src;
-	if(dest_bytes < src_bytes) {
+	// Use uintptr_t for pointer comparisons because otherwise it's undefined behaviour
+	// when dest and src point to different objects.
+	uintptr_t udest = (uintptr_t)dest;
```
no92, line 194: Might be nice to use C++-style `uintptr_t udest = uintptr_t(dest)` here.

options/internal/generic/essential.cpp
```diff
@@ -189,14 +189,21 @@ void *memset(void *dest, int byte, size_t count) {
...
+	uintptr_t usrc = (uintptr_t)src;
+
+	if(udest < usrc || usrc + size <= udest) {
```
no92, line 197: Is there any reason why this would only to memmove in the `usrc < udest` case?

64: forward_copy will clobber itself when src < dst and they overlap

options/internal/generic/essential.cpp
64: I think if we're gonna prefer one style over the other it should be enforced somewhere (e.g in clang-tidy), we do use C-style casts in many places

Qwinci: And also `uintptr_t(dest)` is still doing the same logic as c-style cast so it isn't really any better from a logic standpoint (unlike the proper c++ style casts like static_cast/reinterpret_cast/const_cast that don't just try all the casts after each one like c-style casts do).

options/internal/generic/essential.cpp
avdgrinten: This comment was probably about the case that `src < dst` and there is no overlap.

options/internal/generic/essential.cpp
64: Uses `reinterpret_cast` now

options/internal/generic/essential.cpp
64: Don't quite understand -- that's exactly the case I'm handling via memcpy now

no92: Ah I misread the condition.

no92 (approved):
LGTM

PR #1082, opened 2024-06-12
meta: prevent linux_kernel_headers from being set to `/usr/include`

no92:


Geertiebear (commented):
This feels pretty ad-hoc, this is just one of the many paths you could fill in that are bogus. What is this actually protecting against?

Geertiebear:
I understand the reaction to avoid common errors, but this is not something that should be handled in the build system. I suggest making this a disclaimer in the README instead.

no92:
Updated the PR with @ArsenArsen's suggestion to use a list of compiler search paths to check against, instead of hardcoding `/usr/include`.

mintsuki:
Superseded by #1321

Issue #607, opened 2022-07-01
gcc: `__cxa_finalize` is unimplemented


64:
We don't need this since dlclose is a no-op, but we should still probably call destructors for all shared libraries when the process exits.

PR #1092, opened 2024-07-17
options/rtld: Call fini functions on exit

Qwinci:
This is needed to support functions marked with attribute gnu::destructor. ~The way I implemented it it destructs the objects in reverse order that they were initialized in and first runs functions registered with __cxa_atexit for the object and after that the fini functions for that object, I am not sure if that's completely correct but to me it seems reasonable enough so in case of an object that doesn't have any circular dependencies both the atexit functions and destructor functions are run after the object can't be used by anything anymore.~ It turned out that there is an autogenerated call to __cxa_finalize that should call the functions registered for that dso, so I implemented that.


Dennisbonke (approved):
LGTM

no92 (approved):
LGTM with my limited understanding of how this works.

Geertiebear (commented):
LGTM, could you add some tests for this?

64:
Fixes #607

PR #1093, opened 2024-07-20
meta: rename ANSI option to ANSI tests option

no92:
As disabling the ANSI option makes no sense, convert the option to only disable ANSI option tests.


qookei (commented):
`disable_ansi_tests_option` -> `disable_ansi_option_tests`? It's tests for the ANSI option after all.

no92:
Renamed the option as requested.

qookei (approved):

Geertiebear:
I'd be more in favour of removing this option completely, it seems weird to me to disable specifically the ansi tests.

mintsuki:
I believe #1146 fixes this.

PR #1111, opened 2024-08-19
Fix relr relocations and match glibc behavior in frexp

Qwinci:


options/rtld/generic/linker.cpp
```diff
@@ -1731,11 +1731,14 @@ void Loader::_processStaticRelocations(SharedObject *object) {
 				*addr++ += object->baseAddress;
 			}else {
 				// Odd entry indicates entry is a bitmap of the subsequent locations to be relocated.
-				for(int i = 0; entry; ++i) {
-					if(entry & 1) {
+				for(int i = 0;; ++i) {
```
qookei, line 1734: Maybe this would be clearer if we just did `entry >>= 1;` outside of the loop (maybe with a comment saying that this bit is not actually marking a relocation) and left the loop as it is? Ditto for `relocateSelf`.

Qwinci: I kept the loops as-is and instead removed the addr increment + base addition in the above if as the first bit is also used the same way as the others just that its always set.

qookei (approved):

mintsuki (approved):

PR #1114, opened 2024-08-29
Typo

da4089:


no92:
The commit should follow the style guide. Something like `docs: fix typo` should work.

PR #1115, opened 2024-09-01
options/posix: change sys_getgroup parameters and implement putgrent

Mathewnd:
``sys_getgroup`` should take a ``gid_t *`` instead of a ``const gid_t *``, as it is expected to write there.
Implement ``putgrent``, needed by ``useradd`` to create the user group when creating a user.

### DIFF
--- a/options/posix/generic/grp-stubs.cpp
+++ b/options/posix/generic/grp-stubs.cpp
@@ -300,9 +300,41 @@ int initgroups(const char *, gid_t) {
 	return 0;
 }
 
-int putgrent(const struct group *, FILE *) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+int putgrent(const struct group *g, FILE *f) {
+	auto invalid = [](const char *s) {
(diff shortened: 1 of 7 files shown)


options/posix/generic/grp-stubs.cpp
```diff
@@ -300,9 +300,41 @@ int initgroups(const char *, gid_t) {
...
+
+	for(int i = 0; g->gr_mem != nullptr && g->gr_mem[i] != nullptr; ++i) {
+		r = fprintf(f, "%s%s", i ? "," : "", g->gr_mem[i]);
+		if(r < 0)
+			goto leave;
+	}
+
+	r = fputc('\n', f);
+
+	leave:
```
no92, line 335: This should be unindented once.

no92 (commented):
LGTM with one style complaint.

options/posix/generic/grp-stubs.cpp
Mathewnd: Done

no92 (approved):
LGTM

PR #1116, opened 2024-09-02
fix: options/ansi make UINT_MAX unsigned

awewsomegamer:
Compiling MPFR yielded the following messages:
```
/host/usr/include/limits.h:76:31: warning: integer overflow in expression of type 'int' results in '-2' [-Woverflow]
   76 | #define UINT_MAX (__INT_MAX__ * 2 + 1)
../../mpfr-src/src/invert_limb.h:136:7: note: in expansion of macro 'MPFR_STAT_STATIC_ASSERT'
  136 |       MPFR_STAT_STATIC_ASSERT (4182025 <= UINT_MAX);   
```

After some time looking through at possible fixes, I have discovered that UINT_MAX is evaluated to be a signed integer
rather than an unsigned one in ansi/include/limits.h, this can be fixed by adding 2 U's.

### DIFF
--- a/options/ansi/include/limits.h
+++ b/options/ansi/include/limits.h
@@ -73,7 +73,7 @@
 
 #define INT_MIN (-__INT_MAX__ - 1)
 #define INT_MAX __INT_MAX__
-#define UINT_MAX (__INT_MAX__ * 2 + 1)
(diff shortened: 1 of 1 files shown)


awewsomegamer:
Added fix for __SHRT_MAX_ typo.

PR #1119, opened 2024-09-06
meson: hide all libc.so C++ symbols via a linker script

ArsenArsen:
result: https://gist.github.com/ArsenArsen/570e87410c48f5b17f46d579269f10c2

btw, I noticed that we have a C++ function that probably should not be a C++ function:

```
0000000000068a58 T seed48(unsigned short*)
000000000008354a T login_tty(int)
000000000009f478 T wcstoimax(wchar_t const*, wchar_t**, int)
000000000009f4b7 T wcstoumax(wchar_t const*, wchar_t**, int)
```

this would imply we lack tests for those..

I recall some discussion about how to detect such cases before in the mlibc channel, but I do not recall the result.  if you can find it/remember it, please share

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -507,12 +507,16 @@ if not headers_only
 			dependencies: rtld_deps + rtlib_deps,
 			install: true
 		)
+		hide_everything_ld = (meson.current_source_dir()
+				      / 'scripts/hide-everything.ld')
 		libc_shared = shared_library('c', libc_all_sources + internal_dso_sources,
 			cpp_args: [libc_cpp_args],
 			include_directories: libc_include_dirs,
 			dependencies: libc_deps + rtlib_deps,
 			link_with: [ld_shared_lib],
 			link_whole: libc_sublibs,
+			link_args: ['-Wl,--version-script,' + hide_everything_ld],
+			link_depends: [hide_everything_ld],
 			install: true
 		)
 	endif
--- a/scripts/hide-everything.ld
+++ b/scripts/hide-everything.ld
@@ -1,0 +1,4 @@
+{
+  /* Hide all C++ symbols.  */
+  local: _Z*;
+};


no92:
Detecting such errors might also be a convenient extra feature for a possible ABI checker.

mintsuki (approved):

mintsuki:
LGTM

ArsenArsen:
obviously the abi break is expected

no92:
As is, this needs some changes; managarm uses `__mlibc_getPassthrough` in a few places, so this PR currently breaks boot.

ArsenArsen:
that shouldn't break, since this only excludes `_Z*`?  or am I missing something?

qookei:
It's not marked `extern "C"` :^)

ArsenArsen:
that's a silly name for a C++ function.. but, OK, we can export it also

qookei:
Marking it `extern "C"` would be simpler, and I suppose more correct for a libc

ArsenArsen:
and an abi break, which might be problematic depending on what it's used for

Geertiebear:
Does that actually get used by other binaries?

ArsenArsen:
I imagine boot wouldn't fail without it if it wasn't

ArsenArsen:
based on the other `__mlibc_` functions not being mangled I'm guessing that this is unintentionally not extern C.  anyway, try this.  might require that we start passing `--no-undefined-version` at some point

ArsenArsen:
don't merge yet, need to add a dep to the link_deps of the .so

no92:
With managarm/managarm#715, the need to make `__mlibc_getPassthrough` be `extern "C"` is avoided. It no longer needs to be declared with C linkage or be excepted in the linker script.

ArsenArsen:
okay, I'll revert that

no92:
Works on my machine™

no92 (approved):
LGTM

PR #1120, opened 2024-09-06
Revert "options/ansi: convert use of std::array to frg::array"

mintsuki:
This reverts commit 24dbb3b5a9f8494dbf2de6f4e92ff9a4d815dd43.


mintsuki:
This is because we now use libstdc++ based freestnd-cxx-hdrs.

PR #1124, opened 2024-09-14
Fix sem_trywait and linux sysdep pselect

Qwinci:


Geertiebear (commented):
Thanks for the fixes! How difficult is writing a regression test for this?

Qwinci:
> Thanks for the fixes! How difficult is writing a regression test for this?

Probably not that hard, Ill write some tests for the semaphores.

Geertiebear:
Does that also include a regression test for the bug on ``sem_trywait()``?

Qwinci:
> Does that also include a regression test for the bug on `sem_trywait()`?

Yes, the one that does 4 sem_trywaits in a row on a semaphore that only has 3 count. That would be where the fourth sem_trywait would incorrectly return 0 and the counter would wrap around.

Geertiebear (approved):

PR #1127, opened 2024-09-20
misc: Prefix parameters in headers with underscores

Qwinci:
Prefix function parameters in headers with two underscores to reduce changes of clashing with eg. user defined macros, add missing parameter names and also a couple restrict fixes.


Geertiebear (approved):
LGTM

Dennisbonke (approved):
LGTM

PR #1129, opened 2024-09-21
Fix functions that were mangled by accident

qookei:
As discovered by #1119, some functions were erroneously mangled, which was caused by missing/broken prototypes in headers.


no92 (approved):
LGTM

PR #1132, opened 2024-09-24
Compilation with clang for managarm

no92:
These changes are needed to make mlibc compile for managarm, with clang. The result seems to work fine as a drop-in replacement for a gcc-compiled one.


Dennisbonke (approved):
LGTM

Geertiebear:
Could we add a CI case for this as well maybe? Then we could track our support.

no92:
We already have a GHA job, this patch just carries some changes to work for managarm, too.

PR #1134, opened 2024-10-01
build: Add C standard (as GNU11) to root meson.build file

mintsuki:


no92:
What exactly is the motivation for this? Do we use C features that require at least C11?

mintsuki:
It is always advisable to specify the C version to use. Without this, meson just does not pass `-std=...` altogether to the C compiler driver, which can cause issues in the future if the default C version used by the compiler changes (for gcc, it is currently GNU17).

no92:
That's a fair point. I do wonder what the actual minimal required C std is that mlibc needs; if I have enough time, I'll try to investigate.

mintsuki:
I think this is fine to merge as is, for now. The need for GNU11 over C11 is because there are uses of the `asm` keyword and other stuff that may rely on the GNU variant.

PR #1136, opened 2024-10-03
sysdeps/astral: upstream port

Mathewnd:


meson.build
```diff
@@ -264,6 +264,15 @@ elif host_machine.system() == 'dripos'
 	rtld_include_dirs += include_directories('sysdeps/dripos/include')
 	libc_include_dirs += include_directories('sysdeps/dripos/include')
 	subdir('sysdeps/dripos')
+elif host_machine.system() == 'astral'
+        disable_crypt_option = true
```
no92, line 268: You seem to have mixed indentation here.

sysdeps/astral/crt-x86_64/crt0.S
```diff
@@ -0,0 +1,7 @@
+
+.section .text
+.global _start
+_start:
+    mov $main, %rdi
```
no92, line 5: Indentation uses spaces instead of tabs here.

no92 (commented):
LGTM, all I can complain about is the two places that indent using 4 spaces instead of tabs.

meson.build
```diff
@@ -264,6 +264,15 @@ elif host_machine.system() == 'dripos'
...
+        disable_iconv_option = true
+        disable_intl_option = true
```
mintsuki, line 270: I would leave these settings up to the configurer of mlibc rather than hardcoding them here. It also would align with the other ports.

Geertiebear (changes requested):
Welcome to the mlibc family :)

mintsuki (approved):

PR #1137, opened 2024-10-11
options/internal: correct setjmp/longjmp on aarch64

netbsduser:
The __mlibc_jmpbuf_register_state struct is good,
but the saving/restoring it was not.
As such, no ABI break.


no92 (approved):
LGTM

Dennisbonke (approved):
LGTM

PR #1138, opened 2024-10-11
options/posix: guard some fpu state definitions behind __x86_64__

netbsduser:
They appear to be from glibc's sys/ucontext.h, which only defines them under that condition.


Dennisbonke (commented):
Maybe add a `#warning` or `#error` for other architectures?

64:
I think this is an ok fix for now - ideally we ought to split these definitions out into separate files. Any code on non-x86 arches using fpregset will break because of this, but that seems ok because they 're using the wrong definitions anyway.

64 (approved):
Will merge shortly if no other objections

PR #1139, opened 2024-10-11
options/internal: add padding to Guard

netbsduser:
Compilers targeting the Motorola 68k generally align on a 2-byte boundary, so the guard is laid out wrongly and becomes a 6-byte structure.

### DIFF
--- a/options/internal/gcc/guard-abi.cpp
+++ b/options/internal/gcc/guard-abi.cpp
@@ -31,6 +31,8 @@ struct Guard {
 	// the first byte's meaning is fixed by the ABI.
 	// it indicates whether initialization has already been completed.
 	uint8_t complete;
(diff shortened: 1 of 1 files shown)


Dennisbonke (approved):
LGTM

options/internal/gcc/guard-abi.cpp
```diff
@@ -31,6 +31,7 @@ struct Guard {
 	// the first byte's meaning is fixed by the ABI.
 	// it indicates whether initialization has already been completed.
 	uint8_t complete;
+	uint8_t padding[3];
```
64, line 34: nit: can you add a comment here? Makes it slightly less confusing to read

netbsduser: Added.

PR #1140, opened 2024-10-12
m68k: initial port

netbsduser:
This is an initial port of mlibc to the m68k family, based on what is already working in Keyronex. Sysdeps are not included, but I have made an initial, not yet tested m68k port of the Linux sysdeps in another branch.
Some elements of the port are initially stubbed, viz. 68881-style FPU support and the setjmp family of functions.


options/internal/m68k-include/mlibc/thread.hpp
```diff
@@ -0,0 +1,24 @@
...
+#include <mlibc/internal-sysdeps.hpp>
+#include <mlibc/tcb.hpp>
+#include <bits/ensure.h>
+
+namespace mlibc {
+
+extern "C" void *__m68k_read_tp();
+
+inline Tcb *get_current_tcb() {
+	// On m68k, the TCB is below 0x7000 below the thread pointer.
```
64, line 13: nit: below repeated twice

options/internal/m68k-include/mlibc/thread.hpp
```diff
@@ -0,0 +1,24 @@
...
+	void *ptr = __m68k_read_tp();
+	return reinterpret_cast<Tcb *>((uintptr_t)ptr - 0x7000 - sizeof(Tcb));
```
64, line 15: can you use tlsOffsetFromTp?

netbsduser: It's defined elsewhere, so I followed what the aarch64 port does: https://github.com/managarm/mlibc/blob/8c6e9be5b94e9162146217772c2d3dceb2376c7f/options/internal/aarch64-include/mlibc/thread.hpp#L8-L13
https://github.com/managarm/mlibc/blob/8c6e9be5b94e9162146217772c2d3dceb2376c7f/options/rtld/generic/linker.cpp#L40

options/internal/m68k-include/mlibc/thread.hpp
netbsduser: I worded this in a very ugly way, it was supposed to convey that
```
+-----+---------------|------+
| TCB |  0x7000 bytes | more |
+-----+---------------+------+
                      ^
                      TP
```
but could benefit from being reworded.

64, line 13: I mean:
```suggestion
	// On m68k, the end of the TCB is 0x7000 below the thread pointer.
```

netbsduser: That works, amended with that.

PR #1141, opened 2024-10-13
options/posix: fix realpath(3) absolute symlink handling

Mathewnd:
The old absolute symlink handling code unconditionally ignored one character from the end of the path returned by readlink, which broke realpath on any absolute links without a trailing slash. (discovered when a ``header.h`` path became ``header.`` while compiling GCC)
This new code will ignore all trailing slashes instead.

### DIFF
--- a/tests/posix/realpath.c
+++ b/tests/posix/realpath.c
@@ -17,6 +17,8 @@ void prepare() {
 	assert(!mkdir(TEST_BASE "/dir1", S_IRWXU));
 	assert(!mkdir(TEST_BASE "/dir2", S_IRWXU));
 	assert(!symlink(TEST_BASE "/dir2/", TEST_BASE "/dir1/abs-link"));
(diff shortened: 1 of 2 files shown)


64:
Thanks! Can you add a test?

Mathewnd:
Alright, I've added a test case for both an absolute symlink without a trailing slash and for one with multiples. I also went ahead and deduplicated some of the testing code with a function macro to make it cleaner.

tests/posix/realpath.c
```diff
@@ -53,6 +59,13 @@ void signal_handler(int sig, siginfo_t *info, void *ctx) {
 	abort();
 }
 
+#define TEST_PATH(rpath, expected) \
```
64, line 62: nit: this could be a function, right?

Mathewnd: I feel like this would make it easier to know which test case failed, since from what I have seen there is no backtrace on a fail.

PR #1142, opened 2024-10-13
options/internal: endian-agnostic sigset_t functions

netbsduser:
This makes the `sigset_t` manipulation functions work on big-endian targets.


64:
I'm inclined to merge this, but not 100% satisfied with this fix: `unsigned long` might not match the definition of `sigset_t` (e.g if it's defined as a `long`, or ` uint64_t`, we're opening ourselves up to more UB or endianness issues).

I think a 'proper' fix involves either making these accessors sysdeps, or using some template hackery to deal with different definitions of sigset_t properly. But if this gets the ball rolling, let's go with it for now :)

PR #1147, opened 2024-10-13
options/internal: detect underlying type of sigset_t

### LINKED ISSUE #1143: `sigset_t` does not handle endianness properly
Currently we access `sigset_t` as if it were an array of `unsigned long`. However `sigset_t` may be defined as a `long`, or ` uint64_t`, so we're opening ourselves up to some UB or endianness issues.

I think a 'proper' fix involves either making these accessors sysdeps, or using some template hackery to deal with different definitions of sigset_t properly.

_Originally posted by @64 in https://github.com/managarm/mlibc/issues/1142#issuecomment-2408988819_

64:
Fixes #1143.

### DIFF
--- a/options/internal/generic/sigset.cpp
+++ b/options/internal/generic/sigset.cpp
@@ -3,6 +3,45 @@
 #include <errno.h>
 #include <limits.h>
 #include <string.h>
+#include <stddef.h>
+
+namespace {
+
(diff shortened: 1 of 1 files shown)


no92 (approved):
LGTM

netbsduser (approved):
No problem & works fine on m68k-linux.

PR #1154, opened 2024-10-14
tests/pthread_attr: fix trampling of the stack

netbsduser:
It was previously writing to what was likely the current stack pointer, but should have been writing to the start of the alloca()'d area.


no92 (approved):
LGTM

tests/posix/pthread_attr.c
```diff
@@ -21,7 +21,7 @@ static void *stacksize_worker(void *arg) {
 	size_t alloc_size = default_stacksize + default_stacksize/2;
 	void *area = alloca(alloc_size);
 	// If the allocated stack was not enough this will crash.
-	*(volatile int*)(area + alloc_size) = 1;
+	*(volatile int*)area = 1;
```
mintsuki, line 24: Shouldn't it perhaps be `(area + alloc_size - sizeof(int))`?

PR #1155, opened 2024-10-14
options/internal: ensure FutexLockImpl alignment to 4 bytes

netbsduser:
On m68k the default alignment of 2 bytes for a uint32_t may result in futex calls failing under some conditions, as a futex must be 4-byte aligned.


no92 (approved):
LGTM

PR #1157, opened 2024-10-14
m68k: fix for static builds

netbsduser:
CI revealed some problems with static builds of mlibc on m68k. This fixes these.

### DIFF
--- a/options/lsb/generic/tls.cpp
+++ b/options/lsb/generic/tls.cpp
@@ -2,9 +2,9 @@
 #include <mlibc/thread.hpp>
 #include <mlibc/rtld-abi.hpp>
 
-#if defined(__riscv) && defined(MLIBC_STATIC_BUILD)
-	// On RISC-V, linker optimisation is not guaranteed and so we may still get
-	// calls to this function in statically linked binaries.
+#if (defined(__riscv) || defined(__m68k__)) && defined(MLIBC_STATIC_BUILD)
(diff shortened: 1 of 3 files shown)


options/rtld/generic/main.cpp
```diff
@@ -553,7 +553,16 @@ const char *__dlapi_error() {
...
 extern "C" [[ gnu::visibility("default") ]]
 void *__dlapi_get_tls(struct __abi_tls_entry *entry) {
+#if defined(__m68k__) && MLIBC_STATIC_BUILD
+	// Static binaries sometimes call this on m68k, and being static, there are
+	// no relocations so the object field doesn't get rewritten to point to the
+	// corresponding SharedObject. But since it's guaranteed to have the id of
+	// 0x1, this suffices.
+	__ensure((uintptr_t)entry->object == 0x1);
+	return reinterpret_cast<char *>(accessDtv(executableSO)) + entry->offset;
+#else
```
64, line 563: Can this logic be moved into lsb/generic/tls.cpp, alongside an existing check for RISC-V? https://github.com/managarm/mlibc/blob/13a686b37da12d70dee27b9c6ffab4207feb4156/options/lsb/generic/tls.cpp#L5-L14

64 (changes requested):

options/rtld/generic/main.cpp
netbsduser: Thanks, have done so.

64 (approved):

PR #1159, opened 2024-10-15
treewide: make public headers compile with c89/c99/c11 pedantic errors

### LINKED ISSUE #1150: CI should check headers are c89-compatible
In the past we've accidentally committed single line comments to public headers. We should add a regression test for this.

64:
Fixes #1150. Useful as a backstop to detect unwanted C99/C11/GNUisms.

Note that we omit the check for `long long`, i.e `-Wno-long-long` as it requires us to change too many things. In practice, compiling with zero errors under C89 does not matter - neither musl nor glibc's headers compile with `-std=c89 -Werror -Wsystem-headers`.


options/internal/include/bits/types.h
```diff
@@ -325,39 +325,39 @@ typedef __SIZE_TYPE__    __mlibc_size;
 /* ---------------------------------------------------------------------------- */
 
 #if defined(__cpp_static_assert)
-#	define __MLIBC_STATIC_ASSERT(c, text) static_assert(c, text)
-#elif !defined(__cplusplus)
-#	define __MLIBC_STATIC_ASSERT(c, text) _Static_assert(c, text)
+#	define __MLIBC_STATIC_ASSERT(c, text) static_assert(c, text);
```
64, line 328: semicolon added because `__MLIBC_CHECK_TYPE(foo);` causes a 'redundant semi-colon' error in C89 mode

no92:
The ABI check is complaining about a removed internal union member, this is not an ABI break though.

abis/linux/signal.h
```diff
@@ -400,7 +400,6 @@ typedef struct __ucontext {
 	struct ucontext	*uc_link;
 	stack_t uc_stack;
 	sigset_t uc_sigmask;
-	uint8_t  __unused[1024 / 8 - sizeof(sigset_t)];
```
Geertiebear, line 403: Is this not an ABI break? I guess my real question is: is `sizeof(sigset_t)` always 128? On all architectures?

options/linux/include/sys/sysinfo.h
```diff
@@ -18,7 +20,11 @@ struct sysinfo {
 	unsigned long totalhigh;
 	unsigned long freehigh;
 	unsigned int mem_unit;
+
+	/* This is how the kernel header defines it, so suppress the warning. */
+#if !defined(__MLIBC_PEDANTIC_HEADER_CHECKER)
```
Geertiebear, line 25: Is the libc still conforming if this check fails?

abis/linux/signal.h
64: On Linux, yes currently it's always 128:
https://github.com/managarm/mlibc/blob/0f7479f4645d36897253235c40b134f49ba989f6/abis/linux/signal.h#L123-L125

Perhaps I should suppress the check here just to be safe? Or add some kind of assert?

options/linux/include/sys/sysinfo.h
64: I couldn't come up with a better idea. AFAIK this feature shouldn't be gated behind `_GNU_SOURCE`?

abis/linux/signal.h
Geertiebear: Maybe add a static assert to check the alignment of `struct __ucontext`, since that seems to be the purpose of the padding.

options/internal/include/bits/types.h
avdgrinten: I think this is the wrong fix. The trailing semicolon should be at the call site.

options/linux/include/sys/inotify.h
```diff
@@ -43,7 +44,14 @@ struct inotify_event {
...
 	unsigned int cookie;
 	unsigned int len;
+
+/*
+ * glibc uses a flexible array member here, but we get a warning and they don't:
+ * https://gcc.gnu.org/bugzilla/show_bug.cgi?id=117241
+ */
+#if !defined(__MLIBC_PEDANTIC_HEADER_CHECKER)
 	char name[];
+#endif
```
avdgrinten, line 54: Can't we use some `#pragma GCC diagnostic error` here to suppress this instead of a macro?

abis/linux/signal.h
64: I've kept it but suppressed it with a pragma

options/linux/include/sys/inotify.h
64: Done

options/linux/include/sys/sysinfo.h
64: added pragma

Dennisbonke (approved):
LGTM

PR #1160, opened 2024-10-15
ci: set warnlevel to 2

64:


64:
Unnecessary - warning_level=2 is default

PR #1163, opened 2024-10-15
rtld: invoke fini functions

64:
#1092 was previously reverted because it caused a regression. In particular, it relied on DSOs to invoke `__cxa_finalize`, but mlibc itself is missing a call to `__cxa_finalize` in crtbegin, so any libc-related destructors (such as the guard that flushes all open files) were not run.

Fix this by adding a `[[gnu::destructor]]` function which calls `__cxa_finalize` with our own `__dso_handle`.

Also, while refactoring, I noticed that the `crt{begin,end}.S` are no longer needed (they were used for `.ctor`/`.dtor` handling, which we no longer use), and provided the definition of `__dso_handle` (which I've moved into C++).

Also add a regression test to ensure we run libc destructors somehow.


options/rtld/generic/linker.cpp
```diff
@@ -927,9 +956,22 @@ void doInitialize(SharedObject *object) {
 	__ensure(object->wasLinked);
 	__ensure(!object->wasInitialized);
 
-	// if the object has dependencies we initialize them first
-	for(size_t i = 0; i < object->dependencies.size(); i++)
-		__ensure(object->dependencies[i]->wasInitialized);
+	// If the object has dependencies we initialize them first
```
Geertiebear, line 959: Is this (and down below) really the place we should be checking for circular dependencies? Seems a bit annoying and violates DRY. Isn't it better to consider this an invariant of loaded objects, and validate at object construction time the invariant holds?

options/rtld/generic/linker.hpp
```diff
@@ -86,6 +89,9 @@ struct ObjectRepository {
 
 	frg::hash_map<frg::string_view, SharedObject *,
 			frg::hash<frg::string_view>, MemoryAllocator> _nameMap;
+
+	// Used for destructing the objects, stores all the objects in the order they are initialized.
+	frg::vector<SharedObject *, MemoryAllocator> _destructQueue;
```
Geertiebear, line 94: Nitpick but you could use a ``frg::stack`` here since destruction happens in reverse order anyway.

options/rtld/generic/linker.cpp
64: This is part of #1092

avdgrinten:
I'd prefer to _not_ support circular dependencies as these are quite broken anyway.

Geertiebear:
Agreed, I misread the code and thought it was ensuring there were no circular dependencies, but actually this code is adding explicit support for it. This PR should keep the old code.

options/rtld/generic/linker.cpp
64: Ok, done

64:
I've removed handling of circular deps and removed crt{begin,end}.S so that we no longer support `ctor`/`dtor` (at least, not as part of mlibc - we should be able to load programs compiled with these just fine).

Dennisbonke (approved):
LGTM

PR #1164, opened 2024-10-16
misc: Remove crypt/iconv/intl stubs

mintsuki:

PR #1165, opened 2024-10-16
meson: replace double-negative options

64:
`disable_foo` with default false is more confusing than `foo` with default true. So rename all such options.

Breaking change for people building from master - but they shouldn't be doing that anyway ;)


mintsuki (approved):

mintsuki:
LGTM

no92:
This is a good candidate for PR-ing against abi-break, not master.

64:
This isn't an abi-break though?

no92:
It isn't in the strict sense, however this would break the build setups of pretty much everyone without bumping the version as-is, so the procedure should be pretty similar IMO.

mintsuki:
> It isn't in the strict sense, however this would break the build setups of pretty much everyone without bumping the version as-is, so the procedure should be pretty similar IMO.

We already did that just a few commits ago and nothing happened. That's because no one should track master, and I'd argue breaking their setups is a feature, not a bug.

ElectrodeYT:
> Breaking change for people building from master - but they shouldn't be doing that anyway ;)

Shouldn't; yes; but even managarm does, so this would break basically everyone who doesn't instantly pull changes into `bootstrap-managarm`, so this should go into `abi-break`

no92:
Changing build options in a non-backwards compatible way without even as much as a new tag is immensely bad.

Geertiebear:
But this doesn't require a rebuild the world? I don't see how this is comparable to an abi-break.

mintsuki:
> Changing build options in a non-backwards compatible way without even as much as a new tag is immensely bad.

Hard disagree, this is `master`.

ElectrodeYT:
> > Changing build options in a non-backwards compatible way without even as much as a new tag is immensely bad.
> 
> Hard disagree, this is `master`.

At minimum this requires timing with a bootstrap PR too

Geertiebear (approved):
LGTM, very welcome change.

mintsuki:
> > > Changing build options in a non-backwards compatible way without even as much as a new tag is immensely bad.
> > 
> > 
> > Hard disagree, this is `master`.
> 
> At minimum this requires timing with a bootstrap PR too

If mlibc has to be a portable libc that isn't tied to managarm behind it having a port for it, PRs to mlibc shouldn't be blocked over managarm PRs.

ElectrodeYT (changes requested):
Needs a bootstrap PR to change the options there, then we can consider merging this

ElectrodeYT:
> > > > Changing build options in a non-backwards compatible way without even as much as a new tag is immensely bad.
> > > 
> > > 
> > > Hard disagree, this is `master`.
> > 
> > 
> > At minimum this requires timing with a bootstrap PR too
> 
> If mlibc has to be a portable libc that isn't tied to managarm behind it having a port for it, PRs to mlibc shouldn't be blocked over managarm PRs.

It is a portable libc, yes, however it is also the `managarm` libc, and, currently, the mlibc and mlibc-headers packages on managarm are broken due to #1164

mintsuki:
managarm's CI seems to disagree with you?

no92:
> But this doesn't require a rebuild the world? I don't see how this is comparable to an abi-break.

The point isn't about rebuilding world, but about notifying downstream users about changes that will very likely impact them, right? Changing core build options deserves to be noted in a changelog, preferably in a new release.

no92:
> managarm's CI seems to disagree with you?

The build is broken when built from current master, that is with #1164 + the commit afterwards.

mintsuki:
What is the error, and what is "the commit afterwards"?

mintsuki:
Even if we need a policy for how to deal with changes like this, there is precedent that this has already been done before and merged directly to `master`. Plus this PR has 2 approvals, including Geert's.

no92:
Ah apparently commit `0f0389d1da9b3aba22ff7e3dcf025b4bc97bc5db` on bootstrap-managarm removes the use of the function, another instance of how communication around this has been suboptimal.

ElectrodeYT:
> What is the error, and what is "the commit afterwards"?

I retract what I said about it currently being broken, it was a skill issue on my end, I still want to see a bootstrap PR to accompany this change to mlibc though

mintsuki:
sigh

PR #1166, opened 2024-10-17
meson: use features to enable options

64:
This allows sysdeps to declare X option is not supported, which means we can provide an error to the user if they enable an option on a system that doesn't support it.


mintsuki:
LGTM

mintsuki (approved):

ElectrodeYT (approved):
LGTM

Dennisbonke (approved):
LGTM

PR #1170, opened 2024-10-17
rtld: fix relr

### LINKED ISSUE #1169: RELR is untested

Qwinci:
Fix relr properly now and add a test, the previous fix was not correct. The first bit of a relr entry should be discarded in the bitmap case as it just tells that it's a bitmap and is not supposed to be used for the relocation calculation (though using it like that happened to work with the previous fix's changes as long as there was only one relr bitmap entry per address).

Fixes #1169.


tests/rtld/meson.build
```diff
@@ -30,6 +31,10 @@ foreach test_name : rtld_test_cases
 	test_native_depends = []
 	test_additional_link_args = []
 
+	if test_name == 'relr'
+		test_additional_link_args += ['-Wl,-z,pack-relative-relocs']
+	endif
```
64, line 36: put this in the relr/meson.build

Qwinci: Done.

mintsuki (approved):

Dennisbonke (approved):
LGTM

PR #1180, opened 2024-10-25
ci: consolidate misc workflows

64:


Geertiebear (approved):
LGTM, merge should be paired with a change in the repo config so ill leave it to either @qookei or @avdgrinten to press the big merge button.

PR #1181, opened 2024-10-26
options/internal: fix static asserts for pre-C++03

no92:
This commit correctly checks whether the C++ standard is new enough, and if so, uses the standard-defined `static_assert`. In the case of C, gcc and clang implement `_Static_assert` as an extension irrespective of the chosen C standard. Otherwise, we fall back to not checking the condition. We need to do the `extern int` hack in order to not produce errors in pedantic mode, as a single semicolon or `(void)0` at file scope give warnings or errors then.

This was originally discovered when compiling libstdc++, which uses `-std=gnu++98` and was throwing errors.

### DIFF
--- a/options/internal/include/bits/types.h
+++ b/options/internal/include/bits/types.h
@@ -324,11 +324,13 @@ typedef __SIZE_TYPE__    __mlibc_size;
 /* Sanity checking. Make sure that we agree with the compiler's ABI. */
 /* ---------------------------------------------------------------------------- */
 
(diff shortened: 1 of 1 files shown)


64:
Can we try to enable C++ header checks in CI?

avdgrinten (approved):

avdgrinten:
Merging this since it unbreaks the xbbs build.

64:
Automated check in #1184

PR #1183, opened 2024-11-03
rtld: Ignore DT_BIND_NOW for ld.so

marv7000:
ld.so can contain an `DT_BIND_NOW` entry if mlibc is compiled with `-Wl,-z,now`.

### DIFF
--- a/options/rtld/generic/main.cpp
+++ b/options/rtld/generic/main.cpp
@@ -345,6 +345,7 @@ extern "C" void *interpreterMain(uintptr_t *entry_stack) {
 		case DT_RELRSZ:
 		case DT_RELRENT:
 		case DT_PLTGOT:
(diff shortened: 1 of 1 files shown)


qookei (approved):

no92 (approved):
LGTM as we don't support having ld.so have `DT_NEEDED`.

I do however wonder whether encountering `DT_NEEDED` here should explicitly fail, as opposed to having it (somewhat implicitly) fall into the default case.

64:
I don't entirely see what DT_NEEDED has to do with DT_BIND_NOW? The latter is about eager relocation processing, the former is about dynamic dependencies?

64:
I think this is correct, we immediately relocate ourselves before entering ld.so so should be no problem.

PR #1190, opened 2024-11-24
options/posix: add bound checks for dns response parsing

48cf:


no92:
It would be nice to have some testcases for the issue :^)

Dennisbonke (approved):
LGTM

PR #1197, opened 2024-12-13
Implement various syscalls on Linux for systemd

Dennisbonke:
Part of the systemd on Linux project.


options/linux/include/mlibc/linux-sysdeps.hpp
```diff
@@ -83,6 +83,12 @@ int sys_ioctl(int fd, unsigned long request, void *arg, int *result);
 
 [[gnu::weak]] int sys_getifaddrs(struct ifaddrs **);
 
+[[gnu::weak]] int sys_sendfile(int outfd, int infd, off_t *offset, size_t count, size_t *out);
```
no92, line 86: The use of `off_t` introduces compilation errors, this is missing `#include <bits/off_t.h>`.

PR #1202, opened 2024-12-13
options/posix: Add support for IPv6 addresses in inet_pton

Dennisbonke:
@48cf can you add the tests you had for this to this (or a separate) PR please?

Part of the systemd on Linux project.

### DIFF
--- a/options/posix/generic/arpa-inet.cpp
+++ b/options/posix/generic/arpa-inet.cpp
@@ -181,6 +181,7 @@ const char *inet_ntop(int af, const void *__restrict src, char *__restrict dst,
 	errno = ENOSPC;
 	return NULL;
 }
+
 int inet_pton(int af, const char *__restrict src, void *__restrict dst) {
 	switch (af) {
 		case AF_INET: {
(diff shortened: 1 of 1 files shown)


options/posix/generic/arpa-inet.cpp
```diff
@@ -200,9 +201,89 @@ int inet_pton(int af, const char *__restrict src, void *__restrict dst) {
...
+			}
+
+			if(*src) {
+				return 0;
+			}
+
+			auto addr6 = reinterpret_cast<struct in6_addr*>(dst);
+
+			for(int i = 0; i < 8; i++) {
+				addr6->s6_addr16[i] = addr[i];
```
Geertiebear, line 282: Can we not just work directly into this array?

options/posix/generic/arpa-inet.cpp
```diff
@@ -200,9 +201,89 @@ int inet_pton(int af, const char *__restrict src, void *__restrict dst) {
...
+
+				addr[addr_len++] = htons(value);
+			}
+
+			if(abbrev) {
+				uint16_t array[8] = {};
+				int array_len = 0;
+
+				src = abbrev + 2;
+				while(*src) {
```
Geertiebear, line 245: Why is this case basically copy pasted from the above one? Can they not be merged?

options/posix/generic/arpa-inet.cpp
```diff
@@ -200,9 +201,89 @@ int inet_pton(int af, const char *__restrict src, void *__restrict dst) {
...
+
+				if(addr_len >= 8) {
+					return 0;
+				}
+
+				addr[addr_len++] = htons(value);
+			}
+
+			if(abbrev) {
+				uint16_t array[8] = {};
```
Geertiebear, line 241: This seems broken, the array this block writes into goes out of scope before we can write it into the user provided argument.

Geertiebear (changes requested):

PR #1209, opened 2024-12-27
Symbol versioning support

### LINKED ISSUE #1208: Symbol versioning is unsupported
This breaks libpci and lspci for example, because we end up calling the older version of `pci_filter_init` (`pci_filter_init@LIBPCI_3.0`) and the newer version of `pci_filter_match` (`pci_filter_match@LIBPCI_3.8`), which causes some devices to be filtered out incorrectly. I assume which versions get called depends on the symbol hash table.

qookei:
There's some minor TODOs left in the code (handling weak versions and figuring out what to do about VERDEF dependency names), but this is sufficient to fix #1208, and the TODOs aren't a problem when booting Managarm (which has plenty of libraries with versioned symbols).

Closes #1208.

### DIFF
--- a/options/rtld/generic/linker.cpp
+++ b/options/rtld/generic/linker.cpp
@@ -120,9 +120,11 @@ SharedObject *ObjectRepository::injectObjectFromDts(frg::string_view name,
 	object->baseAddress = base_address;
 	object->dynamic = dynamic;
 	_parseDynamic(object);
+	_parseVerdef(object);
 
 	_addLoadedObject(object);
 	_discoverDependencies(object, globalScope.get(), rts);
+	_parseVerneed(object);
 
 	return object;
 }
@@ -137,9 +139,11 @@ SharedObject *ObjectRepository::injectObjectFromPhdrs(frg::string_view name,
 		name.data(), std::move(path), true, globalScope.get(), rts);
 	_fetchFromPhdrs(object, phdr_pointer, phdr_entry_size, num_phdrs, entry_pointer);
 	_parseDynamic(object);
+	_parseVerdef(object);
 
 	_addLoadedObject(object);
 	_discoverDependencies(object, globalScope.get(), rts);
+	_parseVerneed(object);
 
 	return object;
 }
@@ -279,9 +283,11 @@ frg::expected<LinkerError, SharedObject *> ObjectRepository::requestObjectWithNa
 	}
 
 	_parseDynamic(object);
+	_parseVerdef(object);
 
 	_addLoadedObject(object);
 	_discoverDependencies(object, localScope, rts);
+	_parseVerneed(object);
 
 	return object;
 }
@@ -324,9 +330,11 @@ frg::expected<LinkerError, SharedObject *> ObjectRepository::requestObjectAtPath
 	}
 
 	_parseDynamic(object);
+	_parseVerdef(object);
 
 	_addLoadedObject(object);
 	_discoverDependencies(object, localScope, rts);
+	_parseVerneed(object);
 
 	return object;
 }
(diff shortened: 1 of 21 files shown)


options/rtld/generic/linker.cpp
```diff
@@ -795,6 +817,167 @@ void ObjectRepository::_parseDynamic(SharedObject *object) {
 	}
 }
 
+void ObjectRepository::_parseVerdef(SharedObject *object) {
+	if(!object->versionDefinitionTableOffset) {
+		if(verbose)
+			mlibc::infoLogger()
+				<< "mlibc: Object " << object->name
+				<< " defines no versions" << frg::endlog;
+		return;
```
Geertiebear, line 826: What happens in this case? Is an object still able to be resolved even if it doesn't have a DT_VERDEF?

qookei: If an object defines no versions but was supposed to, the dependent will error out in `_parseVerneed` on line 965.

qookei: If it defines no version and no one wants any versions from it, nothing special happens.

options/rtld/generic/linker.cpp
```diff
@@ -795,6 +816,163 @@ void ObjectRepository::_parseDynamic(SharedObject *object) {
...
+		__ensure(!(def.vd_flags & ~VER_FLG_BASE));
+
+		// TODO(qookie): Figure out what to do about VERDEF dependency names.
+		// They don't make much sense to me, and it seems to work if we just ignore them.
+		elf_verdaux aux;
+		memcpy(&aux, reinterpret_cast<void *>(address + def.vd_aux), sizeof(elf_verdaux));
+
+		const char *name =
+			static_cast<const char *>(
+				reinterpret_cast<void *>(
```
64, line 856: Can't you reinterpret_cast directly to `const char *`?

options/rtld/generic/linker.cpp
```diff
@@ -795,6 +816,163 @@ void ObjectRepository::_parseDynamic(SharedObject *object) {
...
+
+	for(size_t i = 0; i < object->versionRequirementCount; i++) {
+		elf_verneed need;
+		memcpy(&need, reinterpret_cast<void *>(address), sizeof(elf_verneed));
+
+		// Required by spec.
+		__ensure(need.vn_version == 1);
+
+		frg::string_view file =
+			static_cast<const char *>(
```
64, line 902: likewise

options/rtld/generic/linker.cpp
```diff
@@ -795,6 +816,163 @@ void ObjectRepository::_parseDynamic(SharedObject *object) {
...
+					+ object->stringTableOffset + need.vn_file));
+
+		// Figure out the target object from file
+		SharedObject *target = nullptr;
+		for(auto dep : object->dependencies) {
+			if(verbose)
+				mlibc::infoLogger()
+					<< "mlibc: Trying " << dep->name << " (SONAME: "
+					<< dep->soName << ") to satisfy " << file << frg::endlog;
+			if(dep->name == file || (dep->soName && dep->soName == file)) {
```
64, line 914: Looking at `findLoadedObject`, we _only_ compare against SONAME. Is the `dep->name == file` needed? Not a big deal but it would be good to understand what the correct behaviour is in both cases

options/rtld/generic/linker.cpp
```diff
@@ -795,6 +816,163 @@ void ObjectRepository::_parseDynamic(SharedObject *object) {
...
+				<< " version(s) from DSO "
+				<< file << frg::endlog;
+
+		uintptr_t auxAddr = address + need.vn_aux;
+		for(size_t j = 0; j < need.vn_cnt; j++) {
+			elf_vernaux aux;
+			memcpy(&aux, reinterpret_cast<void *>(auxAddr), sizeof(elf_vernaux));
+
+			const char *name =
+				static_cast<const char *>(
```
64, line 942: likewise

options/rtld/generic/linker.cpp
```diff
@@ -1044,16 +1049,74 @@ SharedObject::SharedObject(const char *name, const char *path,
 			frg::string<MemoryAllocator> { path, getAllocator() },
 			is_main_object, localScope, object_rts) {}
 
+frg::tuple<ObjectSymbol, SymbolVersion> SharedObject::getSymbolByIndex(size_t index) {
+	SymbolVersion ver{1}; // If we don't have any version information, treat all symbols as global.
+	ObjectSymbol sym{
+		this,
+		reinterpret_cast<elf_sym *>(
```
64, line 1056: likewise (multiple reinterpret_cast)

64 (commented):
LGTM with a few nits

options/rtld/generic/linker.cpp
qookei: That's true. For some reason I thought you could only `reinterpret_cast` integers into `void *`...

64: I believe that's true of `static_cast`

options/rtld/generic/linker.cpp
qookei: glibc matches both the SONAME and the file name of the object (although in the opposite order compared us AFAICU?). Without checking the file name our test case doesn't pass, since it doesn't have a SONAME (glibc loads it fine).

64 (approved):

Dennisbonke (approved):
LGTM

PR #1210, opened 2024-12-28
sysdeps: simplify entry point

64:
Move some initialisation tasks to gnu::constructor functions defined inside mlibc.


Geertiebear (approved):
LGTM

PR #1214, opened 2024-12-31
options/glibc: Add __riscv_flush_icache()

avdgrinten:
This is a glibc extension: https://www.gnu.org/software/libc/manual/html_node/RISC_002dV.html


Dennisbonke (approved):
LGTM

PR #1217, opened 2025-01-09
Implement various syscalls on Linux for systemd

Dennisbonke:


avdgrinten:
As mentioned on Discord, the clone patch is not ready for inclusion -- clone() needs to integrate with the pthreads code, otherwise things will break horribly with `CLONE_VM` is not passed.

options/posix/include/mlibc/posix-sysdeps.hpp
```diff
@@ -236,6 +236,9 @@ int sys_vm_unmap(void *pointer, size_t size);
 
 [[gnu::weak]] int sys_waitid(idtype_t idtype, id_t id, siginfo_t *info, int options);
 
+[[gnu::weak]] int sys_name_to_handle_at(int dirfd, const char *pathname, struct file_handle *handle, int *mount_id, int flags);
+[[gnu::weak]] int sys_splice(int in_fd, off_t *in_off, int out_fd, off_t *out_off, size_t size, unsigned int flags, size_t *out);
```
avdgrinten, line 240: These are Linux extensions, not POSIX. Do we still keep them in this header?

options/linux/include/mlibc/linux-sysdeps.hpp
```diff
@@ -83,6 +84,12 @@ int sys_ioctl(int fd, unsigned long request, void *arg, int *result);
 
 [[gnu::weak]] int sys_getifaddrs(struct ifaddrs **);
 
+[[gnu::weak]] int sys_sendfile(int outfd, int infd, off_t *offset, size_t count, size_t *out);
```
avdgrinten, line 87: `sendfile()` returns `ssize_t`, not `size_t`. However, the value is always positive, so passing `size_t *` is probably okay. I'm not sure how we handle this for other syscalls, but if it's consistent with other calls, it can stay like this.

The same applies to `sys_splice()`.

avdgrinten (commented):
LGTM except for the clone() patch.

options/linux/include/mlibc/linux-sysdeps.hpp
Dennisbonke: Fixed this at both places.

Dennisbonke:
I dropped the clone patch as requested.

no92 (approved):
LGTM

PR #1218, opened 2025-01-12
lsb: don't provide __dso_handle for static builds

lzcunt:
In static builds, __dso_handle should be provided by the crtbegin.o/crtend.o that is linked into the executable. This patch also fixes failing link tests when building a static gcc targeting mlibc.

### DIFF
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -146,6 +146,7 @@ extra_cflags_test_cases = {
 }
 
 test_sources = []
+test_objects = []
 test_link_args = []
 test_c_args = ['-D_GNU_SOURCE', '-fno-builtin']
 use_pie = false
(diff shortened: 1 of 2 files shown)


tests/meson.build
```diff
@@ -167,6 +168,39 @@ endif
...
+			crtpath = path
+			break
+		endif
+	endforeach
+
+	if crtpath == ''
+		error('could not find crtbegin.o/crtend.o')
+	endif
+	test_objects += [crtpath / 'crtbegin.o', crtpath / 'crtend.o']
+
```
64, line 203: there is test_sources += crt below, maybe remove that?

avdgrinten: Yes, we still need to keep `crt` for `crt1.o`.

64 (approved):

PR #1220, opened 2025-01-13
systemd-udevd on managarm: part 2

no92:


options/glibc/generic/malloc.cpp
```diff
@@ -4,3 +4,8 @@
 size_t malloc_usable_size(void *p) {
 	return getAllocator().get_size(p);
 }
+
+int malloc_info(int, FILE *) {
+	errno = ENOSYS;
```
Dennisbonke, line 9: Shouldn't we just make this a sysdep instead?

options/posix/generic/fcntl.cpp
```diff
@@ -69,8 +69,8 @@ int posix_fallocate(int fd, off_t offset, off_t size) {
 
 // This is a linux extension
 int name_to_handle_at(int, const char *, struct file_handle *, int *, int) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+	errno = ENOSYS;
```
Dennisbonke, line 72: Same here, we should make this a sysdep. Also, there's an open PR that implements this for Linux, so we could drop this commit and piggy back off of that PR.

options/posix/include/netinet/ip.h
```diff
@@ -68,6 +68,8 @@ struct iphdr {
 	uint32_t daddr;
 };
 
+#define	IPTOS_DSCP_EF 0xB8
```
Dennisbonke, line 71: Style issue, two spaces here?

options/posix/generic/fcntl.cpp
avdgrinten: +1, Dennis' PR already turns this into a sysdep.

avdgrinten (approved):

PR #1233, opened 2025-01-26
options/ansi: provide _setjmp and _longjmp in setjmp.h

lzcunt:
This patch makes setjmp.h better conform to both ISO C and POSIX-2017.1 standards.

setjmp/_setjmp/longjmp/_longjmp currently act how they act in glibc systems (they both don't save the signal mask), if there is a use case for it we could also support the BSD convention (where setjmp/longjmp save the signal mask but the underscore variants don't)

### DIFF
--- a/options/ansi/include/setjmp.h
+++ b/options/ansi/include/setjmp.h
@@ -21,6 +21,15 @@ typedef struct __jmp_buf {
 __attribute__((__returns_twice__)) int setjmp(jmp_buf __buffer);
 __attribute__((__noreturn__)) void longjmp(jmp_buf __buffer, int __value);
 
+/* setjmp is defined as a function macro in the ISO C standard */
+#define setjmp(env) setjmp(env)
+
+#if __MLIBC_POSIX_OPTION
+__attribute__((__returns_twice__)) int _setjmp(jmp_buf __buffer);
+/* POSIX-2017.1 says _longjmp shall be declared as a function */
+__attribute__((__noreturn__)) void _longjmp(jmp_buf __buffer, int __value);
+#endif /* __MLIBC_POSIX_OPTION */
+
 #endif /* !__MLIBC_ABI_ONLY */
 
 /* POSIX Non-local jumps signal extensions */
--- a/options/internal/aarch64/setjmp.S
+++ b/options/internal/aarch64/setjmp.S
@@ -27,7 +27,10 @@ __setjmp:
 
 .global setjmp
 .type setjmp, "function"
+.global _setjmp
+.type _setjmp, "function"
 setjmp:
+_setjmp:
 	mov x2, xzr
 	b __setjmp
 
@@ -39,7 +42,10 @@ sigsetjmp:
 
 .global longjmp
 .type longjmp, "function"
+.global _longjmp
+.type _longjmp, "function"
 longjmp:
+_longjmp:
 	ldp x19, x20, [x0, #0]
 	ldp x21, x22, [x0, #16]
 	ldp x23, x24, [x0, #32]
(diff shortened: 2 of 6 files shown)


lzcunt:
I will fix the ABI break soon™

no92 (approved):
LGTM

options/ansi/include/setjmp.h
```diff
@@ -21,6 +21,15 @@ typedef struct __jmp_buf {
 __attribute__((__returns_twice__)) int setjmp(jmp_buf __buffer);
 __attribute__((__noreturn__)) void longjmp(jmp_buf __buffer, int __value);
 
+/* setjmp is defined as a function macro in the ISO C standard */
+#define setjmp(env) setjmp(env)
```
mintsuki, line 25: Why do this when `setjmp` could just be a macro that expands to `_setjmp`? Is `longjmp` likewise also not meant to be a macro that expands to `_longjmp`?

mintsuki (changes requested):

options/ansi/include/setjmp.h
mintsuki: And that way `setjmp`/`longjmp` don't need to be anything other than macros, so the prototypes and actual symbols can be removed, as `_setjmp`/`_longjmp` are enough.

lzcunt: POSIX-2017.1 requires both _longjmp and longjmp to be functions with function prototypes, and we can't remove the setjmp symbol because that would be an ABI break (existing binaries that use the setjmp symbol need it), and IMO it makes sense to have a setjmp function when we also have a longjmp function. Both glibc and musl do the same AFAIK

mintsuki: We can have ABI breaks, we just need to retarget this PR towards the abi-break branch. Also I am not super comfortable with shadowing a function with a macro, but I am sure that has been done elsewhere, not that whataboutism should be a justification for it.

Is the C standard clear about whether a function-like macro or a normal macro is to be used for this?

lzcunt: I don't have a copy of the C standard, but several sources say it is a function macro and IIRC we had it confirmed by someone with a copy of the ISO C standard that it is a function macro. musl seems to get away with it being a normal macro, glibc defines it as a function macro.

mintsuki: Also where can I find a reference that "glibc systems" do not have the difference regarding whether signal masks are saved or not between the underscore/no underscore variants? The comments in the glibc headers don't seem to suggest that.

lzcunt: Quoting the setjmp linux man page:

       POSIX does not specify whether setjmp() will save the signal mask
       (to be later restored during longjmp()).  In System V it will not.
       In 4.3BSD it will, and there is a function _setjmp() that will
       not.  The behavior under Linux depends on the glibc version and
       the setting of feature test macros.  Before glibc 2.19, setjmp()
       follows the System V behavior by default, but the BSD behavior is
       provided if the _BSD_SOURCE feature test macro is explicitly
       defined and none of _POSIX_SOURCE, _POSIX_C_SOURCE, _XOPEN_SOURCE,
       _GNU_SOURCE, or _SVID_SOURCE is defined.  Since glibc 2.19,
       <setjmp.h> exposes only the System V version of setjmp().
       Programs that need the BSD semantics should replace calls to
       setjmp() with calls to sigsetjmp() with a nonzero savesigs
       argument.

mintsuki: Alright, thanks for clearing up my misunderstandings!

mintsuki (approved):

PR #1234, opened 2025-01-27
options/rtld: Set up early TCB to cache TID

avdgrinten:
This avoids lots of unnecessary syscalls to obtain the thread ID. In particular, the TID is needed in futex locks and we need a syscall to obtain it before the TCB is set up.


no92 (approved):
LGTM

PR #1236, opened 2025-01-28
various fixes for failing GMP tests

### LINKED ISSUE #1221: GMP tests for `mpz_inp_str` fail
Currently, all GMP test pass except these two:

tests/mpz/io.c:
```
mpz_inp_str nread doesn't match ftell
  nread  5
  ftell  6
FAIL io (exit status: 134)
```

tests/mpz/t-inp_str.c:
```
mpz_inp_str nread wrong
  inp          "0"
  base         10
  pre          0
  post         1
  got_nread    1
  ftell_nread  2
FAIL t-inp_str (exit status: 134)
```

Seemingly there is an off-by-one somewhere that causes `ftell` or `mpz_inp_str`'s read bytes to be wrong.

lzcunt:
Fixes ftell behaviour after ungetc, and adds a simple test for ftell's functionality which also acts as a simple reproducible test case for the failing GMP tests documented in #1221. This test used to fail because our behaviour when checking ftell after a ungetc didn't match glibc.

Adds support for RADIXCHAR and THOUSEP in nl_langinfo, and adds some glibc aliases (not sure if they should be gated or available for all posix targets)

Fixes empty string handling in fputs and adds a simple test for fputs.

Fixes #1221.

### DIFF
--- a/options/ansi/generic/stdio.cpp
+++ b/options/ansi/generic/stdio.cpp
@@ -934,7 +934,10 @@ int fputc(int c, FILE *stream) {
 }
 
 int fputs_unlocked(const char *__restrict string, FILE *__restrict stream) {
-	if(fwrite_unlocked(string, strlen(string), 1, stream) != 1)
+	// fwrite with a length of 0 will return 0, so we need to explicitly allow
+	// zero length strings.
+	size_t length = strlen(string);
+	if (length != 0 && fwrite_unlocked(string, length, 1, stream) != 1)
 		return EOF;
 	return 1;
 }
--- a/options/ansi/generic/file-io.cpp
+++ b/options/ansi/generic/file-io.cpp
@@ -292,7 +292,9 @@ int abstract_file::tell(off_t *current_offset) {
 	if(int e = io_seek(0, SEEK_CUR, &seek_offset); e)
 		return e;
 
-	*current_offset = seek_offset + (off_t(__offset) - off_t(__io_offset));
+	*current_offset = seek_offset
+		+ (off_t(__offset) - off_t(__io_offset))
+		+ (off_t(__unget_ptr) - off_t(__buffer_ptr));
 	return 0;
 }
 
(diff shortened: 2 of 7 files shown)


lzcunt:
All GMP tests pass when GMP is patched with this tiny patch.
```patch
--- a/tests/cxx/clocale.c
+++ b/tests/cxx/clocale.c
@@ -41,7 +41,7 @@ the GNU MP Library test suite.  If not, see https://www.gnu.org/licenses/.  */
 
 extern char point_string[];
 
-#if HAVE_LOCALECONV && ! defined __MINGW32__
+#if HAVE_LOCALECONV && ! defined __MINGW32__ && 0
 struct lconv *
 localeconv (void)
 #if defined __cplusplus && defined __GLIBC__
@@ -54,7 +54,7 @@ localeconv (void)
 }
 #endif
 
-#if HAVE_NL_LANGINFO
+#if HAVE_NL_LANGINFO && 0
 char *
 nl_langinfo (nl_item n)
 #if defined __cplusplus && defined __GLIBC__
```

options/ansi/generic/stdio.cpp
```diff
@@ -934,7 +934,8 @@ int fputc(int c, FILE *stream) {
 }
 
 int fputs_unlocked(const char *__restrict string, FILE *__restrict stream) {
-	if(fwrite_unlocked(string, strlen(string), 1, stream) != 1)
+	size_t length = strlen(string);
```
no92, line 937: It took me a while to see it, but the reason why this is functionally different from the previous state is that `fwrite` with a `length` (or nitems) of zero will return 0. It might be nice to explain that in a comment.

Nit: the commit message should start with `options/ansi` instead of `ansi`.

lzcunt: My reasoning for starting the commit message with `ansi:` was that it also touched tests, but I understand that the main change in that commit is in options so I'll change that accordingly.

mintsuki (approved):

Issue #1196, opened 2024-12-05
Invalid RPATH handling for dependencies that are found through RPATH


48cf:
systemd shared libraries and binaries are installed under /usr/lib/systemd, while binaries do have RPATH set the shared libraries don't and the dynamic linker fails to find other systemd libraries when they depend on each other
```
[    1.579582] Run /usr/sbin/init as init process
Could not satisfy dependency libsystemd-shared-256.so
```
```
czapek@raptor-wsl:~/src/linux-mlibc$ readelf -d build/system-root/usr/lib/systemd/systemd

Dynamic section at offset 0x17488 contains 31 entries:
  Tag        Type                         Name/Value
 0x0000000000000001 (NEEDED)             Shared library: [libsystemd-core-256.so]
 0x0000000000000001 (NEEDED)             Shared library: [libsystemd-shared-256.so]
 0x0000000000000001 (NEEDED)             Shared library: [libssp.so.0]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so]
 0x000000000000000f (RPATH)              Library rpath: [/usr/lib/systemd]
czapek@raptor-wsl:~/src/linux-mlibc$ readelf -d build/system-root/usr/lib/systemd/libsystemd-core-256.so

Dynamic section at offset 0x20d158 contains 33 entries:
  Tag        Type                         Name/Value
 0x0000000000000001 (NEEDED)             Shared library: [libsystemd-shared-256.so]
 0x0000000000000001 (NEEDED)             Shared library: [libmount.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libssp.so.0]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so]
 0x000000000000000e (SONAME)             Library soname: [libsystemd-core-256.so]
```

PR #1238, opened 2025-02-01
rtld: Use breadth-first search order

lzcunt:
Related to #1196, might fix the systemd issue but doesn't actually implement RPATH inheritance, simply makes RTLD search breadth-first. Will fix the `libgcc_s.so.1` issue in nixpkgs.


avdgrinten (approved):
LGTM but I didn't test this in practice.

lzcunt:
We should get someone to test this with systemd on linux before merging

streaksu:
Seems to work fine at Gloire, as far as the distro requires in terms of dynamic library loading

PR #1242, opened 2025-02-13
options/glibc: stub malloc_info

Dennisbonke:


Dennisbonke:
We follow musl and don't add this function.

PR #1243, opened 2025-02-13
sysdeps/managarm: Implement sys_statx

Dennisbonke:
Blocked on managarm/managarm#846

### DIFF
--- a/sysdeps/managarm/generic/file.cpp
+++ b/sysdeps/managarm/generic/file.cpp
@@ -1975,6 +1975,97 @@ int sys_stat(fsfd_target fsfdt, int fd, const char *path, int flags, struct stat
 	return 0;
 }
 
+int
+sys_statx(int dirfd, const char *pathname, int flags, unsigned int mask, struct statx *statxbuf) {
+	SignalGuard sguard;
+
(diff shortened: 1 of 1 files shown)


sysdeps/managarm/generic/file.cpp
```diff
@@ -927,8 +927,8 @@ int sys_msg_recv(int sockfd, struct msghdr *hdr, int flags, ssize_t *length) {
 	managarm::fs::RecvMsgReply<MemoryAllocator> resp(getSysdepsAllocator());
 	resp.ParseFromArray(recv_resp.data(), recv_resp.length());
 
-	if (resp.error() == managarm::fs::Errors::WOULD_BLOCK) {
-		return EAGAIN;
+	if (resp.error() != managarm::fs::Errors::SUCCESS) {
+			return resp.error() | toErrno;
 	} else {
 		__ensure(resp.error() == managarm::fs::Errors::SUCCESS);
```
no92, line 933: This is a separate change, and this ensure is redundant now.

Dennisbonke: Noted, but the change is listed in the commit message it's in. So should be fine right?

PR #1244, opened 2025-02-13
sysdeps/managarm: Implement sys_statfs

Dennisbonke:
Blocked on managarm/managarm#846


sysdeps/managarm/generic/file.cpp
```diff
@@ -2510,15 +2510,16 @@ int sys_fstatfs(int fd, struct statfs *buf) {
 	managarm::posix::FstatfsRequest<MemoryAllocator> req(getSysdepsAllocator());
 	req.set_fd(fd);
 
-	auto [offer, send_req, recv_resp] = exchangeMsgsSync(
```
no92, line 2513: This is a separate change.

Dennisbonke: I get why you'd think that, but this is needed to implement sys_statfs over the fstatfs request. It was changed in Managarm to reflect this, so this is needed here.

PR #1245, opened 2025-02-13
Systemd fixes

Dennisbonke:
This PR holds various systemd fixes.

### DIFF
--- a/sysdeps/managarm/generic/drm.cpp
+++ b/sysdeps/managarm/generic/drm.cpp
@@ -1209,6 +1209,12 @@ int ioctl_drm(int fd, unsigned long request, void *arg, int *result, HelHandle h
 			                    << frg::endlog;
 			return 0;
 		}
+		case DRM_IOCTL_WAIT_VBLANK: {
+			mlibc::infoLogger() << "\e[35mmlibc: DRM_IOCTL_WAIT_VBLANK"
+			                       " is a noop\e[39m"
+			                    << frg::endlog;
+			return 0;
+		}
 		case DRM_IOCTL_PRIME_HANDLE_TO_FD: {
 			auto param = reinterpret_cast<drm_prime_handle *>(arg);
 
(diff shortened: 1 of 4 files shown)


sysdeps/managarm/generic/socket.cpp
```diff
@@ -616,6 +624,21 @@ int sys_setsockopt(int fd, int layer, int number, const void *buffer, socklen_t
 		) << "\e[31mmlibc: setsockopt() call with SOL_IP and IP_RECVERR is unimplemented\e[39m"
 		  << frg::endlog;
 		return 0;
+	} else if (layer == SOL_SOCKET && number == SO_PASSSEC) {
+		mlibc::infoLogger(
+		) << "\e[31mmlibc: setsockopt() call with SOL_SOCKET and SO_PASSSEC is unimplemented\e[39m"
+		  << frg::endlog;
+		return ENOSYS;
+	} else if (layer == SOL_SOCKET && number == SO_TIMESTAMP) {
```
no92, line 632: This option is already handled above.

sysdeps/managarm/generic/drm.cpp
```diff
@@ -1209,6 +1209,12 @@ int ioctl_drm(int fd, unsigned long request, void *arg, int *result, HelHandle h
 			                    << frg::endlog;
 			return 0;
 		}
+		case DRM_IOCTL_WAIT_VBLANK: {
```
no92, line 1212: How spammy is this? Also, implementing this probably isn't hard as we can already await them on file read.

Dennisbonke: Not that spammy at all if I recall correctly.

sysdeps/managarm/generic/socket.cpp
Dennisbonke: I see, removed.

Dennisbonke:
Addressed the issues, I see pre-commit failing but I'm pretty sure we need that header that clang-format wants me to remove.

no92 (approved):
LGTM

PR #1246, opened 2025-02-13
sysdeps/managarm: Pass through 2 additional getsockopt calls

Dennisbonke:
Blocked on managarm/managarm#852


no92 (approved):
LGTM

PR #1250, opened 2025-02-13
abis/linux: Fix wrong definition of the GRND_* constants on the Linux ABI

Dennisbonke:
Systemd checks the values of these, and they were wrong on the Linux ABI from the start. On the mlibc ABI they remain unchanged.


avdgrinten (approved):

no92 (approved):
LGTM

PR #1255, opened 2025-02-19
abis/linux: vm-flags.h: Ensure Linux-specific ABI defines depend on __MLIBC_LINUX_OPTION

mintsuki:


abis/linux/vm-flags.h
```diff
@@ -13,18 +15,23 @@
 #define MAP_FIXED     0x10
 #define MAP_ANON      0x20
 #define MAP_ANONYMOUS 0x20
+#define MAP_NORESERVE 0x4000
```
no92, line 18: This does not seem to be defined by POSIX, but rather to be Linux-specific?

no92 (commented):
One comment, other than that looking good.

abis/linux/vm-flags.h
mintsuki: Fixed.

no92 (approved):
LGTM

PR #1261, opened 2025-02-20
abis/mlibc: Get rid of the mlibc abi

mintsuki:


avdgrinten:
As mentioned on Discord, I'd like to avoid the duplication here.

mintsuki:
I think we should not be concerning ourselves with data deduplication for things we ourselves are making it so they are per-port.

mintsuki:
Also, note, CI keeps failing for no clear reason.

mintsuki:
> Also, note, CI keeps failing for no clear reason.

this works now.

Dennisbonke (approved):
LGTM

PR #1267, opened 2025-02-23
options/posix: Implement shm{at,ctl,dt,get}

mintsuki:


avdgrinten:
Testing the basic functionality of each function (probably a single test is enough for that) would be nice. Otherwise, LGTM.

PR #1269, opened 2025-02-24
options/bsd: Implement getloadavg()

mintsuki:


sysdeps/linux/generic/sysdeps.cpp
```diff
@@ -1795,6 +1795,20 @@ int sys_readlink(const char *path, void *buf, size_t bufsiz, ssize_t *len) {
 	return 0;
 }
 
+#if __MLIBC_BSD_OPTION
+// getloadavg() from musl
+int sys_getloadavg(int *nret, double *a, int n) {
+	struct sysinfo si;
+	if (n <= 0) return n ? EINVAL : 0;
+	sys_sysinfo(&si);
```
avdgrinten, line 1803: What if this fails?

options/bsd/include/mlibc/bsd-sysdeps.hpp
```diff
@@ -5,6 +5,8 @@ namespace [[gnu::visibility("hidden")]] mlibc {
 
 [[gnu::weak]] int sys_brk(void **out);
 
+[[gnu::weak]] int sys_getloadavg(int *nsample_out, double *samples, int nsample);
```
avdgrinten, line 8: I'd prefer to force the sysdep to always return 3 samples on success. `getloadavg()` (not the sysdep) can handle the case that the user passes n < 3.

mintsuki:
I believe I have addressed the concerns.

avdgrinten (approved):

PR #1271, opened 2025-02-25
options/posix: Fix sys_shmctl() to actually return its return value

mintsuki:


options/posix/include/mlibc/posix-sysdeps.hpp
```diff
@@ -243,7 +243,7 @@ int sys_vm_unmap(void *pointer, size_t size);
 [[gnu::weak]] int sys_splice(int in_fd, off_t *in_off, int out_fd, off_t *out_off, size_t size, unsigned int flags, ssize_t *out);
 
 [[gnu::weak]] int sys_shmat(void **seg_start, int shmid, const void *shmaddr, int shmflg);
-[[gnu::weak]] int sys_shmctl(int shmid, int cmd, struct shmid_ds *buf);
+[[gnu::weak]] int sys_shmctl(int *idx, int shmid, int cmd, struct shmid_ds *buf);
```
avdgrinten, line 246: Don't we use the last argument as out argument by convention?

mintsuki: No, and if we do elsewhere, we shouldn't imho

mintsuki: I will fix it in another commit for all of shm* then (this was not caught in review previously).

avdgrinten: :+1: Makes sense, thank you.

PR #1272, opened 2025-02-25
options/posix: include <sys/stat.h> from <ftw.h>

lzcunt:
POSIX says the inclusion of <ftw.h> may also make visible symbols from <sys/stat.h>. Out in the wild, libcap-ng seems to depend on this behaviour.


mintsuki (approved):

mintsuki:
LGTM

PR #1275, opened 2025-02-26
options/ansi: add flag 'x' in modestrings

lzcunt:
This flag was added in C11 and POSIX.1-2024. This patch is needed for tzdata's zic timezone compiler.


mintsuki:
LGTM

mintsuki (approved):

PR #1278, opened 2025-02-28
sysdeps/ironclad: Define missing port-specific _SC_* macros and add ABI lookup table

mintsuki:


avdgrinten (approved):
I'm fine with this solution.

PR #1280, opened 2025-02-28
options/ansi: implement TZ environment variable support

lzcunt:


no92 (approved):
Fixed some nits, the logic itself LGTM.

PR #1286, opened 2025-03-08
Add argument permutation to getopt_long(_only)

no92:
Supersedes #1248.

### DIFF
--- a/options/glibc/generic/getopt.cpp
+++ b/options/glibc/generic/getopt.cpp
@@ -23,7 +23,7 @@ enum GetoptMode {
 	LongOnly,
 };
 
-int getopt_common(int argc, char * const argv[], const char *optstring, const struct option *longopts, int *longindex, enum GetoptMode mode) {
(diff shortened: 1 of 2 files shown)


options/glibc/generic/getopt.cpp
```diff
@@ -237,9 +222,66 @@ int getopt_common(int argc, char * const argv[], const char *optstring, const st
 			}
 		}
 	}
+
+	optarg = nullptr;
 	return -1;
 }
 
+void permute(char **argv, int dest, int src) {
```
mintsuki, line 230: Either assert that `src > dest` and add a comment, or simply handle the case where `src <= dest`.

mintsuki (changes requested):

options/glibc/generic/getopt.cpp
Dennisbonke: The assert was added, after which an OK from @avdgrinten was received.

PR #1289, opened 2025-03-09
options/rtld: Avoid potentially mapping pages as W+X

mintsuki:
This also fixes a bug where the 'Take care of removing superfluous permissions' code wouldn't do so for the entirety of mappings made when MLIBC_MAP_DSO_SEGMENTS is set, but only for the unbacked portion

### DIFF
--- a/options/rtld/generic/linker.cpp
+++ b/options/rtld/generic/linker.cpp
@@ -558,12 +558,12 @@ frg::expected<LinkerError, void> ObjectRepository::_fetchFromFile(SharedObject *
 			#if MLIBC_MAP_DSO_SEGMENTS
 				void *map_pointer;
 				if(mlibc::sys_vm_map(reinterpret_cast<void *>(map_address),
-						backed_map_size, prot | PROT_WRITE,
+						backed_map_size, PROT_READ | PROT_WRITE,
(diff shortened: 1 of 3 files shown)


options/rtld/generic/linker.cpp
```diff
@@ -558,12 +558,12 @@ frg::expected<LinkerError, void> ObjectRepository::_fetchFromFile(SharedObject *
...
 				void *map_pointer;
 				if(mlibc::sys_vm_map(reinterpret_cast<void *>(map_address),
-						backed_map_size, prot | PROT_WRITE,
+						backed_map_size, prot,
 						MAP_PRIVATE | MAP_FIXED, fd, phdr->p_offset - misalign, &map_pointer))
 					__ensure(!"sys_vm_map failed");
 				if(total_map_size > backed_map_size)
 					if(mlibc::sys_vm_map(reinterpret_cast<void *>(map_address + backed_map_size),
-							total_map_size - backed_map_size, prot | PROT_WRITE,
+							total_map_size - backed_map_size, prot,
```
avdgrinten, line 566: AFAICT this was done to ensure that the memset() below succeeds.

mintsuki: Addressed.

avdgrinten (approved):

PR #1296, opened 2025-03-10
sysdeps/managarm: implement sys_vm_protect for rtld

no92:
Depends on managarm/managarm#880.


mintsuki:
rtld*

Also I am pretty sure that this breaks the non-MAP_DSO_SEGMENTS case.

mintsuki (changes requested):
As stated above.

mintsuki (approved):

PR #1299, opened 2025-03-13
options/posix: Remove incorect `__USE_MISC` guard

marv7000:
The `__USE_MISC` define is a GLIBC thing for BSD 4.x extensions and holds no weight in mlibc since it's already gated behind `__MLIBC_BSD_OPTION`.


Dennisbonke (approved):
LGTM

PR #1303, opened 2025-03-16
options/ansi: implement %z specifier for strftime

### LINKED ISSUE #1300: Unimplemented strftime specifier `%z`
While booting into kmscon with `logind` enabled, `/bin/login` seems to pick up on something and tries to call `strftime (ptime, sizeof (ptime), "%a %b %e %H:%M:%S %z %Y", tm);`, which fails as we don't implement `%z`.
Of note, this happens after trying to log in, so after entering an username and password.
The man page says the following
`%z     The +hhmm or -hhmm numeric timezone (that is, the hour and minute offset from UTC). (SU)`
We should implement this at our earliest convenience as this blocks `logind`.

no92:
Fixes #1300.

### DIFF
--- a/options/ansi/generic/time.cpp
+++ b/options/ansi/generic/time.cpp
@@ -167,6 +167,16 @@ size_t strftime(char *__restrict dest, size_t max_size,
 			c++;
 			break;
 		}
+		case 'z': {
+			auto min = tm->tm_gmtoff / 60;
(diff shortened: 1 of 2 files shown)


Dennisbonke (approved):
LGTM

mintsuki (approved):

PR #1312, opened 2025-03-20
Additions for compiling plymouth

no92:


options/glibc/generic/stdlib.cpp
```diff
@@ -0,0 +1,17 @@
+#include <stdlib.h>
+#include <string.h>
+
+int rpmatch(const char *resp) {
```
Geertiebear, line 4: Where test

options/ansi/include/stdlib.h
```diff
@@ -115,6 +115,7 @@ size_t wcstombs(char *__restrict __mb_string, const wchar_t *__restrict __wc_str
 
 #if __MLIBC_GLIBC_OPTION
 typedef int (*comparison_fn_t) (const void *__a, const void *__b);
+int rpmatch(const char *resp);
```
mintsuki, line 118: Please do not continue this broken tradition of inlining glibc stuff in the ANSI stdlib.h!

Make a `stdlib.h` for the glibc option and put it there.

options/posix/include/sys/fcntl.h
```diff
@@ -0,0 +1 @@
+#include <fcntl.h>
```
mintsuki, line 1: Please add the include guards, even if they aren't especially needed.

options/glibc/generic/stdlib.cpp
```diff
@@ -0,0 +1,17 @@
...
+	if(!resp || resp[0] == '\0')
+		return -1;
+
+	if(!strcmp("yes", resp))
+		return 1;
+	if(!strcmp("no", resp))
```
mintsuki, line 10: Maybe these `strcmp`s should be `strcasecmp`s?

mintsuki (changes requested):
See comments

options/posix/include/sys/fcntl.h
no92: IMO include guards on passthrough headers are needless complexity and also not something we do so far, see for instance `sys/syslog.h`.

no92:
Fixed comments + provided a test for rpmatch.

mintsuki (approved):

options/glibc/include/bits/glibc/glibc_stdlib.h
```diff
@@ -0,0 +1,19 @@
+#ifndef MLIBC_GLIBC_STDLIB_H
+#define MLIBC_GLIBC_STDLIB_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#ifndef __MLIBC_ABI_ONLY
+
+typedef int (*comparison_fn_t) (const void *__a, const void *__b);
```
mintsuki, line 10: Actually i just noticed this - this is not a function, so it shouldn't be wrapped by the `#ifndef __MLIBC_ABI_ONLY`

options/glibc/generic/stdlib.cpp
```diff
@@ -0,0 +1,17 @@
...
+	if(!resp || resp[0] == '\0')
+		return -1;
+
+	if(!strcasecmp("yes", resp))
+		return 1;
+	if(!strcasecmp("no", resp))
+		return 0;
+	if(resp[0] == 'y' || resp[0] == 'Y')
+		return 1;
+	if(resp[0] == 'n' || resp[0] == 'N')
```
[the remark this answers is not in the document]
no92: It should, and the man page specifically mentions this.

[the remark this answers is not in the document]
mintsuki: Where does it say that? Because i can't find it.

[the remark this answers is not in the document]
mintsuki: afaict it just says you have to match `^[Yy]` and `^[Nn]` regardless of language, not `^[Yy].*` or `^[Nn].*`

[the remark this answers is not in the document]
mintsuki: okay, that being the case I think this is stupid behaviour that we should not be emulating

no92:
Updated to apply fixes.

PR #1096, opened 2024-07-26
options/posix: Implement swab

Dennisbonke:

PR #1314, opened 2025-03-23
options/posix: Implement swab

mintsuki:
- **options/posix: Implement swab**
- **tests/posix: Add tests for swab**


mintsuki:
Supersedes #1096

PR #1322, opened 2025-03-27
Add loongarch64 support

marv7000:
This PR adds support for LoongArch64 and updates Linux syscall numbers to 6.14


no92:
I ran the testsuite locally for both shared and static mlibc builds, all tests passed.

PR #1326, opened 2025-03-29
options/ansi: Remove unused math-stubs.cpp file

mintsuki:


avdgrinten (approved):

avdgrinten:
We should probably also consider whether we can update the math code from upstream musl. Maybe we can do this in an automated way?

Dennisbonke (approved):
LGTM

mintsuki:
> We should probably also consider whether we can update the math code from upstream musl. Maybe we can do this in an automated way?

I was actually gonna do a PR for that and the other vendoered files in a bit :)

PR #1331, opened 2025-03-29
tests/posix: make SHM tests more resilient to preexisting system shmids

streaksu:
if the system has 23123123 as a used key for a preexisting shm segment, the tests will spuriously fail. This implements a better approach.


no92 (approved):
LGTM

PR #1332, opened 2025-03-30
misc: Remove libm and complex code

mintsuki:


Dennisbonke:
Can we leave this as an option? Same as we do with the crypt, iconv and intl stuff?

mintsuki:
We removed crypt/iconv/intl a while back and it has been fine.

Dennisbonke:
You are right we did, must've slipped my mind. I'd still request keeping this as an option as I have not yet seen an easy to use drop in replacement (and before you comment openlibm, see [this](https://github.com/JuliaMath/openlibm/issues/216))

PR #1333, opened 2025-03-30
sysdeps: Add Menix

marv7000:
This PR adds support for the Menix kernel, including x86_64, aarch64, riscv64 and loongarch64 support.

Some sysdeps are not implemented yet, until then the syscall numbers are all set to 0.


mintsuki:
LGTM

PR #1338, opened 2025-04-01
sysdeps: add nyaux sysdeps

rayanmargham:
Adds Nyaux Sysdeps


mintsuki:
The `abi-bits` headers should really be symlinks to files inside `abi/` (usually Linux for new OSes unless you have a reason to override).

rayanmargham:
On It

PR #1348, opened 2025-04-14
options/glibc: implement __fsetlocking()

rayanmargham:
I've implemented the __fsetlocking() glibc extension. tell me if changes are needed :)

### DIFF
--- a/options/ansi/include/mlibc/file-io.hpp
+++ b/options/ansi/include/mlibc/file-io.hpp
@@ -21,7 +21,26 @@ enum class buffer_mode {
 	line_buffer,
 	full_buffer
 };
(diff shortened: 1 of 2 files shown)


options/ansi/include/mlibc/file-io.hpp
```diff
@@ -21,7 +23,25 @@ enum class buffer_mode {
 	line_buffer,
 	full_buffer
 };
-
+struct fsetlock : RecursiveFutexLock {
```
avdgrinten, line 26: Rename to `StdioLock` or something like that. Also, it would be good to use composition and not inheritance.

PR #1375, opened 2025-06-22
(almost) complete ppc64 support

localcc:
This PR adds almost complete ppc64 support to mlibc. There are mainly two issues with this port currently:

* a couple of TODOs left unimplemented as i never got them to trigger in a testing environment so those will be fixed later. 
* long double support is not completely finished, some functions work properly and some are left broken due to the implementation being too complex to do right now for a feature I wouldn't use

I am also currently not sure if ppc64 support compiles on anything but my custom fork of llvm, so there might be dragons there. I will test it with gcc at some point and report the results. Considering the compilation difficulties I am first creating this as a draft PR so that code can be gradually reviewed.

### DIFF
--- a/options/rtld/powerpc64/entry.S
+++ b/options/rtld/powerpc64/entry.S
@@ -41,9 +41,15 @@ pc:
     bl interpreterMain
     nop
 
+    stdu 1, -112(1)
+    std 2, 40(1)
+
     ld 2, 8(3) // loading toc and jumping to Scrt1.S entry
     ld 3, 0(3)
     mtctr 3
     bctr
+
+    addi 1, 1, 112
+    trap
 
 .section .note.GNU-stack,"",%progbits
--- a/options/rtld/powerpc64/runtime.S
+++ b/options/rtld/powerpc64/runtime.S
@@ -1,9 +1,29 @@
 .abiversion 1
 .global pltRelocateStub
 pltRelocateStub:
-	# we need to save / restore all registers than can hold function arguments
-	# we do not need to save callee-saved registers as they will not be trashed by lazyRelocate
-    # TODO: save floating point argument registers
+	// we need to save / restore all registers than can hold function arguments
+	// we do not need to save callee-saved registers as they will not be trashed by lazyRelocate
+    // TODO: save floating point argument registers
     // todo(localcc): implement
-    trap
+
+    stdu 1, -112(1)
+    std 2, 40(1)
+
+    std 3, 48(1)
+    std 4, 56(1)
+
+    ld 3, 112(1)
+    ld 4, 120(1)
+
+    bl lazyRelocate
+    nop
+    mtctr 3
+
+    ld 3, 48(1)
+    ld 4, 56(1)
+
+    addi 1, 1, 112
+
+    bctr
+
 .section .note.GNU-stack,"",%progbits
--- a/sysdeps/linux/powerpc64/crt-src/crt1.S
+++ b/sysdeps/linux/powerpc64/crt-src/crt1.S
@@ -8,9 +8,13 @@ _start:
 
 .text
 _realstart:
-    // todo(localcc): implement
-    trap
+    mr 3, 1
+
+    addis 4, 2, main@toc@ha
+    addi 4, 4, main@toc@l
+
+    bl __mlibc_entry
     nop
-    trap
+
 .section .note.GNU-stack,"",%progbits
 
--- a/sysdeps/linux/powerpc64/crt-src/crti.S
+++ b/sysdeps/linux/powerpc64/crt-src/crti.S
@@ -1,13 +1,2 @@
-.section .init
-.global _init
-_init:
-    // todo(localcc): implement
-    trap
-
-.section .fini
-.global _fini
-_fini:
-    // todo(localcc): implement
-    trap
 .section .note.GNU-stack,"",%progbits
 
--- a/sysdeps/linux/powerpc64/crt-src/crtn.S
+++ b/sysdeps/linux/powerpc64/crt-src/crtn.S
@@ -1,9 +1,2 @@
-.section .init
-    // todo(localcc): implement
-    trap
-
-.section .fini
-    // todo(localcc): implement
-    trap
 .section .note.GNU-stack,"",%progbits
 


marv7000:
Just rebase onto master

options/ansi/include/bits/ansi/fenv.h
```diff
@@ -87,6 +87,21 @@
...
+#define FE_INVALID 0x20000000
+#define FE_OVERFLOW 0x10000000
+#define FE_UNDERFLOW 0x8000000
+
+#define FE_ALL_EXCEPT (FE_DIVBYZERO | FE_INEXACT | FE_INVALID | FE_OVERFLOW | FE_UNDERFLOW)
+
+#define FE_TONEAREST 0
+#define FE_DOWNWARD 0b11
+#define FE_TOWARDZERO 0b01
+#define FE_UPWARD 0b10
```
no92, line 103: Binary literals are compiler extensions and should therefore not be used in public headers.

options/rtld/generic/linker.cpp
```diff
@@ -858,6 +861,10 @@ void ObjectRepository::_parseDynamic(SharedObject *object) {
 			case DT_RELRENT:
 #ifdef __riscv
 			case DT_TEXTREL: // Work around https://sourceware.org/bugzilla/show_bug.cgi?id=24673.
+#endif
+#if defined(__powerpc64__)
+			case DT_LOPROC:
+			case 0x70000003: // unknown
```
no92, line 867: From binutils source, this seems to be `DT_PPC64_OPT` which specifies what optimizations can be used with the ELFv2 ABI.

options/rtld/generic/main.cpp
```diff
@@ -180,7 +189,159 @@ extern "C" void relocateSelf() {
 		}
 	}
 }
-#else
+#elif defined(__powerpc64__)
+// ppc64 elfv1 can't rely on the OPD before relocations, therefore calling the getLdsoBase function
```
no92, line 193: I wonder if we could just force-inline the function on ppc64?

options/rtld/generic/main.cpp
```diff
@@ -405,6 +568,16 @@ extern "C" void *interpreterMain(uintptr_t *entry_stack) {
 				    << " in program interpreter" << frg::endlog;
 				break;
 			}
+#if defined(__powerpc64__)
+			case DT_LOPROC:
+			case 0x70000003: // unknown
```
no92, line 573: See above.

options/rtld/generic/relocation.hpp
```diff
@@ -0,0 +1 @@
+#pragma once
```
no92, line 1: Leftover?

scripts/get-linux-headers.sh
```diff
@@ -4,7 +4,7 @@ set -ex
 
 LINUX_MAJOR=v6.x
 LINUX_VERSION=6.13.8
-: "${ARCH:=$(uname -m)}"
```
no92, line 7: Leftover?

sysdeps/linux/generic/sysdeps.cpp
```diff
@@ -112,6 +133,16 @@ int sys_tcb_set(void *pointer) {
...
 	asm volatile("move $tp, %0" ::"r"(thread_data));
+#elif defined(__powerpc64__)
+	// todo(localcc): fix this?
+	//	tcbhead_t* tcbhead = reinterpret_cast<tcbhead_t*>(reinterpret_cast<uintptr_t>(pointer) -
+	// sizeof(tcbhead_t)); 	const auto& aux_vec = get_aux_vec();
+	//    tcbhead->hwcap = (static_cast<uint64_t>(aux_vec.at_hwcap) << 32) | aux_vec.at_hwcap2;
+	//    tcbhead->hwcap_extn = (static_cast<uint64_t>(aux_vec.at_hwcap3) << 32) |
+	//    aux_vec.at_hwcap4; tcbhead->at_platform = aux_vec.at_platform;
+
+	uintptr_t thread_data = reinterpret_cast<uintptr_t>(pointer) + 0x7000 + sizeof(Tcb);
```
no92, line 144: It would be preferable to use the constant for the TCB offset.

sysdeps/linux/generic/sysdeps.cpp
```diff
@@ -663,6 +694,11 @@ int sys_clone(void *tcb, pid_t *pid_out, void *stack) {
 	// TODO: We should change the sysdep so that we don't need to do this.
 	auto tp = reinterpret_cast<char *>(tcb) + sizeof(Tcb) - 0x10;
 	tcb = reinterpret_cast<void *>(tp);
+#elif defined(__powerpc64__)
+	// TP should point to the address 0x7000 bytes after the TCB.
+	// TODO: We should change the sysdep so that we don't need to do this.
+	auto tp = reinterpret_cast<char *>(tcb) + sizeof(Tcb) + 0x7000;
```
no92, line 700: Constant.

sysdeps/linux/generic/sysdeps.cpp
```diff
@@ -112,6 +133,16 @@ int sys_tcb_set(void *pointer) {
 #elif defined(__loongarch64)
 	uintptr_t thread_data = reinterpret_cast<uintptr_t>(pointer) + sizeof(Tcb);
 	asm volatile("move $tp, %0" ::"r"(thread_data));
+#elif defined(__powerpc64__)
+	// todo(localcc): fix this?
+	//	tcbhead_t* tcbhead = reinterpret_cast<tcbhead_t*>(reinterpret_cast<uintptr_t>(pointer) -
+	// sizeof(tcbhead_t)); 	const auto& aux_vec = get_aux_vec();
+	//    tcbhead->hwcap = (static_cast<uint64_t>(aux_vec.at_hwcap) << 32) | aux_vec.at_hwcap2;
+	//    tcbhead->hwcap_extn = (static_cast<uint64_t>(aux_vec.at_hwcap3) << 32) |
+	//    aux_vec.at_hwcap4; tcbhead->at_platform = aux_vec.at_platform;
```
no92, line 142: Why is hwcap setup done here instead of somewhere further up the callchain?

sysdeps/linux/generic/sysdeps.cpp
```diff
@@ -663,6 +694,11 @@ int sys_clone(void *tcb, pid_t *pid_out, void *stack) {
 	// TODO: We should change the sysdep so that we don't need to do this.
 	auto tp = reinterpret_cast<char *>(tcb) + sizeof(Tcb) - 0x10;
 	tcb = reinterpret_cast<void *>(tp);
+#elif defined(__powerpc64__)
+	// TP should point to the address 0x7000 bytes after the TCB.
```
no92, line 698: `after the end of the TCB`

abis/linux/shm.h
```diff
@@ -77,6 +77,18 @@ struct shmid_ds {
...
+	struct ipc_perm shm_perm;
+	time_t shm_atime;
+	time_t shm_dtime;
+	time_t shm_ctime;
+	size_t shm_segsz;
+	pid_t shm_cpid;
+	pid_t shm_lpid;
+	unsigned long shm_nattch;
+	unsigned long __unused[2];
+};
```
no92, line 91: I'm not sure this is correct - this seems to match the definition of `shmid64_ds` for ppc64, which is a different struct. I think the correct definition is the one above.

options/rtld/generic/main.cpp
localcc: doesn't work unfortunately

sysdeps/linux/generic/sysdeps.cpp
localcc: no clue tbh where it should be done

PR #1378, opened 2025-07-12
add a bit more localization functions

localcc:
These allow for llvm libcxx to be built with localization enabled now


Dennisbonke:
CI is failing, seems like you're missing an include in a header?

no92:
Implemented by the locale PRs and fully obsoleted by #1776.

PR #1382, opened 2025-07-16
tests/posix: Add test for execve

marv7000:


tests/posix/exec.c
```diff
@@ -0,0 +1,23 @@
+#include <assert.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+
+int main(int argc, const char **argv) {
+	if (argc == 1) {
+		int pid = fork();
```
no92: `fork(3)` returns a `pid_t`. We should also assert that it returns a value >= 0.

tests/posix/exec.c
```diff
@@ -0,0 +1,23 @@
...
+	}
+
+	size_t len = strlen(argv[0]);
+	char *program = calloc(len + 1, sizeof(char));
+	memcpy(program, argv[0], len);
+
+	char foo[] = "foo";
+	char *args[] = {program, foo, NULL};
+
+	return execve(program, args, NULL);
```
no92: `execvpe` would be preferable here?

tests/posix/exec.c
```diff
@@ -0,0 +1,23 @@
+#include <assert.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+
+int main(int argc, const char **argv) {
```
no92: `argv` should be not const.

tests/posix/exec.c
no92: Maybe it would also be preferable to print out the error?


```suggestion
	execvpe(program, args, NULL);
	perror("execvpe");
	return 1;
```

PR #1383, opened 2025-07-16
options/posix: added new sysdep sys_thread_sigmask and used in pthread_sigmask

### LINKED ISSUE #1381: `pthread_sigmask` assumes non-portable unspecified behavior
Hi,
while porting mlibc to an OS, I noticed the following:

https://github.com/managarm/mlibc/blob/1590bbbc70db342adef22c784b6028d108acaae4/options/posix/generic/posix_signal.cpp#L18-L28
Here, mlibc currently appears to assume that `sys_sigprocmask` will affect only the signal mask of the calling thread. According to https://pubs.opengroup.org/onlinepubs/7908799/xsh/sigprocmask.html, however, this behavior is guaranteed only on single-threaded applications:
>  In a single-threaded process, the sigprocmask() function allows the calling process to examine or change (or both) the signal mask of the calling thread. 
> [...]
>  The use of the sigprocmask() function is unspecified in a multi-threaded process. 
> The [pthread_sigmask()](https://pubs.opengroup.org/onlinepubs/7908799/xsh/pthread_sigmask.html) function is used to examine or change (or both) the calling thread's signal mask, regardless of the number of threads in the process. The effect is the same as described for sigprocmask(), without the restriction that the call be made in a single-threaded process.

Thus, as I interpret the spec, conforming system implementations are allowed to handle `sigprocmask` differently in multi-threaded processes, and relying on `sigprocmask` for `pthread_sigmask` results in undefined behavior and may cause compatibility issues with systems that do not adopt a shared namespace for threads and processes (specifically, systems that implement separate paths for setting the current thread's signal mask and for the process's signal mask).
 
If confirmed, a possible fix would be to check for the existence of a new sysdep (something like `sys_thread_sigmask`) and call `sys_sigprocmask` only if this function is missing so not to break preexisting ports that assume the current behavior. A preprocessor macro (like `MLIBC_PTHREAD_SIGMASK_ENABLE_FALLBACK`) may also be used to toggle fallback on `sys_sigprocmask`.

Best regards

Alessandro-Salerno:
This should fix #1381. All automated tests seem to pass.

### DIFF
--- a/options/posix/generic/posix_signal.cpp
+++ b/options/posix/generic/posix_signal.cpp
@@ -16,6 +16,13 @@ int sigsuspend(const sigset_t *sigmask) {
 }
 
 int pthread_sigmask(int how, const sigset_t *__restrict set, sigset_t *__restrict retrieve) {
+ 	if(mlibc::sys_thread_sigmask) {
+	 	if(int e = mlibc::sys_thread_sigmask(how, set, retrieve); e) {
+	 	 	return e;
+	 	}
+ 	 	return 0;
+ 	}
+
 	if(!mlibc::sys_sigprocmask) {
 		MLIBC_MISSING_SYSDEP();
 		return ENOSYS;
--- a/options/posix/include/mlibc/posix-sysdeps.hpp
+++ b/options/posix/include/mlibc/posix-sysdeps.hpp
@@ -168,6 +168,8 @@ int sys_vm_unmap(void *pointer, size_t size);
 [[gnu::weak]] int sys_shutdown(int sockfd, int how);
 [[gnu::weak]] int sys_sigprocmask(int how, const sigset_t *__restrict set,
 		sigset_t *__restrict retrieve);
+[[gnu::weak]] int sys_thread_sigmask(int how, const sigset_t *__restrict set,
+		sigset_t *__restrict retrieve);
 [[gnu::weak]] int sys_sigaction(int, const struct sigaction *__restrict,
 		struct sigaction *__restrict);
 // NOTE: POSIX says that behavior of timeout = nullptr is unspecified. We treat this case


options/posix/generic/posix_signal.cpp
```diff
@@ -16,6 +16,14 @@ int sigsuspend(const sigset_t *sigmask) {
...
 
 int pthread_sigmask(int how, const sigset_t *__restrict set, sigset_t *__restrict retrieve) {
+    if(mlibc::sys_sigthreadmask) {
+        if(int e = mlibc::sys_sigthreadmask(how, set, retrieve); e) {
+            return e;
+        }
+        return 0;
+    }
+
+#ifndef MLIBC_PTHREAD_SIGMASK_NO_FALLBACK
```
Geertiebear: This should also be in the mlibc build config then (can be found in the root).

Alessandro-Salerno: I have updated the build system with a new -Dno_pthread_sigmask_fallback boolean option. Let me know if any further changes are needed.

avdgrinten: Does a build config option really make sense? Each sysdep / OS should know whether it wants this enabled or not.

avdgrinten: I think we should rather remove `MLIBC_PTHREAD_SIGMASK_NO_FALLBACK` entirely and rely on the detection to work correctly.

options/posix/include/mlibc/posix-sysdeps.hpp
```diff
@@ -168,6 +168,8 @@ int sys_vm_unmap(void *pointer, size_t size);
 [[gnu::weak]] int sys_shutdown(int sockfd, int how);
 [[gnu::weak]] int sys_sigprocmask(int how, const sigset_t *__restrict set,
 		sigset_t *__restrict retrieve);
+[[gnu::weak]] int sys_sigthreadmask(int how, const sigset_t *__restrict set,
+		sigset_t *__restrict retrieve);
```
avdgrinten: For consistency with the other `sys_thread_` functions, this should probably be named `sys_thread_sigmask`.

options/posix/generic/posix_signal.cpp
Alessandro-Salerno: What do you suggest? 
I don't know mlibc very well, though I'm trying not to break anything (especially preexisting ports). I thought the flow could be:
- Try to call `sys_sigthreadmask`
- If this fails **and** `MLIBC_PTHREAD_SIGMASK_NO_FALLBACK` is disabled, then try with `sys_sigprocmask`
In theory, this should make is so:
- New or updated ports can plug into the API easily
- Preexisting ports don't have do do anything as mlibc will continuo to behave the same for them
- WIP ports that don't have `sys_sigthreadmask` can use `MLIBC_PTHREAD_SIGMASK_NO_FALLBACK` to avoid calling `sys_sigprocmask` (which, in multi-threaded processes is unspecified by POSIX) and just return `ENOSYS`

options/posix/include/mlibc/posix-sysdeps.hpp
Alessandro-Salerno: Renamed `sys_sigthreadmask` to `sys_thread_sigmask`

options/posix/generic/posix_signal.cpp
avdgrinten: The last case is unnecessary -- if ports don't want the fallback behavior they can just define `sys_thread_sigmask` (to simply return an error).

Alessandro-Salerno: Makes sense, removed it.

avdgrinten (approved):

Alessandro-Salerno:
Thanks! It should all be squashed into one commit now.

no92 (approved):
LGTM

PR #1384, opened 2025-07-16
utmp(x) implementation

no92:
Drafting because I'm awaiting replies on how to deal with ironclad and vinix sysdeps, as they roll their own implementations of utmpx functions.

This implements utmp and utmpx completely and adds tests. utmpx was moved to the posix option, as it is defined by POSIX bar two extensions. Because `struct utmpx` is different between operating systems, it is left as an ABI bit to define for sysdeps, same for the type defines.

### DIFF
--- a/tests/linux/utmp.c
+++ b/tests/linux/utmp.c
@@ -1,0 +1,132 @@
+#include <assert.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <utmp.h>
+
+int main() {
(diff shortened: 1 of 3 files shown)


Dennisbonke (approved):
LGTM

avdgrinten:
Can we test that we can properly read files creates by glibc on Linux? For example, by reading the system's utmp / wtmp (if they exist) and doing some basic sanity tests on them?

no92:
Testing with a host systems' utmp file is a bit hard because it is writing out struct utmp(x), which on linux depends on `__WORDSIZE_TIME64_COMPAT32` to determine its size.

PR #1385, opened 2025-07-20
options/glibc: implement `backtrace`

no92:


ElectrodeYT (approved):
LGTM

meson.build
```diff
@@ -452,6 +452,7 @@ rtlib_deps = []
 
 if not headers_only
 	if libgcc_dependency
+		libgcc_s = meson.get_compiler('c').find_library('gcc_s', required: false)
```
avdgrinten: `libgcc_s.so` has `libc.so` as a dependency (and `libc.so` has `ld.so` as dependency), so we cannot link `libgcc_s.so` with either `ld.so` or `libc.so`.

no92: Changed this to use `libgcc_eh.a`

Dennisbonke (approved):
LGTM

options/glibc/generic/execinfo.cpp
```diff
@@ -1,11 +1,67 @@
...
+
+	if (!libgccHandle) {
+		libgccHandle = dlopen("libgcc_s.so.1", RTLD_LAZY | RTLD_LOCAL);
+		if (!libgccHandle || libgccHandle.value() == nullptr) {
+			mlibc::infoLogger() << "Failed to load libgcc_s.so.1: " << (dlerror() ? dlerror() : "") << frg::endlog;
+			return 0;
+		}
+
+		unwindBacktrace = reinterpret_cast<_Unwind_Reason_Code (*)(_Unwind_Trace_Fn, void *)>(dlsym(libgccHandle.value(), "_Unwind_Backtrace"));
+		unwindGetIP = reinterpret_cast<_Unwind_Ptr (*)(_Unwind_Context *)>(dlsym(libgccHandle.value(), "_Unwind_GetIP"));
```
avdgrinten: Adding typedefs for these would make the code less ugly.

options/glibc/generic/execinfo.cpp
```diff
@@ -1,11 +1,67 @@
...
+		ip -= 2;
+#else
+#warning "Missing support for architecture"
+		ip--;
+#endif
+	}
+
+	UnwindState *state = static_cast<UnwindState *>(arg);
+	state->frames[state->current_frame++] = reinterpret_cast<void *>(ip);
+	return (state->current_frame >= state->count) ? _URC_END_OF_STACK : _URC_NO_REASON;
```
avdgrinten: Doesn't it make more sense to check for overflow before writing? For example, this will fail for `state->count == 0`.

options/rtld/generic/linker.cpp
```diff
@@ -254,6 +254,39 @@ frg::expected<LinkerError, SharedObject *> ObjectRepository::requestObjectWithNa
...
+			}
+			return result.error();
+		}
+
+		_parseDynamic(object);
+		_parseVerdef(object);
+		_addLoadedObject(object);
+
+		return object;
+	}
```
avdgrinten: This logic should not be duplicated. This should be moved into an own function that is called both on the RPATH code path and on the library path code path. Also, error handling can potentially be improve with scope_exit.

PR #1388, opened 2025-07-24
options/posix: fix bug with mkostemp

oberrow:
mkostemp would previously pass 0 to flags and flags to suffixlen when forwarding its parameters to the mkostemps function. This commit should fix that.


Dennisbonke (approved):
LGTM, good catch! Thanks for the fix.

PR #1407, opened 2025-08-09
getaddrinfo improvements

no92:


options/posix/generic/lookup.cpp
```diff
@@ -120,6 +123,9 @@ int lookup_name_dns(struct lookup_result &buf, const char *name,
 		if (response_header->identification != header.identification)
 			return -EAI_FAIL;
 
+		if ((ntohs(response_header->flags) & 0xF) == RETURN_NXDOMAIN)
```
rayanmargham, line 132: You need to handle AI_NUMERICSERV

no92: Ah, turns out this is already handled by `lookup_serv_by_name`.

Dennisbonke (approved):
LGTM

PR #1412, opened 2025-08-12
sysdeps/managarm: Add EBADF handling to sys_fsync

Dennisbonke:
`glib` testsuite expects this for invalid `fd` in `fsync`, and it matches spec. Return EBADF in fsync on closed fd's.


no92 (approved):
LGTM

PR #1426, opened 2025-08-21
sysdeps/ironclad: Enable MLIBC_MMAP_ALLOCATE_DSO

mintsuki:


sysdeps/ironclad/generic/generic.cpp
```diff
@@ -698,6 +698,10 @@ struct futex_item {
 int sys_futex_wait(int *pointer, int expected, const struct timespec *time) {
 	int ret, errno;
 	struct futex_item item = {.addr = (uint64_t)pointer, .expected = (uint32_t)expected, .flags = 0};
+	if (*pointer != expected) {
+		return EAGAIN;
+	}
```
avdgrinten: Does the kernel also check this after taking the futex lock? Since just checking it in userspace is not enough to satisfy the futex requirements.

Mintsuki: I removed this problematic hunk pending future resolution.

PR #1436, opened 2025-08-24
options/posix: fix issues with utmpx's file being created by root and being used by non root users

streaksu:


mintsuki (approved):

Dennisbonke:
Wouldn't it be beneficial to check for EPERM, and only try to reopen as read-only if we failed due to permission error?

PR #1438, opened 2025-08-28
rust-libc bindings generator updates

no92:
This seems to generate libc bindings that compile without warnings as of `0.2.175`.


Dennisbonke (approved):
LGTM

PR #1448, opened 2025-09-08
options/rtld: Implement relocations in text sections (DT/DF_TEXTREL)

oberrow:


options/rtld/generic/linker.cpp
```diff
@@ -775,18 +786,17 @@ void ObjectRepository::_parseDynamic(SharedObject *object) {
 				object->eagerBinding = true;
 
 			auto ignored = DF_BIND_NOW | DF_SYMBOLIC | DF_STATIC_TLS;
-#ifdef __riscv
-			// Work around https://sourceware.org/bugzilla/show_bug.cgi?id=24673.
 			ignored |= DF_TEXTREL;
```
Dennisbonke: We can just add this to the `auto ignored` line above no?

options/rtld/generic/linker.cpp
```diff
@@ -896,6 +906,19 @@ void ObjectRepository::_parseDynamic(SharedObject *object) {
 		object->soName = reinterpret_cast<const char *>(object->baseAddress
 				+ object->stringTableOffset + *soname_offset);
 	}
+	if (object->haveTextRel)
```
Dennisbonke: Codestyle mismatcch, braces go on the same line, no space between `if` and the opening brace (same for `for`). `*` goes to the right side of the statement, and it's `(void *)`, not `(void*)`.

options/rtld/generic/linker.cpp
```diff
@@ -1,3 +1,4 @@
+#include "mlibc/internal-sysdeps.hpp"
```
Dennisbonke: We use `<>` for header includes.

options/rtld/generic/linker.cpp
oberrow: Yes, we can, I don't remember why I did like that.

options/rtld/generic/linker.cpp
oberrow: I will fix that and commit the changes.

avdgrinten:
Why is this needed? On which architecture / target do you need `.text` relocations?

options/rtld/generic/linker.hpp
```diff
@@ -248,6 +249,8 @@ struct SharedObject {
 	size_t phdrEntrySize = 0;
 	size_t phdrCount = 0;
 
+	frg::vector<void*, MemoryAllocator> exec_phdrs;
```
avdgrinten, line 252: We already have `phdrPointer` etc. which can be used to iterate over the PHDRs.

options/rtld/generic/linker.cpp
```diff
@@ -896,6 +902,17 @@ void ObjectRepository::_parseDynamic(SharedObject *object) {
...
+		for(auto &phdr_ptr : object->exec_phdrs) {
+			elf_phdr* phdr = (elf_phdr *)phdr_ptr;
+			void* addr = (void *)(phdr->p_vaddr + (uintptr_t)object->baseAddress);
+			int prot = PROT_WRITE | PROT_READ | PROT_EXEC;
+			if (mlibc::sys_vm_protect)
+				mlibc::sys_vm_protect(addr, phdr->p_memsz, prot);
+			else
+				__ensure(!"sys_vm_protect required when DF_TEXTREL/DT_TEXTREL is present");
+		}
+	}
```
avdgrinten, line 915: This could be done in `linkObjects()` to be symmetrical with the code path that removes `PROT_WRITE` again.

options/rtld/generic/linker.cpp
```diff
@@ -896,6 +902,17 @@ void ObjectRepository::_parseDynamic(SharedObject *object) {
 		object->soName = reinterpret_cast<const char *>(object->baseAddress
 				+ object->stringTableOffset + *soname_offset);
 	}
+	if(object->haveTextRel) {
+		for(auto &phdr_ptr : object->exec_phdrs) {
+			elf_phdr* phdr = (elf_phdr *)phdr_ptr;
+			void* addr = (void *)(phdr->p_vaddr + (uintptr_t)object->baseAddress);
+			int prot = PROT_WRITE | PROT_READ | PROT_EXEC;
```
avdgrinten, line 909: We should never map as RWX. Instead, map as RW first and then change that to RX.

PR #1451, opened 2025-09-11
add cryptix sysdeps - adjusted based on marv's guidelines

Vitriol1744:


sysdeps/cryptix/meson.build
```diff
@@ -0,0 +1,11 @@
+sysdep_supported_options = {
+  'posix': true,
+  'linux': true,
```
no92, line 3: AFAICT the linux option seems to be incompatible with this PR - enabling this option requires providing linux kernel headers, while parts of this PR directly re-implement things they would provide.

sysdeps/cryptix/sysdeps/meson.build
```diff
@@ -0,0 +1,12 @@
...
+  'internal.cpp',
+  'memory.cpp',
+  'process.cpp',
+  'signal.cpp',
+  'system.cpp',
+  'time.cpp',
+  'vfs.cpp',
+)
+
+rtld_dso_sources += common_sources
```
no92, line 11: Are you sure these should only go into the dso sources?

sysdeps/cryptix/sysdeps/internal.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+namespace mlibc {
+static void do_log(const char *message, size_t length) { Syscall(SYS_DEBUG_LOG, message, length); }
+void sys_libc_log(const char *message) {
+	static constexpr char MLIBC_SIG[] = "[mlibc]: ";
+
+	size_t len = strlen(message);
+
+	long nwritten = 0;
+	do_log(MLIBC_SIG, sizeof(MLIBC_SIG));
```
no92, line 21: If you have a writev equivalent, it might be useful to use it here.

sysdeps/cryptix/include/cryptix/syscall.h
```diff
@@ -0,0 +1,217 @@
...
+	}
+
+inline int syscall_error(intptr_t ret) {
+	auto v = static_cast<intptr_t>(ret);
+	if (static_cast<uintptr_t>(v) > -4096ul)
+		return -v;
+	return 0;
+}
+
+#pragma region
```
no92, line 21: Where are these pragmas supported? AFAICT this is not listed by GCC or clang anywhere, but it doesn't give warnings because we use `-Wno-unknown-pragmas`.

sysdeps/cryptix/include/abi-bits/do
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
```
no92, line 1: This looks like it shouldn't be checked in - we already have a helper script.

sysdeps/cryptix/entry/entry.cpp
```diff
@@ -0,0 +1,63 @@
...
+	LibraryGuard();
+};
+
+static LibraryGuard guard;
+namespace mlibc {
+
+int sys_anon_allocate(size_t size, void **pointer);
+}
+
+LibraryGuard::LibraryGuard() {
```
no92, line 47: Why is this here instead of just using the `init_libc` constructor?

sysdeps/cryptix/arch/x86_64/crt0.S
```diff
@@ -0,0 +1,7 @@
+.section .text
+.global _start
+_start:
+	mov %rsp, %rdi
+	mov $main, %rsi
+	call __mlibc_entry
+.section .note.GNU-stack,"",%progbitst
```
no92, line 7: You might want to use `GNU_STACK_NOTE` from the `mlibc-asm/helpers.h` header.

sysdeps/cryptix/include/cryptix/syscall.h
Qwinci: They are originally an MSVC extension but GCC >= 13 supports them and Clang since like 3.0 or something like that (+ they are only editor hints anyway to make the block of code between them foldable so they are ignored by the compilers).

PR #1453, opened 2025-09-13
Astral + dnotify + nice

Mathewnd:
Update Astral's sysdep, add some dnotify defines needed for wine and unstub sys_nice.


avdgrinten (approved):

options/posix/include/mlibc/posix-sysdeps.hpp
```diff
@@ -253,6 +253,8 @@ int sys_vm_unmap(void *pointer, size_t size);
 
 [[gnu::weak]] int sys_inet_configured(bool *ipv4, bool *ipv6);
 
+[[gnu::weak]] int sys_nice(int nice, int *ret);
```
avdgrinten: `ret` should be called `new_nice`.

options/posix/generic/unistd.cpp
```diff
@@ -571,9 +571,16 @@ int lockf(int fd, int op, off_t size) {
...
 }
 
-int nice(int) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+int nice(int nice) {
+	int new_nice;
+
+	MLIBC_CHECK_OR_ENOSYS(mlibc::sys_nice, -1);
+	if(int e = mlibc::sys_nice(nice, &new_nice); e) {
```
no92, line 578: It would be very nice to implement this correctly in the linux sysdeps right away (to notice any unforeseen defects), especially as implementing this correctly is not entirely non-trivial due to possible range shifts between this and `[sg]etpriority`.

options/posix/generic/unistd.cpp
```diff
@@ -571,9 +571,16 @@ int lockf(int fd, int op, off_t size) {
 	return -1;
 }
 
-int nice(int) {
-	__ensure(!"Not implemented");
-	__builtin_unreachable();
+int nice(int nice) {
+	int new_nice;
```
no92, line 575: Given that POSIX specifies what range of values is even acceptable and should be clamped to, it would be useful to do this clamping here instead of within the sysdeps.

avdgrinten: IMO it's better to let the sysdeps handle that -- the syscalls need to protect against OOB anyway and doing it here might mask issues in the sysdeps. In general, it's not libc's business to check the validity of sysdep arguments.

avdgrinten: A better approach is to have a test that tests whether illegal values are clamped by the sysdep.

sysdeps/linux/generic/sysdeps.cpp
```diff
@@ -839,6 +839,28 @@ int sys_setpriority(int which, id_t who, int prio) {
...
+	int current;
+	if (int e = sys_getpriority(0, 0, &current); e)
+		return e;
+
+	if (increment == 0) {
+		*new_nice = current;
+		return 0;
+	}
+
+	if (int e = sys_setpriority(0, 0, current + increment); e)
```
no92, line 852: It would be nice to note that the syscall silently clamps the value to the range; otherwise it might seem like clamping would be the sysdep's responsibility.

sysdeps/linux/generic/sysdeps.cpp
```diff
@@ -839,6 +839,28 @@ int sys_setpriority(int which, id_t who, int prio) {
 	return 0;
 }
 
+int sys_nice(int increment, int *new_nice) {
+	int current;
+	if (int e = sys_getpriority(0, 0, &current); e)
```
no92, line 844: It would be nice (pun unintended) to comment what the actual constant is supposed to be. I don't think that breaking ABI for this right now would be beneficial, as I have a few more fixes queued up in my ABI checker PR that I'll probably post early next week.

no92 (approved):
LGTM

PR #1458, opened 2025-09-24
ABI checker fixes part 1

no92:


options/ansi/include/fenv.h
```diff
@@ -2,22 +2,12 @@
...
 
-typedef struct {
-	__mlibc_uint32 __control_word;
-	__mlibc_uint32 __status_word;
-	__mlibc_uint32 __unused[5];
-	__mlibc_uint32 __mxcsr;
-} fenv_t;
-
-typedef __mlibc_uint16 fexcept_t;
-
```
avdgrinten, line 20: While this is a useful fix, it probably has non-trivial implications and requires a world bump.

options/ansi/include/bits/ansi/fenv.h
```diff
@@ -1,6 +1,10 @@
 #ifndef MLIBC_FENV_H
 #define MLIBC_FENV_H
 
+#if !defined(__ASSEMBLER__)
+#include <bits/types.h>
+#endif /* !defined(__ASSEMBLER__) */
```
avdgrinten, line 6: This file won't compile in assembler mode anyway, will it?

avdgrinten (approved):
LGTM

options/ansi/include/bits/ansi/fenv.h
no92: m68k's `fenv.S` includes this to use some constants.

PR #1459, opened 2025-09-24
ABI checker fixes part 2

no92:


abis/linux/msg.h
```diff
@@ -15,7 +15,7 @@ typedef unsigned long msgqnum_t;
 
 struct msqid64_ds {
 	struct ipc64_perm msg_perm;
-#if (UINTPTR_MAX == UINT64_MAX) /* || x32 ABI */
+#if (__INTPTR_WIDTH__ == 64) /* || x32 ABI */
```
avdgrinten, line 18: Why is this done?

abis/linux/shm.h
```diff
@@ -59,10 +59,7 @@ struct shmid_ds {
 	pid_t shm_cpid;
 	pid_t shm_lpid;
 	unsigned long shm_nattch;
-	unsigned long __unused[3];
-	time_t shm_atime;
-	time_t shm_dtime;
-	time_t shm_ctime;
+	unsigned long __unused[2];
 };
```
avdgrinten, line 63: This does not match glibc/musl. In particular, removing `shm_{a,d,c}time` without replacement will most likely break downstream programs.

abis/linux/msg.h
no92: iirc the header of these macros was not included, but that did not throw warnings so the check was silently broken.

abis/linux/shm.h
no92: Fixed in the fixup commit - the layout already matches what glibc exposes, only changing the names was necessary.

abis/linux/signal.h
```diff
@@ -348,15 +348,42 @@ struct sigcontext {
...
+	uint32_t cw;
+	uint32_t sw;
+	uint32_t tag;
+	uint32_t ipoff;
+	uint32_t cssel;
+	uint32_t dataoff;
+	uint32_t datasel;
+	struct _fpreg _st[8];
+	uint16_t status;
+};
```
avdgrinten: This should be re-investigated: this does not match glibc's `struct _libc_fpstate`. Also, we should probably break up the unified `#if defined(__x86_64__) || defined(__i386__)` since the two cases are not quite similar.

PR #1460, opened 2025-09-24
ABI checker fixes part 3

no92:


options/posix/include/sys/resource.h
```diff
@@ -9,9 +9,9 @@
 #include <bits/posix/id_t.h>
 #include <bits/posix/timeval.h>
 
-#define PRIO_PROCESS 1
-#define PRIO_PGRP 2
-#define PRIO_USER 3
+#define PRIO_PROCESS 0
+#define PRIO_PGRP 1
+#define PRIO_USER 2
```
avdgrinten: Is this Linux kernel ABI that should be moved to `abis/`?

options/posix/include/bits/posix/fd_set.h
```diff
@@ -3,7 +3,7 @@
 
 #include <bits/types.h>
 
-typedef struct {
+typedef struct __attribute__((__aligned__(__alignof__(long)))) {
 	__mlibc_uint8 fds_bits[128];
 } fd_set;
```
avdgrinten, line 10: Is this kernel ABI that should go to `abis/`?

options/posix/include/sys/sem.h
```diff
@@ -29,6 +29,7 @@ struct semid_ds {
 	time_t          sem_ctime;
 
 	unsigned long   sem_nsems;
+	unsigned long   __unused[2];
```
avdgrinten: Is `struct semid_ds` kernel ABI and should be moved to `abis/`?

options/glibc/include/sys/procfs.h
```diff
@@ -1,17 +1,24 @@
...
 typedef unsigned long elf_greg_t;
 
-#define ELF_NGREG (sizeof (struct user_regs_struct) / sizeof (elf_greg_t))
+#if defined(__x86_64__) || defined(__i386__) || defined(__loongarch64) || defined(__m68k__)
+#	include <sys/user.h>
+
+#	define ELF_NGREG (sizeof (struct user_regs_struct) / sizeof (elf_greg_t))
+#else
+#	define ELF_NGREG NGREG
+#endif
```
avdgrinten, line 20: Should the architectures in the `#else` be listed explicitly?

options/posix/include/dirent.h
```diff
@@ -24,7 +24,7 @@ extern "C" {
 /* The character array d_name is of unspecified size, but the number of bytes preceding
  * the terminating null byte will not exceed {NAME_MAX}. */
 #define __MLIBC_DIRENT_BODY ino_t d_ino; \
-			off_t d_off; \
+			off64_t d_off; \
```
avdgrinten: This change doesn't make a lot of sense to me since we define off_t == off64_t anyway.

options/posix/include/fcntl.h
```diff
@@ -22,16 +22,16 @@ extern "C" {
 struct flock {
 	short l_type;
 	short l_whence;
-	off_t l_start;
-	off_t l_len;
+	off64_t l_start;
+	off64_t l_len;
```
avdgrinten: Same (doesn't make sense).

options/posix/include/fcntl.h
```diff
@@ -22,16 +22,16 @@ extern "C" {
...
 	pid_t l_pid;
 };
 
 struct flock64 {
 	short l_type;
 	short l_whence;
-	off_t l_start;
-	off_t l_len;
+	off64_t l_start;
+	off64_t l_len;
```
avdgrinten: Same (doesn't make sense).

options/glibc/include/net/route.h
```diff
@@ -20,7 +20,11 @@ struct rtentry {
 	unsigned long int rt_pad3;
 	unsigned char rt_tos;
 	unsigned char rt_class;
+#if __INTPTR_WIDTH__ == 64
 	short int rt_pad4[3];
+#else
+	short int rt_pad4;
+#endif
```
avdgrinten: Is this kernel ABI that should be in `abis/`?

options/posix/include/unistd.h
```diff
@@ -362,11 +362,7 @@ extern int optopt;
 #endif /* !__MLIBC_ABI_ONLY */
 
 /* Non-POSIX functions supported by Linux. */
-#if UINTPTR_MAX == UINT64_MAX
-typedef __mlibc_uint64 useconds_t;
-#else
 typedef __mlibc_uint32 useconds_t;
-#endif
```
avdgrinten, line 369: I think this change is fine but glibc does have `useconds64_t` and we may need to break ABI again in the future if we want to support 64-bit time_t.

Issue #1431, opened 2025-08-22
Assertion failure in rtld while running `glib` tests


Dennisbonke:
Two tests in the `glib` test suite, being `module-test-library` and `module-test-plugin`, trip [this assert](https://github.com/managarm/mlibc/blob/e435d34da5b3e8c56bc7dd44bbb6dcbedc998327/options/rtld/generic/linker.cpp#L162). We should investigate and fix this issue, increasing resilience inside rtld and enabling more tests to succeed.

PR #1461, opened 2025-09-24
Handle errors in shared object loading

no92:
This PR aims to have the dynamic linker tolerate failures in opening a shared object, in which case it should continue going down its list of paths to find an appropriate shared object.

This can reasonably occur in the wild, as for instance RPATHs can point to broken shared objects. Alternatively, I have encountered the case on my machine (Fedora) where the first `libgcc_s.so.1` found is actually for i386, which is rejected, and further down the list of library paths the x86_64 `libgcc_s.so.1` is found.


no92:
I'm not sure and can't test while not at home, but I think this might address the issue in #1431.

options/rtld/generic/linker.cpp
```diff
@@ -285,79 +287,101 @@ frg::expected<LinkerError, SharedObject *> ObjectRepository::requestObjectWithNa
...
+	auto trySharedObjectSetup = [&] (frg::string<MemoryAllocator> path, int fd) -> frg::expected<LinkerError, SharedObject *> {
+		auto object = frg::construct<SharedObject>(getAllocator(),
+			name.data(), std::move(path), false, localScope, rts);
+
+		auto result = _fetchFromFile(object, fd);
+		closeOrDie(fd);
+		if(!result) {
+			if (verbose || stillSlightlyVerbose)
+				mlibc::infoLogger() << "rtld: failed to open " << name << frg::endlog;
+			frg::destruct(getAllocator(), object);
```
avdgrinten: This could return a `unique_ptr<SharedObject>` to avoid the need for manual `frg::destruct`.

options/rtld/generic/linker.cpp
```diff
@@ -285,79 +287,101 @@ frg::expected<LinkerError, SharedObject *> ObjectRepository::requestObjectWithNa
...
 			auto path = rpath.sub_string(start, idx - start);
 			start = idx + 1;
-			auto [fd_, fullPath] = processRpath(path);
-			if (fd_ != -1) {
-				fd = fd_;
-				chosenPath = std::move(fullPath);
-				break;
+			auto rpathResult = processRpath(path);
+			if (rpathResult) {
+				res = trySharedObjectSetup(rpathResult.value().get<0>(), rpathResult.value().get<1>());
```
avdgrinten, line 350: Doesn't it make more sense to propagate the error if the open succeeds and the load fails? Basically, why don't we handle this like in the last case?

options/rtld/generic/linker.cpp
```diff
@@ -285,79 +287,101 @@ frg::expected<LinkerError, SharedObject *> ObjectRepository::requestObjectWithNa
...
+		if (!res) {
 			auto path = rpath.sub_string(start, rpath.size() - start);
-			auto [fd_, fullPath] = processRpath(path);
-			if (fd_ != -1) {
-				fd = fd_;
-				chosenPath = std::move(fullPath);
-			}
+			auto rpathResult = processRpath(path);
+			if (rpathResult)
+				res = trySharedObjectSetup(rpathResult.value().get<0>(), rpathResult.value().get<1>());
```
avdgrinten, line 362: Likewise.

options/rtld/generic/linker.cpp
no92: libsmarter currently does not have a `unique_ptr`.

avdgrinten: `frg` has a unique_ptr though.

options/rtld/generic/linker.cpp
no92: Added an explanatory comment why we want this.

avdgrinten (approved):

PR #1476, opened 2025-10-06
sysdeps/managarm: ppoll and friends

no92:
Depends on managarm/managarm#1067.

### DIFF
--- a/options/ansi/generic/signal.cpp
+++ b/options/ansi/generic/signal.cpp
@@ -34,6 +34,11 @@ int raise(int sig) {
 
 // This is a POSIX extension, but we have it in here for sigsetjmp
 int sigprocmask(int how, const sigset_t *__restrict set, sigset_t *__restrict retrieve) {
+	if (set && how != SIG_BLOCK && how != SIG_SETMASK && how != SIG_UNBLOCK) {
+		errno = EINVAL;
+		return -1;
+	}
+
 	MLIBC_CHECK_OR_ENOSYS(mlibc::sys_sigprocmask, -1);
 	if(int e = mlibc::sys_sigprocmask(how, set, retrieve); e) {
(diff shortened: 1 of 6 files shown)


options/ansi/generic/signal.cpp
```diff
@@ -34,6 +34,11 @@ int raise(int sig) {
 
 // This is a POSIX extension, but we have it in here for sigsetjmp
 int sigprocmask(int how, const sigset_t *__restrict set, sigset_t *__restrict retrieve) {
+	if (set && how != SIG_BLOCK && how != SIG_SETMASK && how != SIG_UNBLOCK) {
+		errno = EINVAL;
+		return -1;
+	}
```
avdgrinten: This should not be done in mlibc but in the sysdep. The sysdep has to do it anyway (assuming that mlibc is not the only way to call into the corresponding syscall).

The best way to handle this in mlibc is adding a test that checks that the sysdep has the right behavior.

options/posix/generic/posix_signal.cpp
```diff
@@ -16,6 +16,9 @@ int sigsuspend(const sigset_t *sigmask) {
 }
 
 int pthread_sigmask(int how, const sigset_t *__restrict set, sigset_t *__restrict retrieve) {
+	if (set && how != SIG_BLOCK && how != SIG_SETMASK && how != SIG_UNBLOCK)
+		return EINVAL;
```
avdgrinten: Same (check should be done in sysdep).

options/ansi/generic/signal.cpp
no92: We are required to reject non-SIG_{BLOCK,SETMASK,UNBLOCK} according to POSIX, though.

PR #1480, opened 2025-10-08
Proper locale support

no92:


options/internal/generic/locale.cpp
```diff
@@ -1,98 +1,627 @@
...
+	}
+
+	return false;
+}
+
+bool findLocaleArchiveRecord(int category, const char *name, mlibc::localeinfo *out) {
+	if (!localeArchive)
+		return false;
+
+	__ensure(category != LC_ALL);
```
Dennisbonke, line 502: Can't we just return false here? I know we're AssertOS but if we can gracefully fail here I think that's better.

Dennisbonke (commented):
I'm no locale expert but this looks nice. One question popped up in my mind while reviewing but I think this is good.

options/internal/generic/locale.cpp
no92: Passing LC_ALL to this archive parsing function would be a severe logic bug, hence the assert.

Dennisbonke: Alright makes sense then.

options/ansi/generic/ctype.cpp
```diff
@@ -118,209 +119,160 @@ int isascii(int nc) {
 // --------------------------------------------------------------------------------------
 
 int iswalpha(wint_t nc) {
+	auto l = mlibc::getActiveLocale();
```
Mathewnd: Is there a reason for this ``mlibc::getActiveLocale()`` to be separate here? It is inconsistent with how it was done in the first half of the file

Mathewnd: Same applies to all other instances below

options/ansi/generic/ctype.cpp
```diff
@@ -4,6 +4,7 @@
```
Mathewnd, line 4: It would be nice to have some sort of code deduplication in this file for maintainability.

options/ansi/generic/stdio.cpp
```diff
@@ -18,6 +18,7 @@
 #include <mlibc/allocator.hpp>
```
Mathewnd, line 18: There is some more missing locale handling in this file, especially scanf stuff:
Line 526 ``/* TODO: numeric seperators locale stuff */``, as well as floating point decimal point stuff

options/internal/generic/ctype.cpp
```diff
@@ -0,0 +1,46 @@
+#include <mlibc/charcode.hpp>
```
Mathewnd, line 1: Again, concerns about code duplication

options/internal/generic/charset.cpp
```diff
@@ -2,6 +2,92 @@
...
+#if __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
+	return 1 << std::to_underlying(bit);
+#else
+	return std::to_underlying(bit) < 8
+		? ((1 << std::to_underlying(bit)) << 8)
+		: ((1 << std::to_underlying(bit)) >> 8);
+#endif
+}
+
+int wctype_table_lookup(frg::span<const uint32_t> table, uint32_t wc) {
```
Mathewnd, line 35: The only thing that differs between this and ``wctrans_table_lookup`` is the calculation of ``index3`` and of the return value, maybe have a function to handle the other calculations?

tests/posix/locale.c
```diff
@@ -0,0 +1,240 @@
...
+	assert(upper == 'A');
+
+	assert(isalpha('z'));
+	assert(!isalpha('z' + 1));
+
+	snprintf(buf, sizeof(buf), "%'d", 1234567);
+	assert(!strcmp("1.234.567", buf));
+
+	snprintf(buf, sizeof(buf), "%'g", 12345.67);
+	// assert(!strcmp("12.345,7", buf));
```
Mathewnd, line 196: Forgot to uncomment?

tests/posix/locale.c
```diff
@@ -0,0 +1,240 @@
+#ifndef _GNU_SOURCE
```
Mathewnd, line 1: Missing scanf tests

tests/ansi/strtof.c
```diff
@@ -22,6 +23,8 @@
 	assert(pEnd == (off == -1 ? s + strlen(s) : s + off)); })
 
 int main () {
+	setlocale(LC_ALL, "C");
```
Mathewnd, line 26: Since locale can affect functions like printf and scanf as well, might be reasonable to add this to their test files?

tests/ansi/locale.c
```diff
@@ -1,18 +1,42 @@
...
+	char lower = tolower('A');
+	assert(lower == 'a');
+	char upper = toupper(lower);
+	assert(upper == 'A');
+
+	assert(isalpha('z'));
+	assert(!isalpha('z' + 1));
+
+	snprintf(buf, sizeof(buf), "%'g", 12345.67);
+	// assert(!strcmp("12.345,7", buf));
```
Mathewnd, line 39: Forgot to uncomment?

options/posix/generic/posix_ctype.cpp
```diff
@@ -2,135 +2,251 @@
 #include <wctype.h>
```
Mathewnd, line 2: Again, code deduplication

options/internal/generic/locale.cpp
```diff
@@ -1,98 +1,803 @@
...
+	if (!normCodeset.empty()) {
+		frg::string<MemoryAllocator> str{*lang, getAllocator()};
+		str += '.';
+		str += normCodeset;
+		ret.push_back(std::move(str));
+	}
+	// lang
+	ret.push_back(frg::string<MemoryAllocator>{*lang, getAllocator()});
+
+	// if (debugLocale) {
```
Mathewnd: Supposed to be commented out?

tests/ansi/locale.c
no92: Nope, it's just not implemented in frigg yet.

options/ansi/generic/stdio.cpp
no92: Intentionally left out to keep the PR size reasonable (lol)

tests/posix/locale.c
no92: It's not implemented in frigg yet.

PR #1489, opened 2025-10-16
Hex float support

no92:
Depends on managarm/frigg#102.

### DIFF
--- a/tests/ansi/sprintf.c
+++ b/tests/ansi/sprintf.c
@@ -4,6 +4,14 @@
 #include <math.h>
 #include <string.h>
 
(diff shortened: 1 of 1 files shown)


options/ansi/generic/stdio.cpp
```diff
@@ -887,6 +901,160 @@ static int do_scanf(H &handler, const char *fmt, __builtin_va_list args) {
 
 				continue;
 			}
+			case 'e':
+			case 'g':
+			case 'E':
+			case 'a':
+			case 'f': {
```
Mathewnd: This is missing some uppercase options: 'G' 'A' 'F'

no92: Good catch, they are listed in the C spec as well.

PR #1496, opened 2025-10-28
options/ansi: fix stack corruption in do_scanf

Mathewnd:
This fixes a stack corruption that happened when do_scanf had to handle values bigger than 127 in *fmt when handling the '[' specifier. This bug lead to a SIGSEGV when running openjdk in Astral.

### DIFF
--- a/options/ansi/generic/stdio.cpp
+++ b/options/ansi/generic/stdio.cpp
@@ -802,19 +802,23 @@ int do_scanf(H &handler, const char *fmt, __builtin_va_list args) {
 				}
 
 				for (; *fmt != ']'; fmt++) {
+					auto fmt_unsigned = reinterpret_cast<const unsigned char *>(fmt);
+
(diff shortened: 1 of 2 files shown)


options/ansi/generic/stdio.cpp
```diff
@@ -802,19 +802,23 @@ int do_scanf(H &handler, const char *fmt, __builtin_va_list args) {
...
 					}
-					scanset[1 + *fmt] = 1 - invert;
+					scanset[1 + *fmt_unsigned] = 1 - invert;
 				}
 
 				char c = handler.look_ahead();
 				EOF_CHECK(c == '\0');
 				while (c && (!width || count < width)) {
-					if (!scanset[1 + c])
+					unsigned char uc = reinterpret_cast<unsigned char&>(c);
```
no92: A `static_cast` would be more readable here.

PR #1498, opened 2025-10-30
options/rtld: fix SO promotion

Mathewnd:
This fixes SOs not being promoted to global scope when dlopen'd first with RTLD_LOCAL and then with RTLD_GLOBAL.
Fixing this allows for minecraft to run under mlibc.


Mathewnd:
Tests LGTM.

avdgrinten (approved):

PR #1507, opened 2025-11-05
Fix some header pollution

no92:


avdgrinten:
Expressions like `(defined(_POSIX_C_SOURCE) && _POSIX_C_SOURCE >= 202405L)` should probably go into a separate header. One possibility would be adding them to `mlibc-config.h` as we need to include that anyway from many headers.

options/ansi/generic/wchar.cpp
```diff
@@ -1,3 +1,8 @@
+#ifndef _XOPEN_SOURCE
+#define _XOPEN_SOURCE
+#endif
+
+#define _POSIX_C_SOURCE 202405L
```
avdgrinten: This pattern is weird. Why is this not defined using `-D` in the build scripts? Why is an `#ifndef` needed? The build system should know whether the compiler passes `_XOPEN_SOURCE` or not (so we can `#error` if the expectation is not satisfied).

options/posix/include/dlfcn.h
```diff
@@ -37,16 +37,23 @@ void *dlvsym(void *__restrict __handle, const char *__restrict __name, const cha
...
 typedef struct {
 	const char *dli_fname;
 	void *dli_fbase;
 	const char *dli_sname;
 	void *dli_saddr;
 } Dl_info;
 
+typedef Dl_info Dl_info_t;
+
+int dladdr(const void *__ptr, Dl_info_t *__out);
```
Mintsuki, line 51: This function's implementation should probably be moved to the glibc option.

options/posix/include/dlfcn.h
```diff
@@ -82,8 +89,7 @@ struct dl_find_object {
 
 #ifndef __MLIBC_ABI_ONLY
 
-int dladdr(const void *__ptr, Dl_info *__out);
-int dladdr1(const void *__ptr, Dl_info *__out, void **__extra, int __flags);
+int dladdr1(const void *__ptr, Dl_info_t *__out, void **__extra, int __flags);
 int dlinfo(void *__restrict __handle, int __request, void *__restrict __info);
 int _dl_find_object(void *__address, struct dl_find_object *__result);
```
Mintsuki, line 95: Ditto for these functions.

options/ansi/include/wchar.h
```diff
@@ -112,12 +113,17 @@ size_t wcsnrtombs(char *__restrict __mbs, const wchar_t **__restrict __wcs, size
...
 int wcwidth(wchar_t __wc);
 int wcswidth(const wchar_t *__s, size_t __size);
+#endif
+
+#if __MLIBC_POSIX2008
 wchar_t *wcsdup(const wchar_t *__s);
 int wcsncasecmp(const wchar_t *__a, const wchar_t *__b, size_t __size);
 int wcscasecmp(const wchar_t *__a, const wchar_t *__b);
 size_t wcsnlen(const wchar_t *__s, size_t __maxlen);
+#endif /* __MLIBC_POSIX2008 */
```
Mintsuki, line 127: If all these functions are POSIX extensions (as also underlined by the tests being moved there), then their implementation should be in the POSIX option IMO

options/posix/include/fcntl.h
```diff
@@ -66,22 +68,26 @@ struct file_handle {
 #ifndef __MLIBC_ABI_ONLY
 
 #ifdef _GNU_SOURCE
+int fallocate(int __fd, int __mode, off_t __offset, off_t __len);
```
Mintsuki, line 71: This should also be gated behind the Linux option and the implementation moved to the Linux option directory.

mintsuki:
Fixed it all myself!

options/ansi/include/assert.h
```diff
@@ -9,8 +9,8 @@ extern "C" {
 #ifndef __MLIBC_ABI_ONLY
 
 /* NOTE: This is not ISO C. Declared in LSB */
-__attribute__ ((__noreturn__)) void __assert_fail(const char *assertion, const char *file, unsigned int line,
-		const char *function);
+__attribute__ ((__noreturn__)) void __assert_fail(const char *__assertion, const char *__file, unsigned int __line,
+		const char *__function);
```
avdgrinten, line 13: We could also simply remove the argument names, can't we?

tests/posix/rindex.c
```diff
@@ -1,3 +1,7 @@
+#ifndef _DEFAULT_SOURCE
+#define _DEFAULT_SOURCE
+#endif
```
avdgrinten, line 3: Don't we compile all tests with `_DEFAULT_SOURCE`?

avdgrinten (approved):

PR #1508, opened 2025-11-05
misc: Gate the `*64(` function family behind the Linux option

Mintsuki:


mintsuki:
I am gating more *64 functions, i guess we can move them to separate headers later on, but also i think it's fine like this. In any case please waiting before merging.

mintsuki:
Okay, it should be ready now.

PR #1518, opened 2025-11-11
netbsd port: round 0

lzcunt:
These are some various patches with the common theme of reducing Linux-isms and GNU-isms that are needed to port to NetBSD.


abis/lemon/auxv.h
```diff
@@ -0,0 +1,13 @@
+#ifndef _ABIBITS_AUXV_H
```
no92: Lemon got removed in the meantime.

abis/dripos/auxv.h
```diff
@@ -0,0 +1,12 @@
+#ifndef _ABIBITS_AUXV_H
+#define _ABIBITS_AUXV_H
```
no92: dripOS got removed in the meantime.

abis/aero/auxv.h
```diff
@@ -0,0 +1,13 @@
+#ifndef _ABIBITS_AUXV_H
```
no92: aero got removed in the meantime.

options/posix/generic/unistd.cpp
```diff
@@ -236,8 +236,8 @@ int faccessat(int dirfd, const char *pathname, int mode, int flags) {
 }
 
 int fchown(int fd, uid_t uid, gid_t gid) {
-	MLIBC_CHECK_OR_ENOSYS(mlibc::sys_fchownat, -1);
```
no92: Another possibility would be to try to use `sys_fchownat` and fall back to a `sys_fchown`.

options/posix/generic/termios.cpp
```diff
@@ -8,6 +8,7 @@
 #include <bits/ensure.h>
 #include <mlibc/posix-sysdeps.hpp>
 
+#ifdef CBAUD
```
no92: CBAUD is XSI-reserved, technically; the solution here would be to have a sysdep to implement this in the OS-specific way.

meson.build
```diff
@@ -91,6 +91,9 @@ if not headers_only
 	mlibc_c_args = ['-Wno-unused-function', '-D__MLIBC_BUILDING_MLIBC', '-nostdinc', '-fno-builtin', '-ffreestanding']
 	mlibc_link_args = ['-nostdlib']
 	add_project_arguments('-Werror=misleading-indentation', language: ['c', 'cpp'])
+	add_project_arguments('-D_DEFAULT_SOURCE', '-D_GNU_SOURCE', language: ['c', 'cpp'])
```
no92: All arguments but `_GNU_SOURCE` are certainly wrong; `_DEFAULT_SOURCE` should be set by `mlibc-config.h`, and the rest is set to sane values by default.

The `_GNU_SOURCE` case is more interesting on BSDs - I assume it is not defined there, even for C++?

options/posix/generic/unistd.cpp
```diff
@@ -11,6 +11,7 @@
 #include <stdio.h>
 #include <pwd.h>
 #include <sys/stat.h>
+#include <abi-bits/ioctls.h>
```
no92: ABI bits for astral, keyronex and menix are missing from this commit?

no92 (commented):
Gave it a first pass.

options/posix/generic/unistd.cpp
lzcunt: They link Linux abi-bits/ioctls.h

options/posix/generic/unistd.cpp
lzcunt: Any reason would that be preferred?

abis/aero/auxv.h
marv7000: Removed in the meantime

abis/dripos/auxv.h
marv7000: Removed in the meantime

options/elf/include/elf.h
```diff
@@ -701,10 +701,6 @@ typedef struct {
 #define AT_FLAGS 8
 #define AT_ENTRY 9
 #define AT_NOTELF 10
-#define AT_UID 11
```
no92, line 704: I am not entirely happy with this change; x86_64 ELF psABI lists these with these values specifically.

https://refspecs.linuxfoundation.org/elf/x86_64-SysV-psABI.pdf section 3.4.1

options/ansi/generic/file-io.cpp
```diff
@@ -411,7 +410,8 @@ int abstract_file::_save_pos() {
 	if (_type == stream_type::file_like && _bufmode != buffer_mode::no_buffer) {
 		off_t new_offset;
 		auto seek_offset = (off_t(__offset) - off_t(__io_offset));
-		if (int e = io_seek(seek_offset, SEEK_CUR, &new_offset); e) {
```
no92, line 425: The changes in this file seem unrelated.

options/internal/include/bits/types.h
```diff
@@ -60,6 +60,36 @@ typedef __INT64_TYPE__ __mlibc_int64;
 
 /* Fast types (signed). */
 
+#if defined(__NetBSD__)
```
no92: Switching on OS in an mlibc-internal generic header is not the architecture we'd want; if we need to change these types per-OS, they should be turned into an abi-bit.

no92: Not to mention that this change doesn't seem to fit the rest of the commit.

options/posix/generic/pthread.cpp
```diff
@@ -261,7 +261,9 @@ int pthread_attr_setsigmask_np(pthread_attr_t *__restrict attr,
 	attr->__mlibc_sigmaskset = 1;
 
 	// Filter out internally used signals.
+#ifdef SIGCANCEL
```
no92: Why is SIGCANCEL removed? It is not part of the syscall ABI, we carve it out of the RT signal range ourselves for libc-internal use.

sysdeps/managarm/generic/file.cpp
```diff
@@ -2619,6 +2619,14 @@ int Sysdeps<Fchmodat>::operator()(int fd, const char *pathname, mode_t mode, int
 	return 0;
 }
 
+int Sysdeps<Fchown>::operator()(int fd, uid_t owner, gid_t group) {
```
no92: No need to stub out unrelated sysdeps.

options/posix/generic/unistd.cpp
```diff
@@ -228,7 +228,7 @@ int faccessat(int dirfd, const char *pathname, int mode, int flags) {
 }
 
 int fchown(int fd, uid_t uid, gid_t gid) {
-	if(int e = mlibc::sysdep_or_enosys<Fchownat>(fd, "", uid, gid, AT_EMPTY_PATH); e) {
```
no92, line 231: What we could consider doing here is attempt to use Fchownat first, and if that is not provided by the sysdeps, fall back to Fchown.

options/posix/generic/termios.cpp
```diff
@@ -8,6 +8,7 @@
 #include <bits/ensure.h>
 #include <mlibc/all-sysdeps.hpp>
 
+#ifdef CBAUD
```
no92: Just ifdefing these out is broken, as the header still provides these functions.

NetBSD seems to use `struct termios`'s `c_ibaud` and `c_obaud` for this.

meson.build
```diff
@@ -107,6 +107,9 @@ if not headers_only
 	mlibc_c_args = ['-Wno-unused-function', '-D__MLIBC_BUILDING_MLIBC', '-fno-builtin', '-ffreestanding']
 	mlibc_link_args = ['-nostdlib']
 	add_project_arguments('-Werror=misleading-indentation', language: ['c', 'cpp'])
+	add_project_arguments('-D_DEFAULT_SOURCE', '-D_GNU_SOURCE', language: ['c', 'cpp'])
```
no92: Defining all of these is wrong. `_GNU_SOURCE` alone is sufficient, as that effectively means to enable everything, including extensions. The other flags are set as part of `mlibc-config.h` to appropriate values if not provided. Some of what you set here results in stricter modes than necessary.

options/posix/generic/unistd.cpp
lzcunt: Fchownat is provided by NetBSD, AT_EMPTY_PATH isn't. AT_EMPTY_PATH is a linux extension

options/elf/include/elf.h
lzcunt: The only explanation I could come up with is that NetBSD on i386 followed the 1997 i386 SysV psABI published by SCO, and not the later incompatible revisions. They must have kept the same auxiliary vector numbering as i386.

That psABI does not define AT_UID, in fact the value 11 is used for something else entirely. Which is probably why NetBSD adopted the Solaris 2000+ auxvs.

PR #1525, opened 2025-11-14
sysdeps,book: add demo sysdeps and porting mdbook 

Mathewnd:
Adds a demo sysdep and a porting guide.
Supersedes #1211

### DIFF
--- a/book/src/porting/choosing_a_compiler.md
+++ b/book/src/porting/choosing_a_compiler.md
@@ -1,19 +1,8 @@
 # Choosing a compiler
 
-To compile [`mlibc`](https://github.com/managarm/mlibc) and any userspace programs which link against it, you'll need a suitable compiler. Roughly speaking you have two choices:
+To compile [`mlibc`](https://github.com/managarm/mlibc) and any userspace programs which link against it, you'll need a suitable compiler. For that, you will need to build a full [OS Specific Toolchain](https://wiki.osdev.org/OS_Specific_Toolchain). Using your host's compiler or a generic toolchain could lead into strange build failures and is not recommended nor supported.
 
-1. **Build a full [OS Specific Toolchain](https://wiki.osdev.org/OS_Specific_Toolchain).**
-    
-    With this, you'll have a compiler that implicitly links against (m)libc, so we recommend doing this _after_ you have a basic port working.
-
-1. **Use a generic ELF toolchain.**
-
-    If compiling your kernel with a generic toolchain like `x86_64-unknown-elf-gcc` or `riscv64-elf-gcc`, you may re-use the same compiler for this.
-
-    Note that when compiling userspace programs, you must manually provide a number of command line arguments (such as the location of your compiled libc and any headers).
-
-
-For [`mlibc-demo-os`](https://github.com/64/mlibc-demo-os), we'll rely on a generic `riscv64-elf-gcc` toolchain provided by our distro.
+After that, you'll have a compiler that implicitly links against (m)libc. For [`mlibc-demo-os`](https://github.com/64/mlibc-demo-os), we'll rely on the host's `riscv64-linux-gnu-gcc` compiler to avoid setting up a hosted OS-specific compiler for a brief demo. The `CC` variable will be used to indicate the C compiler in Makefiles.
 
 ## Creating a Meson cross file
 
--- a/book/src/porting/implementing_sysdeps.md
+++ b/book/src/porting/implementing_sysdeps.md
@@ -14,7 +14,9 @@ Now, create the `sysdeps/demo` folder. At minimum it should contain a `meson.bui
 {{#include ../../../sysdeps/demo/meson.build:sources-and-includes}}
 ```
 
-The `sysdep_supported_options` tells mlibc which 'options' your sysdeps supports. For example, the `unistd.h` header is only available when the POSIX option is enabled.
+The `sysdep_supported_options` tells mlibc which 'options' your sysdeps support. For example, the `unistd.h` header is only available when the POSIX option is enabled. 
+
+While `sysdep_supported_options` only states which options are supported to be used by the sysdeps, you are free to disable options with meson options at `meson setup` time. The default behavior is to enable supported options.
 
 ## Preparing to build mlibc
 
@@ -24,7 +26,7 @@ Now, from a fresh clone of `mlibc` you can run:
 $ meson setup --cross-file=path/to/your.cross-file -Ddefault_library=static -Dbuildtype=debug build
 ```
 
-The `-Ddefault_library=static` tells meson to only produce a statically linked library (`libc.a`). We *strongly* recommend getting statically linked binaries to work before dynamically linked ones.
+The `-Ddefault_library=static` tells meson to only produce a statically linked library (`libc.a`). We suggest getting statically linked binaries to work before dynamically linked ones.
 
 Now run `ninja -C build` to start the build. You'll likely get a large number of compilation errors at this point, and we'll work on fixing these in the next sections.
 
@@ -34,7 +36,7 @@ Many of the compile errors are because we need to tell mlibc about the ABI betwe
 
 We _strongly_ recommend re-using an existing ABI where possible rather than creating your own, because many programs assume a particular ABI and will fail to compile or subtly break if you choose a different one.
 
-For the demo OS, we'll symlink each ABI header to Linux's definition in `sysdeps/linux/include/abi-bits`. Then, we'll add the following to our `meson.build`:
+For the demo OS, we'll symlink each ABI header to Linux's definition in `abis/linux`. Then, we'll add the following to our `meson.build`:
 
 ```meson
 {{#include ../../../sysdeps/demo/meson.build:abi-includes}}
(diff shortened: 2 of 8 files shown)


notes.txt
```diff
@@ -0,0 +1,9 @@
+setup: make cross file pointing to compiler
```
no92: Accidental commit?

sysdeps/demo/meson.build
```diff
@@ -0,0 +1,66 @@
+# ANCHOR: sources-and-includes
+sysdep_supported_options = {
+	'posix': false,
+	'linux': false,
+	'glibc': false,
+	'bsd': false,
+}
+
+rtld_sources += files(
+	'sysdeps.cpp',
```
no92, line 10: Broken indentation in this and some other `meson.build` files.

options/rtld/generic/main.cpp
```diff
@@ -16,6 +16,7 @@
 
 #include "elf.hpp"
 #include "linker.hpp"
+#include "mlibc/ansi-sysdeps.hpp"
```
no92: This should use `#include <mlibc/ansi-sysdeps.hpp>`, like the `rtld-sysdeps.hpp` above. Is this even needed though, or is it an LSP artifact?

options/elf/generic/startup.cpp
```diff
@@ -11,6 +11,7 @@
 #include <mlibc/debug.hpp>
 
 #include <frg/string.hpp>
+#include <errno.h>
```
no92: LSP artifact?

options/elf/generic/startup.cpp
```diff
@@ -49,6 +50,7 @@ void parse_exec_stack(void *opaque_sp, exec_stack_data *data) {
 
 // TODO: This does not have to be here; we could also move it to options/internal.
 void set_startup_data(int argc, char **argv, char **envp) {
+#if defined(_GNU_SOURCE)
```
no92: Why is `_GNU_SOURCE` not defined for a C++ file? This seems broken.

book/src/porting/choosing_a_compiler.md
```diff
@@ -0,0 +1,26 @@
...
+
+To compile [`mlibc`](https://github.com/managarm/mlibc) and any userspace programs which link against it, you'll need a suitable compiler. Roughly speaking you have two choices:
+
+1. **Build a full [OS Specific Toolchain](https://wiki.osdev.org/OS_Specific_Toolchain).**
+    
+    With this, you'll have a compiler that implicitly links against (m)libc. This is the recommended choice, as using your host's or a generic toolchain could lead into strange build failures.
+
+1. **Temporarily using your host's toolchain or a generic ELF toolchain.**
+
+    If compiling your kernel with a generic toolchain like `x86_64-unknown-elf-gcc` or `riscv64-elf-gcc` or even your host's, you may re-use the same compiler for this.
```
no92: We should probably remove this line, and add a warning about why this is inherently broken (and therefore not recommended or supported) instead.

book/src/porting/choosing_a_compiler.md
```diff
@@ -0,0 +1,26 @@
...
+
+    Note that when compiling userspace programs, you must manually provide a number of command line arguments (such as the location of your compiled libc and any headers).
+
+
+For [`mlibc-demo-os`](https://github.com/64/mlibc-demo-os), we'll rely on the host's `riscv64-linux-gnu-gcc` compiler. The `CC` variable will be used to indicate the C compiler.
```
no92:
```suggestion
For [`mlibc-demo-os`](https://github.com/64/mlibc-demo-os), we'll rely on the host's `riscv64-linux-gnu-gcc` compiler to avoid setting up a hosted OS-specific compiler for a brief demo. The `CC` variable will be used to indicate the C compiler in Makefiles.
```

book/src/porting/implementing_sysdeps.md
```diff
@@ -0,0 +1,153 @@
...
+    # ...
+```
+
+Now, create the `sysdeps/demo` folder. At minimum it should contain a `meson.build` file where we'll declare all the `.cpp` files and select header files to install:
+
+```meson
+{{#include ../../../sysdeps/demo/meson.build:sources-and-includes}}
+```
+
+The `sysdep_supported_options` tells mlibc which 'options' your sysdeps supports. For example, the `unistd.h` header is only available when the POSIX option is enabled.
```
no92:
```suggestion
The `sysdep_supported_options` tells mlibc which 'options' your sysdeps support. For example, the `unistd.h` header is only available when the POSIX option is enabled. While `sysdep_supported_options` only states which options are supported to be used by the sysdeps, you are free to disable options with meson options at `meson setup` time. The default behavior is to enable supported options.
```

book/src/porting/implementing_sysdeps.md
```diff
@@ -0,0 +1,153 @@
...
+
+## Preparing to build mlibc
+
+Now, from a fresh clone of `mlibc` you can run:
+
+```
+$ meson setup --cross-file=path/to/your.cross-file -Ddefault_library=static -Dbuildtype=debug build
+```
+
+The `-Ddefault_library=static` tells meson to only produce a statically linked library (`libc.a`). We *strongly* recommend getting statically linked binaries to work before dynamically linked ones.
```
no92: Not sure if we should strongly recommend (instead of just suggesting) static linkage first, it's not that hard to go to dynamic linking anyways.

book/src/porting/implementing_sysdeps.md
```diff
@@ -0,0 +1,153 @@
...
+
+We _strongly_ recommend re-using an existing ABI where possible rather than creating your own, because many programs assume a particular ABI and will fail to compile or subtly break if you choose a different one.
+
+For the demo OS, we'll symlink each ABI header to Linux's definition in `abis/linux`. Then, we'll add the following to our `meson.build`:
+
+```meson
+{{#include ../../../sysdeps/demo/meson.build:abi-includes}}
+```
+
+This is not an exhaustive list of ABI headers but should be enough to get started.
```
no92, line 45: We should also note that the Linux ABI will be silently changed if it turns out that it is wrong - you end up subscribing to what amounts to ABI breaks if that (hopefully only rarely) happens.

book/src/porting/implementing_sysdeps.md
```diff
@@ -0,0 +1,153 @@
...
+
+## Implementing a C++ entry point
+
+Now, we'll implement the C++ entry point that we call from `crt1.S`. 
+
+```cpp
+{{#include ../../../sysdeps/demo/entry.cpp}}
+```
+
+The call to `__dlapi_enter` is used to perform initialisation in statically linked executables (but is a no-op in dynamically linked ones). For example, any global constructors in the program will be called from here.
```
no92, line 71: Maybe we should mention somewhere that people should take a look at the specs and implement some required but often-forgotten-about things like auxv?

book/src/porting/implementing_sysdeps.md
```diff
@@ -0,0 +1,153 @@
...
+- `mlibc::sys_exit`
+- `mlibc::sys_close`
+- `mlibc::sys_futex_wake`
+- `mlibc::sys_futex_wait`
+- `mlibc::sys_read`
+- `mlibc::sys_open`
+- `mlibc::sys_vm_map`
+- `mlibc::sys_clock_get`
+
+Note that many of these functions are declared as weak symbols. You _must_ include the relevant headers (e.g `<mlibc/all-sysdeps.hpp>`) before providing definitions otherwise these won't be emitted as weak symbols and you'll see strange errors.
```
no92: Maybe we should better describe the breakage?
> Diverging from the declared sysdep signature will result in different mangling of the function (because they are not declared extern "C", but have C++ linkage instead), which makes it look like the sysdep is missing - this is a silent breakage that likely does not result in compiler errors or even warnings.

book/src/porting/next_steps.md
```diff
@@ -0,0 +1,52 @@
+# Next steps
+
+## Dynamic linking
+
+So far you have compiled and run a statically linked executable. If you desire dynamic linking, it is not much more work in mlibc compared to static linking. 
+
+To get it to work, you will need to reconfigure mlibc without the `-Ddefault_library=static` option, load *both* the ELF and its intepreter (stored in `PT_INTERP`), and jump to the *interpreter's* entry point.
```
no92:
```suggestion
To get it to work, you will need to reconfigure mlibc without the `-Ddefault_library=static` option, load *both* the ELF and its interpreter (stored in `PT_INTERP`), and jump to the *interpreter's* entry point.
```

book/src/porting/next_steps.md
```diff
@@ -0,0 +1,52 @@
...
+
+The list of sysdeps for every option can be found under its include directory. For example, the sysdeps for the POSIX option are declared in `options/posix/include/mlibc/posix-sysdeps.hpp`. Make sure your definitions match the ones in the header, as mlibc won't be able to find them otherwise.
+
+## Enabling more options
+
+For the demo sysdeps, only the bare minimum is enabled. However, most software you would want to port over will need more options enabled. The toggleable options in mlibc are:
+
+- POSIX option, for APIs defined by posix like `fork` and `wait`
+- Linux option, for Linux-specific system calls like `epoll_create` and `statx`. Note that this option requires you to provide Linux kernel headers.
+- glibc option, for glibc-specific extensions like `backtrace` and `getopt_long`. Make sure that your compiler defines `_GNU_SOURCE` if you enable this option.
```
no92: I'm not sure whether compilers should define `_GNU_SOURCE` by default in C mode?

book/src/porting/next_steps.md
```diff
@@ -0,0 +1,52 @@
...
+elif host_machine.system() == 'demo'
+	internal_conf.set10('CONFIG_NAME', true)
+    # ...
+```
+
+## Porting software
+
+With mlibc working, you are now ready to start [porting software](https://wiki.osdev.org/Cross-Porting_Software). For your sanity, use a compiler which targets your OS and has a sysroot directory.
+
+If you run into build failures, make sure that you have enabled the options needed by the application as well as defined the ABI correctly. If you truly believe something is an mlibc bug, please [open an issue](https://github.com/managarm/mlibc/issues/new).
```
no92:
```suggestion
If you run into build failures, make sure that you have enabled the options needed by the application as well as defined the ABI correctly. If you truly believe something is an mlibc bug, please reach out to us [on Discord](https://discord.gg/7WB6Ur3) in `#mlibc-dev` or [open an issue](https://github.com/managarm/mlibc/issues/new).
```

no92 (approved):
Except for the one remaining open comment, LGTM.

PR #1528, opened 2025-11-16
options/posix: add some more feature test macros

ElectrodeYT:
os-test shows we don't define these even though we support everything needed for them.

### DIFF
--- a/options/posix/include/unistd.h
+++ b/options/posix/include/unistd.h
@@ -37,6 +37,17 @@ extern "C" {
 #define _POSIX_THREAD_SAFE_FUNCTIONS _POSIX_VERSION
 #define _POSIX_MONOTONIC_CLOCK 0
 
+/* mmap, msync, munmap */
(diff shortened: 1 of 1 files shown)


ElectrodeYT:
> The comment style is not correct. Also, are these macros supposed to be defined to the posix version or just some numeric value?

As far as I can tell, has to be at least some not-zero number, but internet says its commonly defined to the POSIX version,
for comment style, how do you think it could be done while ideally keeping what functions/features are needed for each macro? Might be useful when going through them in the future if we decide to add things such as only enabling the ones where the sysdeps required for each function are present

no92:
One line from the spec is a bit annoying:
> Runtime checks with fpathconf( ), pathconf( ), or sysconf shall indicate that the option is supported.
This applies to `_POSIX_MEMLOCK`, `_POSIX_MEMLOCK_RANGE` and `_POSIX_SYNCHRONIZED_IO`.

PR #1535, opened 2025-11-20
options/ansi: correctly support printf positional arguments

no92:
Needs to bump frigg once managarm/frigg#108 is merged.


Dennisbonke (approved):
LGTM

PR #1536, opened 2025-11-20
options/bsd: move getifaddrs from linux to bsd

marv7000:
`getifaddrs` is also available on different BSDs. The function signatures and `struct ifaddrs` aligns with the current mlibc declaration.


no92:
Given that it is also supported on Linux, I am not sure that moving it is the best play - maybe we should enable it if `__MLIBC_LINUX_OPTION || __MLIBC_BSD_OPTION`?

PR #1544, opened 2025-11-24
[ABI break] Change jmp_buf to match sigjmp_buf

### LINKED ISSUE #1410: ABI Mismatch: jmpbuf and sigjmpbuf do not match
Both musl and glibc assume `sigjmp_buf` and `jmp_buf` are the same, and the `libpng` API assumes this as well. This is not a problem for most applications, but `GTK4` assumes that these are the same and passes this `libpng` defined buffer to both `setjmp` and `sigsetjmp`, giving compiler errors on mlibc.

no92:
Fixed #1410.


Dennisbonke (approved):
LGTM

PR #1546, opened 2025-11-24
C++-style sysdeps without weak symbols

no92:
This PR reworks the sysdep mechanism to avoid using weak symbols. Instead, we use tag-based dispatch and build on top of that using concepts. This allows for compile-time detection of the following issues:
- Calling a sysdep that is not provided
- Implementing a sysdep that is not advertised
- Advertising a sysdep that is not implemented

Apart from this, the `sysdep<Tag>` and `sysdep_or_enosys<Tag>` helpers streamline use of sysdeps quite nicely.

### DIFF
--- a/book/src/porting/implementing_sysdeps_p2.md
+++ b/book/src/porting/implementing_sysdeps_p2.md
@@ -2,7 +2,7 @@
 
 Let's try building mlibc:
 
-```
+```sh
 $   meson \
     setup \
     --cross-file=path/to/your.cross-file \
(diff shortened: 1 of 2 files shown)


options/internal/include/mlibc/sysdep-tags.hpp
```diff
@@ -0,0 +1,691 @@
...
+#include <concepts>
+#include <mlibc-config.h>
+#include <mlibc/fsfd_target.hpp>
+#include <stddef.h>
+#include <stdint.h>
+
+// [[noreturn]] void sys_exit(int status);
+struct Exit { static constexpr bool is_noreturn = true; using return_type = void; };
+// [[noreturn, gnu::weak]] void sys_thread_exit();
+struct ThreadExit { static constexpr bool is_noreturn = true; using return_type = void; };
```
avdgrinten: Where is `is_noreturn` and `return_type` used?

no92: As discussed on Discord, the concepts in `all-sysdeps.hpp` use these.

PR #1547, opened 2025-11-25
options/posix: accept IP addresses in gethostbyname

sasdallas:
man page for `gethostbyname()`:
> The gethostbyname() function returns a structure of type hostent for the given host name.  Here name is either a  hostname  or
       an  IPv4 address in standard dot notation (as for inet_addr(3)).  If name is an IPv4 address, no lookup is performed and geth‐
       ostbyname() simply copies name into the h_name field and its struct in_addr equivalent into the h_addr_list[0]  field  of  the
       returned  hostent  structure.   If  name  doesn't end in a dot and the environment variable HOSTALIASES is set, the alias file
       pointed to by HOSTALIASES will first be searched for name (see hostname(7) for the file format).  The current domain  and  its
       parents are searched unless name ends in a dot.

`gethostbyname()` should also perform an IP lookup on the specified address

### DIFF
--- a/tests/posix/gethostbyname.c
+++ b/tests/posix/gethostbyname.c
@@ -1,0 +1,25 @@
+#include <netdb.h>
+#include <assert.h>
+#include <string.h>
+#include <sys/socket.h>
+#include <netinet/in.h>
+#include <arpa/inet.h>
+
+int main() {
+    struct hostent *ent;
+    char ip[INET_ADDRSTRLEN];
+    
+    ent = gethostbyname("dns.google");
+    assert(ent);
+    assert(ent->h_addrtype == AF_INET);
+    assert(inet_ntop(AF_INET, ent->h_addr, ip, sizeof(ip)));
+    assert(strcmp(ip, "8.8.8.8") == 0 || strcmp(ip, "8.8.4.4") == 0);
+    
+    ent = gethostbyname("8.8.8.8");
+    assert(ent);
+    assert(ent->h_addrtype == AF_INET);
+    assert(inet_ntop(AF_INET, ent->h_addr, ip, sizeof(ip)));
+    assert(strcmp(ip, "8.8.8.8") == 0);
+
+    return 0;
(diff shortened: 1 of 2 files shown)


options/posix/generic/netdb.cpp
```diff
@@ -290,8 +290,9 @@ struct hostent *gethostbyname(const char *name) {
 	struct mlibc::lookup_result buf;
 	frg::string<MemoryAllocator> canon{getAllocator()};
 	int ret = 0;
-	if ((ret = mlibc::lookup_name_hosts(buf, name, canon, AF_UNSPEC)) <= 0)
-		ret = mlibc::lookup_name_dns(buf, name, canon, AF_UNSPEC);
+	if ((ret = mlibc::lookup_name_ip(buf, name, AF_UNSPEC)) <= 0)
```
no92: The use of `AF_UNSPEC` here is incorrect, as we should only ever deal with IPv4 addresses here, so `AF_INET`.

This behavior seems to be a glibc (?) extension, as POSIX 2001 (last issue before this function got removed) states:
> The name argument of gethostbyname() shall be a node name; the behavior of gethostbyname() when passed a numeric address string is unspecified. For IPv4, a numeric address string shall be in the dotted-decimal notation described in inet_addr().

I am personally a bit torn on whether we should default to the glibc extension behavior or whether we should guard that behind the glibc option.

sasdallas: @no92 apologies. In my personal opinion, I don't see how defaulting to the glibc ext. could in any way affect applications so I don't see any reason to guard it.

tests/posix/gethostbyname.c
```diff
@@ -0,0 +1,30 @@
+#include <netdb.h>
+#include <assert.h>
+#include <string.h>
```
no92, line 3: Missing includes of `<netinet/in.h>` and `<sys/socket.h>`.

PR #1552, opened 2025-11-26
book: add proper toolchain section

Mathewnd:
Enable the POSIX option for the demo sysdep and add a section to the book about building a toolchain.

### DIFF
--- a/book/src/porting/implementing_sysdeps_p1.md
+++ b/book/src/porting/implementing_sysdeps_p1.md
@@ -1,0 +1,88 @@
+# Implementing your sysdeps (part 1)
+
+As we need libc headers to build the toolchain, we will first implement enough of the sysdeps to install the headers.
+
+## Creating a Meson cross file
+
+mlibc uses a build system called [Meson](https://mesonbuild.com/). When cross-compiling, you must tell meson about your compiler and target via a [_cross file_](https://mesonbuild.com/Cross-compilation.html).
+
+For [`mlibc-demo-os`](https://github.com/64/mlibc-demo-os), our cross file looks like this:
+
+```toml
+[host_machine]
+system = 'demo'
+cpu_family = 'riscv64'
+cpu = 'riscv64'
+endian = 'little'
+[binaries]
+c = 'riscv64-demo-gcc'
+cpp = 'riscv64-demo-g++'
+ar = 'riscv64-demo-ar'
+strip = 'riscv64-demo-strip'
+ld = 'riscv64-demo-ld'
+[built-in options]
+c_args = ['-march=rv64gc', '-mabi=lp64d']
+cpp_args = ['-march=rv64gc', '-mabi=lp64d']
+```
+
+## Telling mlibc about your sysdeps
+
+In mlibc's top level `meson.build` file, you'll see a large `if`/`else` chain that selects the right sysdeps subdirectory based on the system name. We'll add ourselves here:
+```meson
+{{#include ../../../meson.build:demo-sysdeps}}
+    # ...
+```
+
+Now, create the `sysdeps/demo` folder. At minimum it should contain a `meson.build` file. Here, all the source and include files are declared for the sysdep. For now, we will only add this:
+
+```meson
+{{#include ../../../sysdeps/demo/meson.build:supported-options}}
+```
+
+The `sysdep_supported_options` tells mlibc which 'options' your sysdeps support. For example, the `unistd.h` header is only available when the POSIX option is enabled. 
+
+While `sysdep_supported_options` only states which options are supported to be used by the sysdeps, you are free to disable options with meson options at `meson setup` time. The default behavior is to enable supported options.
+
+## Selecting ABI headers
+
+Your userland needs an ABI to communicate with your kernel. For example, we must define the layout of `struct stat`.
+
+Re-using an existing ABI is _strongly_ recommended. Many programs assume a particular ABI and will fail to compile or subtly break if you choose a different one (This is the program's fault and not mlibc's or your sysdeps').
+
+Note that using an existing ABI does _not_ imply you have to implement OS specific features. For example, even if you use the Linux ABI you do not have to implement Linux specific features.
+
+For the demo OS, we'll symlink each ABI header to Linux's definition in `abis/linux`. This is so we don't have to define a whole new ABI for a brief demo. Then, we'll add the following to our `meson.build`:
+
+```meson
+{{#include ../../../sysdeps/demo/meson.build:abi-includes}}
+```
+
+This is not an exhaustive list of ABI headers but should be enough to get started.
+
+Note that the Linux ABI will be silently changed if it turns out that it is wrong - you end up subscribing to what amounts to ABI breaks if that (hopefully only rarely) happens.
+
+For this reason, it is also _strongly_ recommended to pin the commit or version of mlibc which will be used, as otherwise there could be changes which silently break your compiled software.
+
+If subscribing to the Linux ABI is not desired, you can still fork it by creating a copy under `abis/$os/` and symlinking to it instead of the Linux ABI.
+
+## Configuring mlibc for headers only 
+
+Now, you can run:
+
+```
+$ meson \
+    setup \
+    --cross-file=path/to/your.cross-file \
+    --prefix=/usr \
+    -Dheaders_only=true \
+    headers-build
+```
+
+The `-Dheaders_only=true` option tells meson that we only want the headers and not a full libc.
+
+To install the headers into the sysroot, run: 
+```
+DESTDIR=${SYSROOT_DIR} ninja -C headers-build install
+```
+
+We are now ready to start building the toolchain!
--- a/book/src/porting/implementing_sysdeps_p2.md
+++ b/book/src/porting/implementing_sysdeps_p2.md
@@ -1,58 +1,30 @@
-# Implementing your sysdeps
+# Implementing your sysdeps (part 2)
 
-## Telling mlibc about your sysdeps
-
-In mlibc's top level `meson.build` file, you'll see a large `if`/`else` chain that selects the right sysdeps subdirectory based on the system name. We'll add ourselves here:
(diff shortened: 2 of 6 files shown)


book/src/porting/implementing_sysdeps_p1.md
```diff
@@ -0,0 +1,79 @@
...
+
+For the demo OS, we'll symlink each ABI header to Linux's definition in `abis/linux`. Then, we'll add the following to our `meson.build`:
+
+```meson
+{{#include ../../../sysdeps/demo/meson.build:abi-includes}}
+```
+
+This is not an exhaustive list of ABI headers but should be enough to get started.
+
+Note that the Linux ABI will be silently changed if it turns out that it is wrong - you end up subscribing to what amounts to ABI breaks if that (hopefully only rarely) happens.
```
Dennisbonke, line 62: Maybe add a note that this is why you pin a commit and don't blindly follow master?

book/src/porting/toolchain.md
```diff
@@ -0,0 +1,69 @@
...
+# --disable-multilib disables building a multilib gcc
+    CFLAGS_FOR_TARGET="-march=rv64gc -mabi-lp64d" \
+    CXXFLAGS_FOR_TARGET="-march=rv64gc -mabi-lp64d" \
+../configure \
+    --target=riscv64-demo \
+    --prefix=/usr \
+    --with-sysroot="${SYSROOT_DIR}" \
+    --enable-languages=c,c++ \
+    --enable-threads=posix \
+    --disable-multilib \
```
Dennisbonke, line 54: `--disable-bootstrap` comes to mind here? Also `--disable-fixincludes` seems to be a thing. No shame in passing that too. Please don't mindlessly copy this suggestion, I see that LFS does it for the final GCC build but omits the flags for the first 2 passes, so maybe we can't use the flags for making a cross compiler here.

book/src/porting/implementing_sysdeps_p2.md
```diff
@@ -0,0 +1,127 @@
...
+
+Now run `ninja -C build` to start the build. You'll likely get a large number of compilation errors at this point, and we'll work on fixing these in the next sections.
+
+## Implementing (S)crt1
+
+We must provide a definition of the ELF entry point (traditionally named `_start`). Each ISA tends to require its own definition written in assembly; for example, RISC-V targets must initialise the `gp` register before jumping to C++.
+
+Traditionally the file that defines `_start` is called `crt1.S` (or `Scrt1.S` for position independent executables). This produces an object file which has to be linked into every application.
+
+Part of configuring an [OS Specific Toolchain](https://wiki.osdev.org/OS_Specific_Toolchain) is [specifying the location of `(S)crt1.o`](https://wiki.osdev.org/OS_Specific_Toolchain#Start_Files_Directory) so that it is linked automatically.
```
Dennisbonke: Should we link to this one or to osdev.wiki?

book/src/porting/implementing_sysdeps_p2.md
```diff
@@ -0,0 +1,127 @@
...
+```bash
+$ riscv64-demo-gcc -march=rv64gc -mabi=lp64d helloworld.c -o helloworld
+```
+
+## Troubleshooting
+
+Something not working? Here are some common issues to look out for:
+
+- Your kernel isn't loading the ELF file correctly. Double check that you're loading the contents from file to the right addresses and zeroing the uninitialised portions.
+- Your `sys_anon_allocate` function is broken. Try commenting out `sys_anon_free` and see if that helps. Try printing the addresses to see if you're getting unique allocations.
```
Dennisbonke, line 119: Another good thing to verify is that memory returned by anon_allocate and friends is expected to be zeroed by mlibc.

book/src/porting/implementing_sysdeps_p1.md
```diff
@@ -0,0 +1,79 @@
...
+
+The `sysdep_supported_options` tells mlibc which 'options' your sysdeps support. For example, the `unistd.h` header is only available when the POSIX option is enabled. 
+
+While `sysdep_supported_options` only states which options are supported to be used by the sysdeps, you are free to disable options with meson options at `meson setup` time. The default behavior is to enable supported options.
+
+## Selecting ABI headers
+
+Your userland needs an ABI to communicate with your kernel. For example, we must define the layout of `struct stat`.
+
+We _strongly_ recommend re-using an existing ABI where possible rather than creating your own, because many programs assume a particular ABI and will fail to compile or subtly break if you choose a different one.
```
Dennisbonke: Do note that technically the programs are at fault there. And maybe explain why you're picking the Linux ABI (and that using the Linux ABI does not necessarily force you to implement Linux specific features.

book/src/porting/implementing_sysdeps_p2.md
```diff
@@ -0,0 +1,127 @@
...
+- `mlibc::sys_vm_map`
+- `mlibc::sys_clock_get`
+
+Note that many of these functions are declared as weak symbols. You _must_ include the relevant headers (e.g `<mlibc/all-sysdeps.hpp>`) before providing definitions.
+
+Diverging from the declared sysdep signature will result in different mangling of the function (because they are not declared extern "C", but have C++ linkage instead), which makes it look like the sysdep is missing - this is a silent breakage that likely does not result in compiler errors or even warnings.
+
+Most sysdep functions return an integer error code (0 for success, or a value that matches the sysdep's abi-bits errno definitions on failure) and return data via out parameters. Note that sysdeps shouldn't set errno directly - mlibc will set it from the error code you return.
+
+As a general strategy, it's a good idea to stub whatever's required to make things compile, and then add proper implementations later. For example:
```
Dennisbonke, line 83: Maybe note that for non required sysdeps, if they aren't provided, it'll log that it isn't implemented (using sys_libc_log) and returns ENOSYS?

Dennisbonke (commented):
A few suggestions here and there.

book/src/porting/toolchain.md
Mathewnd: `--disable-bootstrap` is not useful in this case and could lead to bugs. `--disable-fixincludes` is not needed as the patches add the demo OS to the no-op list in mkfixinc.sh

book/src/porting/implementing_sysdeps_p2.md
Mathewnd: It is already talked about in the "next steps" section under "implementing more sysdeps"

book/src/porting/choosing_a_compiler.md
```diff
@@ -1,6 +1,6 @@
 # Choosing a compiler
 
-To compile [`mlibc`](https://github.com/managarm/mlibc) and any userspace programs which link against it, you'll need a suitable compiler. For that, you will need to build a full [OS Specific Toolchain](https://wiki.osdev.org/OS_Specific_Toolchain). Using your host's compiler or a generic toolchain could lead into strange build failures and is not recommended nor supported.
```
no92, line 3: I am not convinced that using a fork of the osdev.org wiki article should be preferred here, given that the OG wiki is active and publically editable.

book/src/porting/implementing_sysdeps_p1.md
```diff
@@ -0,0 +1,83 @@
...
+
+For the demo OS, we'll symlink each ABI header to Linux's definition in `abis/linux`. This is so we don't have to define a whole new ABI for a brief demo. Then, we'll add the following to our `meson.build`:
+
+```meson
+{{#include ../../../sysdeps/demo/meson.build:abi-includes}}
+```
+
+This is not an exhaustive list of ABI headers but should be enough to get started.
+
+Note that the Linux ABI will be silently changed if it turns out that it is wrong - you end up subscribing to what amounts to ABI breaks if that (hopefully only rarely) happens.
```
no92, line 62: Maybe we should mention that copying (so basically forking) the ABI into your own ABI under `abis/$os/` is possible if subscribing to Linux fixes is not desired.

book/src/porting/toolchain.md
```diff
@@ -0,0 +1,69 @@
+# Building the toolchain
+
+To compile [`mlibc`](https://github.com/managarm/mlibc) and any userspace programs which link against it, you'll need a suitable compiler. For that, you will need to build a full [OS Specific Toolchain](https://osdev.wiki/OS_Specific_Toolchain). Using your host's compiler or a generic toolchain could lead into strange build failures and is not recommended nor supported.
```
no92: Same wiki concern as above.

book/src/porting/implementing_sysdeps_p1.md
```diff
@@ -0,0 +1,83 @@
...
+cpu_family = 'riscv64'
+cpu = 'riscv64'
+endian = 'little'
+[binaries]
+c = 'riscv64-demo-gcc'
+cpp = 'riscv64-demo-g++'
+ar = 'riscv64-demo-ar'
+strip = 'riscv64-demo-strip'
+ld = 'riscv64-demo-ld'
+```
```
no92, line 26: It might be preferable to provide `CFLAGS` and `CXXFLAGS` via the cross file:

```ini
[built-in options]
c_args = ['-march=rv64gc', '-mabi=lp64d']
cpp_args = ['-march=rv64gc', '-mabi=lp64d']
``

File: book/src/porting/implementing_sysdeps_p2.md [hunk 11]
```diff
@@ -1,58 +1,32 @@
...
-```meson
-{{#include ../../../meson.build:demo-sysdeps}}
-    # ...
 ```
-
-Now, create the `sysdeps/demo` folder. At minimum it should contain a `meson.build` file where we'll declare all the `.cpp` files and select header files to install:
-
-```meson
-{{#include ../../../sysdeps/demo/meson.build:sources-and-includes}}
+$ CFLAGS="-march=rv64gc -mabi=lp64d" \
```
no92: Providing CFLAGS and CXXFLAGS should be done in the cross file.

no92 (approved):

PR #1553, opened 2025-11-26
pthread_spin family

sasdallas:
Integrates the pthread_spin family

### DIFF
--- a/tests/posix/pthread_spin.c
+++ b/tests/posix/pthread_spin.c
@@ -1,0 +1,47 @@
+#include <assert.h>
+#include <pthread.h>
+#include <stddef.h>
+#include <errno.h>
+#include <stdio.h>
+
+pthread_spinlock_t lock;
+int variable = 0;
+
+void *worker(void* arg) {
+	(void)arg;
+
+	int ret = pthread_spin_trylock(&lock);
+	if (!variable) {
+		assert(ret == EBUSY);
+	} else {
+		assert(ret == 0);
+		variable = 0;
+		pthread_spin_unlock(&lock);
+	}
+
+	return NULL;
+}
+
+int main() {
+	assert(!pthread_spin_init(&lock, PTHREAD_PROCESS_PRIVATE));
+	assert(!pthread_spin_lock(&lock));
+	
+	/* glibc does not implement this optional POSIX behavior and hangs */
+#ifndef USE_HOST_LIBC
+	assert(pthread_spin_lock(&lock) == EDEADLK);
+#endif
+
+	pthread_t thread;
+	assert(!pthread_create(&thread, NULL, worker, NULL));
+	pthread_join(thread, NULL);
+
+	variable = 1;
+	assert(!pthread_spin_unlock(&lock));
+	assert(!pthread_create(&thread, NULL, worker, NULL));
+	pthread_join(thread, NULL);
+
+	assert(variable == 0);
+
+	pthread_spin_destroy(&lock);
+	return 0;
+}
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -50,6 +50,7 @@ all_test_cases = [
 	'posix/pthread_thread_local',
 	'posix/pthread_attr',
 	'posix/pthread_schedparam',
+	'posix/pthread_spin',
 	'posix/pwd',
 	'posix/fdopen',
 	'posix/fopencookie',


options/posix/generic/pthread.cpp
```diff
@@ -1426,3 +1427,46 @@ int pthread_getcpuclockid(pthread_t, clockid_t *) {
...
+
+int pthread_spin_destroy(pthread_spinlock_t*) {
+	return 0;
+}
+
+int pthread_spin_lock(pthread_spinlock_t *__lock) {
+	unsigned int desired = mlibc::this_tid();
+	unsigned int expected = 0;
+
+	if (__atomic_load_n(&__lock->lock, __ATOMIC_RELAXED) == desired)
```
no92: This `if` case can be rolled into the while below, as on failure to compare-and-exchange we get a return value of zero and the lock's value is written into `expected`; there, we could inspect the value of expected instead.

options/posix/generic/pthread.cpp
```diff
@@ -1426,3 +1427,46 @@ int pthread_getcpuclockid(pthread_t, clockid_t *) {
...
+
+	return 0;
+}
+
+int pthread_spin_trylock(pthread_spinlock_t *__lock) {
+	unsigned int desired = mlibc::this_tid();
+	unsigned int expected = 0;
+
+	if (__atomic_load_n(&__lock->lock, __ATOMIC_RELAXED) == desired)
+		return EDEADLK;
```
no92: This is not a supported return error for this function; this function can only return `EBUSY` errors.

options/posix/generic/pthread.cpp
```diff
@@ -1426,3 +1427,46 @@ int pthread_getcpuclockid(pthread_t, clockid_t *) {
...
+
+	if (!__atomic_compare_exchange_n(&__lock->lock, &expected, desired, false, __ATOMIC_ACQUIRE, __ATOMIC_ACQUIRE))
+		return EBUSY;
+	
+	return 0;
+}
+
+int pthread_spin_unlock(pthread_spinlock_t *__lock) {
+	__ensure(__atomic_load_n(&__lock->lock, __ATOMIC_RELAXED) == mlibc::this_tid());
```
no92: This assert can be removed, as unlocking a spinlock this thread doesn't hold is UB anyways.

tests/posix/pthread_spin.c
```diff
@@ -0,0 +1,39 @@
...
+    } else {
+        assert(ret == 0);
+        variable = 0;
+        pthread_spin_unlock(&lock);
+    }
+
+    return NULL;
+}
+
+int main() {
```
no92, line 25: The `EDEADLK` error return should be tested, too.

tests/posix/pthread_spin.c
```diff
@@ -0,0 +1,39 @@
...
+int variable = 0;
+
+void *worker(void*) {
+    int ret = pthread_spin_trylock(&lock);
+    if (!variable) {
+        assert(ret == EBUSY);
+    } else {
+        assert(ret == 0);
+        variable = 0;
+        pthread_spin_unlock(&lock);
```
no92: This is UB, as the lock is not held by the calling thread.

sasdallas: it is, see above trylock.

tests/posix/pthread_spin.c
sasdallas: it is

options/posix/generic/pthread.cpp
sasdallas: shouldn't there always be a check for it?

no92: Not really, and this atomic is probably just being expensive without doing anything :^)

options/posix/generic/pthread.cpp
avdgrinten: That's not true, POSIX allows additional errors.

sasdallas: Still though, EBUSY is probably a better return value so I'll side with @no92

options/posix/generic/pthread.cpp
avdgrinten: For mutexes we have seen that programs rely on glibc detecting unlocks from the wrong thread, so I'd be in favor of keeping the assertion.

options/posix/include/pthread.h
```diff
@@ -124,6 +125,12 @@ struct __mlibc_rwlockattr {
 };
 typedef struct __mlibc_rwlockattr pthread_rwlockattr_t;
 
+struct __mlibc_spinlock {
+	volatile unsigned int lock;
```
no92: The member name should start with a `__` (double underscore) to avoid polluting the namespace. Aside from that, I think the `volatile` keyword is undesireable here, as we modify this using atomics anyways.

options/posix/include/pthread.h
```diff
@@ -314,6 +321,15 @@ int pthread_rwlock_unlock(pthread_rwlock_t *__rwlock);
 
 int pthread_getcpuclockid(pthread_t __thrd, clockid_t *__clockid);
 
+/* ---------------------------------------------------------------------------- */
+/* pthread_spin functions. */
+/* ---------------------------------------------------------------------------- */
+int pthread_spin_init(pthread_spinlock_t *__lock, int pshared);
```
no92: `pshared` needs double underscores to avoid namespace pollution.

options/posix/generic/pthread.cpp
```diff
@@ -1426,3 +1427,42 @@ int pthread_getcpuclockid(pthread_t, clockid_t *) {
 			<< frg::endlog;
 	return ENOENT;
 }
+
+int pthread_spin_init(pthread_spinlock_t *__lock, int) {
```
no92, line 1431: Optionally, you can use non-double-underscore-prefixed argument names here if you like, same below. We only need to make sure to not pollute the namespace in public headers.

options/posix/generic/pthread.cpp
```diff
@@ -12,6 +12,7 @@
 #include <bits/ensure.h>
 #include <frg/allocation.hpp>
 #include <frg/array.hpp>
+#include <frg/spinlock.hpp>
```
no92, line 15: Unnecessary?

no92 (commented):
Another pass.

options/posix/generic/pthread.cpp
sasdallas: No. Provides frg::detail::loophint

PR #1563, opened 2025-12-01
`pthread_rwlock_{clock,timed}{rw,wr}lock` support

no92:
Depends on managarm/managarm#1148.

### DIFF
--- a/options/posix/generic/pthread.cpp
+++ b/options/posix/generic/pthread.cpp
@@ -20,6 +20,7 @@
 #include <mlibc/tcb.hpp>
 #include <mlibc/tid.hpp>
 #include <mlibc/threads.hpp>
+#include <mlibc/time-helpers.hpp>
 
 static bool enableTrace = false;
 
@@ -1190,7 +1191,7 @@ int pthread_barrier_wait(pthread_barrier_t *barrier) {
 // ----------------------------------------------------------------------------
 
 namespace {
-	void rwlock_m_lock(pthread_rwlock_t *rw, bool excl) {
+	int rwlock_m_lock(pthread_rwlock_t *rw, bool excl, const struct timespec *__restrict timeout = nullptr) {
 		unsigned int m_expected = __atomic_load_n(&rw->__mlibc_m, __ATOMIC_RELAXED);
 		while(true) {
 			if(m_expected) {
@@ -1205,7 +1206,9 @@ namespace {
 				}
 
 				// Wait on the futex.
-				mlibc::sys_futex_wait((int *)&rw->__mlibc_m, m_expected | mutex_waiters_bit, nullptr);
+				int e = mlibc::sys_futex_wait((int *)&rw->__mlibc_m, m_expected | mutex_waiters_bit, timeout);
+				if (e != 0 && e != EAGAIN && e != EINTR)
+					return e;
 
 				// Opportunistically try to take the lock after we wake up.
 				m_expected = 0;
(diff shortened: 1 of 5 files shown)


sysdeps/managarm/generic/fork-exec.cpp
```diff
@@ -38,13 +38,35 @@ int sys_futex_tid() {
...
+		case kHelErrNone: return 0;
+		case kHelErrTimeout: return ETIMEDOUT;
+		case kHelErrCancelled: return EINTR;
+		case kHelErrIllegalArgs: return EINVAL;
+		default: {
+			mlibc::infoLogger() << "mlibc: helFutexWait returned unexpected error "
+								<< err << frg::endlog;
+			return EINVAL;
+		}
 	}
```
avdgrinten, line 61: There is no need to duplicate the error handling here.

options/internal/include/mlibc/time-helpers.hpp
```diff
@@ -0,0 +1,51 @@
...
+#include <mlibc/ansi-sysdeps.hpp>
+#include <mlibc/debug.hpp>
+#include <time.h>
+
+namespace mlibc {
+
+// Converts the absolute time `abstime` to a relative time `reltime`.
+// Returns false if the conversion failed (e.g. due to over-/underflow).
+// If the absolute time has already passed, `reltime` is set to 0.
+bool time_relative_to_absolute(const clockid_t clock, const struct timespec *__restrict abstime, struct timespec *reltime) {
```
avdgrinten: I'm not sure if this is a good implementation strategy. It obviously doesn't work for things like process CPU time etc. However, we could keep this implementation until we come up with a better one (i.e., conditionally forward absolute time to the sysdep).

options/posix/generic/pthread.cpp
```diff
@@ -1352,6 +1357,70 @@ int pthread_rwlock_wrlock(pthread_rwlock_t *rw) {
...
+	if (rw->__mlibc_flags != 0) {
+		mlibc::panicLogger() << "mlibc: pthread_rwlock_t flags were non-zero"
+			<< frg::endlog;
+	}
+
+	// Adjust for the fact that sys_futex_wait accepts a *timeout*, but
+	// pthread_rwlock_timedwrlock accepts an *absolute time*.
+	struct timespec timeout;
+	if (!mlibc::time_relative_to_absolute(clock, abstime, &timeout))
+		return EINVAL;
```
avdgrinten, line 1376: Shouldn't we be doing an `absolute_to_relative` and not a `relative_to_absolute` conversion here?

options/posix/generic/pthread.cpp
```diff
@@ -1352,6 +1357,70 @@ int pthread_rwlock_wrlock(pthread_rwlock_t *rw) {
...
+				continue;
+		}
+
+		if (!mlibc::time_relative_to_absolute(clock, abstime, &timeout))
+			return EINVAL;
+		if (timeout.tv_sec == 0 && timeout.tv_nsec == 0)
+			return ETIMEDOUT;
+
+		// Wait on the futex.
+		int e = mlibc::sys_futex_wait((int *)&rw->__mlibc_rc, rc_expected | rc_waiters_bit, &timeout);
```
avdgrinten: Doesn't this cause us to exhaust the timeout multiple times? Once when acquiring the lock and once while waiting for readers to finish?

PR #1565, opened 2025-12-02
abis/linux: expose all XSI-permitted `sys/socket.h` identifiers

no92:
POSIX permits any identifiers beginning with `AF_`, `MSG_`, `PF_`, `SCM_`, `SHUT_`, `SO` to be part of the namespace of `<sys/socket.h>` under the XSI option.


Dennisbonke (approved):
LGTM

PR #1568, opened 2025-12-02
Enable strict ABI mismatch checking

no92:


options/internal/include/bits/nl_item.h
```diff
@@ -102,8 +102,6 @@ enum { /* NOLINT(readability-enum-initial-value) */
 	_NL_CTYPE_EXTRA_MAP_12,
 	_NL_CTYPE_EXTRA_MAP_13,
 	_NL_CTYPE_EXTRA_MAP_14,
-	_NL_CTYPE_EXTRA_MAP_15,
-	_NL_CTYPE_EXTRA_MAP_16,
 	_NL_NUM_LC_CTYPE,
```
avdgrinten, line 105: The commit message says that this is not an ABI break but this changes the value of `_NL_NUM_LC_CTYPE`.

scripts/header-abi-compare-config.yml
```diff
@@ -62,6 +62,9 @@ ignored_structs:
   - fd_set
   - semid_ds
 
+  # we introduce padding and change cmsg_len to socklen_t to match POSIX
+  - cmsghdr
```
avdgrinten, line 66: Isn't this a syscall ABI struct? Don't we have to match whatever the Linux kernel does for the Linux ABI?

no92: While Linux's struct defines it as a `size_t`, it limits the value of `cmsg_len` to net.core.optmem_max, which is not allowed to exceed INT_MAX.

options/internal/include/bits/nl_item.h
no92: `_NL_NUM_LC_CTYPE` is not public API, and this change reflects what glibc defines for it.

no92: To refer to POSIX:
> With the exception of identifiers beginning with the prefix \_POSIX\_ and those identifiers which are lexically identical to keywords defined by the ISO C standard (for example \_Bool), all identifiers that begin with an \<underscore\> and either an uppercase letter or another \<underscore\> are always reserved for any use by the implementation.

avdgrinten: Standards conformance is one thing but it's more important whether this is actually an ABI break in practice or not.

PR #1592, opened 2025-12-12
options/linux-epoll: split sys_epoll sysdeps into their own option

marv7000:
This PR seeks to split out the `epoll` sysdeps into their own option, so they can be implemented by e.g. Menix without requiring the Linux option.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -288,10 +288,10 @@ endif
 # Configuration based on enabled options.
 #----------------------------------------------------------------------------------------
 
-posix_option = get_option('posix_option').require(sysdep_supported_options.get('posix')).allowed()
-linux_option = get_option('linux_option').require(sysdep_supported_options.get('linux')).allowed()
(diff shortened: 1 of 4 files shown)


sysdeps/keyronex/meson.build
```diff
@@ -3,6 +3,7 @@ sysdep_supported_options = {
 	'linux': true,
 	'glibc': true,
 	'bsd': true,
+	'sys-epoll': true,
```
no92: The addition of `sys-epoll: true` needs to go in the first commit so that the commits are atomic.

meson.build
```diff
@@ -607,6 +607,7 @@ summary_info += {'POSIX option': get_option('posix_option')}
 summary_info += {'Linux option': get_option('linux_option')}
 summary_info += {'glibc option': get_option('glibc_option')}
 summary_info += {'BSD option': get_option('bsd_option')}
+summary_info += {'sys-epoll option': get_option('sys_epoll_option')}
```
no92: Should go into the first commit.

PR #1598, opened 2025-12-15
sysdeps/demo: fix sysdep return values

Mathewnd:
Fix errno return values and add some comments about them to the demo sysdep


avdgrinten (approved):
LGTM, thank you for fixing it.

In a future iteration, it may make sense to let the syscalls return errors.

PR #1605, opened 2025-12-19
fix: ignore flush errors when exiting

Valeryum999:
When a process closes stdout and stderr before exiting (such as all the coreutils), mlibc complains that it cannot flush stdout stderr.
This pr adds a variable "isExiting" that gets set as true when exiting, and removes the error logs when trying to flush stdout and stderr

### DIFF
--- a/options/internal/generic/exiting.cpp
+++ b/options/internal/generic/exiting.cpp
@@ -1,0 +1,5 @@
+#include <mlibc/exit.hpp>
+
+namespace mlibc {
+std::atomic<bool> processIsExiting = false;
+} // namespace mlibc
--- a/meson.build
+++ b/meson.build
@@ -405,6 +405,7 @@ internal_sources = [
 	'options/internal/generic/debug.cpp',
 	'options/internal/generic/ensure.cpp',
 	'options/internal/generic/essential.cpp',
+	'options/internal/generic/exiting.cpp',
 	'options/internal/generic/frigg.cpp',
 	'options/internal/generic/getopt.cpp',
 	'options/internal/generic/global-config.cpp',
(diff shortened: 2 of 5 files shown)


options/internal/generic/exiting.cpp
```diff
@@ -0,0 +1,5 @@
+#include <mlibc/exit.hpp>
+
+namespace mlibc{
+    bool isExiting = false;
```
no92: Given that multiple threads may access this and regular `bool` makes (iirc) no memory ordering guarantees, this should be atomic.

no92: I'm torn on this, but maybe naming it `processIsExiting` or similar would help with the thread vs. process distinction?

Valeryum999: Done

options/internal/generic/exiting.cpp
```diff
@@ -0,0 +1,5 @@
+#include <mlibc/exit.hpp>
+
+namespace mlibc{
+    std::atomic<bool> processIsExiting = false;
```
no92: You might want to run clang-format over your PR, as the `.hpp` files use 4 spaces for indentation. Also not that we don't indent in namespaces, and place a comment at the closing brace: `} // namespace mlibc`

avdgrinten: When running clang-format only reformat the diff though (and not the entire files since they are not all in clang-format style).

Valeryum999: Oh right, sorry, fixing right now

Valeryum999: Allegedly done

no92:
It would be nice to squash the commits into one with a commit messages in mlibc style; something like `options/ansi: ignore flush errors when exiting` would work.

PR #1612, opened 2025-12-24
options/posix: implement `posix_getdents`

no92:


options/posix/generic/dirent.cpp
```diff
@@ -85,6 +85,21 @@ struct dirent *readdir(DIR *dir) {
 	return &dir->__current;
 }
 
+ssize_t posix_getdents(int fildes, void *buf, size_t nbyte, int flags) {
+	if (flags) {
+		errno = EINVAL;
+		return -1;
+	}
```
avdgrinten, line 92: This should be checked by the sysdeps not by the wrapper.

avdgrinten: Discussed offline: that would require a separate sysdep for `posix_getdents`.

PR #1621, opened 2025-12-28
[ABI BREAK] abis/linux: make sa_flags POSIX-compliant

no92:
This does not change the struct size or alignment, but on 64-bit architectures the member is smaller and this implicitly adds padding.


Dennisbonke (approved):
LGTM

PR #1623, opened 2025-12-28
`stdio.h` and `wchar.h` improvements

no92:


options/ansi/include/mlibc/file-io.hpp
```diff
@@ -3,6 +3,7 @@
 
 #include <stdio.h>
 
+#include <mlibc/file.hpp>
```
avdgrinten: I'm not convinced that this is a good idea. We basically gain nothing here and we lose the ability to use inline functions that access struct file internals (which will be quite a lot faster than PLT calls).

PR #1630, opened 2025-12-30
Debug environment variables

Mathewnd:
As of now, the only logging option controlled by an environment is malloc's. This PR switches over some already existing constexpr enables into a proper environment variable.

Solves #614

### DIFF
--- a/options/internal/generic/global-config.cpp
+++ b/options/internal/generic/global-config.cpp
@@ -26,6 +26,7 @@ GlobalConfig::GlobalConfig() {
 	debugLocale = envEnabled("MLIBC_DEBUG_LOCALE");
 	debugPthreadTrace = envEnabled("MLIBC_DEBUG_PTHREAD_TRACE");
 	debugPathResolution = envEnabled("MLIBC_DEBUG_PATH_RESOLUTION");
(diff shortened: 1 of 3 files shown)


options/rtld/generic/main.cpp
```diff
@@ -387,6 +388,19 @@ extern "C" void *interpreterMain(uintptr_t *entry_stack) {
 	const char *env_ld_library_path = nullptr;
 	const char *env_ld_preload = nullptr;
 
+	auto match_config = [](frg::string_view v) {
+		static const frg::tuple<const frg::string_view, bool *> config_mapping[] = {
```
no92: I don't think the `static` is needed here, and I think a reference (`bool &`) should work instead of a pointer here?

For better visibility, moving this table up in the file and putting it in an anonymous namespace would be better IMO.

no92 (approved):
One code organization suggestion, other than that LGTM provided it works as intended.

options/rtld/generic/main.cpp
Mathewnd: Moving it to an anonymous namespace does not really work, as it causes a constructor to be added to rtld which then causes an assert to trip ``ldso: Found unexpected own global constructor(s), init_array starts at: 0x7fa7009d5eb0``
Other two are fixed now

PR #1658, opened 2026-01-11
[ABI BREAK] abis: Fix definition of fd_set

Dennisbonke:


abis/linux/fd_set.h
```diff
@@ -5,8 +5,13 @@
 
 #define FD_SETSIZE 1024
 
-typedef struct __attribute__((__aligned__(__alignof__(long)))) {
-	__mlibc_uint8 fds_bits[128];
+typedef long int __fd_mask;
```
no92, line 10: While I get that the linux kernel has this as `long`, making this signed breaks our implementations of the `FD_*` macros.

abis/linux/fd_set.h
```diff
@@ -5,8 +5,13 @@
...
+#define __NFDBITS (8 * (int) sizeof (__fd_mask))
+
+typedef struct {
+	/* libtirpc reaches into fds_bits directly, so we must keep the name and type exactly as expected.
+	 * This now matches the definition in sys/select.h from glibc. */
+	__fd_mask fds_bits[FD_SETSIZE / __NFDBITS];
```
no92, line 17: We might want to have the name of this be switched between this and a double underscore-prefixed variant based on the appropriate feature test macros, as earlier POSIX specs required this name, while current ones don't.

no92 (changes requested):

abis/linux/fd_set.h
no92:
```suggestion
#if __MLIBC_XOPEN
	__fd_mask fds_bits[FD_SETSIZE / __NFDBITS];
#else
	__fd_mask __fds_bits[FD_SETSIZE / __NFDBITS];
#endif
```

abis/linux/fd_set.h
```diff
@@ -5,8 +5,13 @@
 
 #define FD_SETSIZE 1024
 
-typedef struct __attribute__((__aligned__(__alignof__(long)))) {
-	__mlibc_uint8 fds_bits[128];
+typedef long int __fd_mask;
+#define __NFDBITS (8 * (int) sizeof (__fd_mask))
```
no92, line 11: I would suggest keeping the `__NFDBITS` macro in the `options/posix/include/sys/select.h` header, as I see no point for anyone to diverge.

abis/ironclad/fd_set.h
```diff
@@ -5,8 +5,11 @@
 
 #define FD_SETSIZE 1024
 
+typedef __mlibc_uint8 __fd_mask;
+#define __NFDBITS (8 * (int) sizeof (__fd_mask))
```
no92, line 9: See below.

options/posix/generic/sys-select.cpp
```diff
@@ -9,17 +11,21 @@
...
 #include <mlibc/posix-sysdeps.hpp>
 
+using fd_mask_unsigned = std::make_unsigned_t<__fd_mask>;
+
+#define FD_MASK(n) (fd_mask_unsigned)(fd_mask_unsigned{1} << (n))
+
 void __FD_CLR(int fd, fd_set *set) {
 	__ensure(fd < FD_SETSIZE);
-	set->fds_bits[fd / 8] &= ~(1 << (fd % 8));
+	set->fds_bits[fd / NFDBITS] &= std::bit_cast<fd_mask>(~FD_MASK(fd % NFDBITS));
```
no92: The modulo could be replaced by `fd & (NFDBITS-1)`. We should then also add a `static_assert(std::has_single_bit(NFDBITS));` to make sure that `NFDBITS` is a power of two.

options/posix/generic/sys-select.cpp
```diff
@@ -9,17 +11,21 @@
...
 }
 int __FD_ISSET(int fd, fd_set *set) {
 	__ensure(fd < FD_SETSIZE);
-	return set->fds_bits[fd / 8] & (1 << (fd % 8));
+	return (std::bit_cast<fd_mask_unsigned>(set->fds_bits[fd / NFDBITS]) & FD_MASK(fd % NFDBITS)) != 0;
```
no92: Modulo.

options/posix/generic/sys-select.cpp
```diff
@@ -9,17 +11,21 @@
...
 }
 void __FD_SET(int fd, fd_set *set) {
 	__ensure(fd < FD_SETSIZE);
-	set->fds_bits[fd / 8] |= 1 << (fd % 8);
+	set->fds_bits[fd / NFDBITS] = std::bit_cast<fd_mask>(std::bit_cast<fd_mask_unsigned>(set->fds_bits[fd / NFDBITS]) | FD_MASK(fd % NFDBITS));
```
no92: Modulo.

abis/linux/fd_set.h
no92: You will also need to `#include <mlibc-config.h>` at the top.

PR #1678, opened 2026-02-16
[ABI BREAK] pthread structs future-proofing

no92:
This would allow for future expansion of the structs to accomodate possible additions, e.g. in POSIX, without breaking backwards compatability. Adding a new field would bump the revision of the affected struct. The new fields would then only be accessed if the revision matches.


Dennisbonke (approved):
LGTM

options/internal/include/bits/threads.h
```diff
@@ -1,6 +1,10 @@
 #ifndef _MLIBC_INTERNAL_THREADS_H
 #define _MLIBC_INTERNAL_THREADS_H
 
+#ifdef __cplusplus
+extern "C" {
+#endif
```
avdgrinten: This needs to go below the `#include`s.

options/internal/include/mlibc/sysdep-signatures.hpp
```diff
@@ -95,6 +95,8 @@ SYSDEP_FUNC(VmReadahead, void *pointer, size_t size);
 #endif /* MLIBC_BUILDING_RTLD */
 
 #if __MLIBC_POSIX_OPTION
+#include <sched.h>
```
avdgrinten: In all other cases we include the abi-bit directly (at the top of the file).

options/internal/generic/threads.cpp
```diff
@@ -442,13 +409,10 @@ int thread_mutexattr_settype(struct __mlibc_mutexattr *attr, int type) {
...
-	auto pshared = attr ? attr->__mlibc_pshared : __MLIBC_THREAD_PROCESS_PRIVATE;
-
-	cond->__mlibc_clock = clock;
-	cond->__mlibc_flags = pshared;
-
-	__atomic_store_n(&cond->__mlibc_seq, 1, __ATOMIC_RELAXED);
+	if (attr) {
+		cond->__mlibc_clock = attr->__mlibc_clock;
+		cond->__mlibc_flags = attr->__mlibc_pshared;
+	}
```
avdgrinten, line 452: Why is this dropped? This seems to be a correctness bug, the `cond` variable will contain garbage.

Also, we should probably use placement new for this struct (and others) since we're now relying on `std::atomic`.

options/internal/include/bits/pthread_types.h
```diff
@@ -0,0 +1,63 @@
...
+	void *__heap_ptr;
+} pthread_barrier_t;
+
+typedef struct {
+	void *__heap_ptr;
+} pthread_barrierattr_t;
+
+typedef struct {
+	char __opaque[__MLIBC_THREAD_COND_SIZE];
+} __attribute__((aligned(__INTPTR_WIDTH__ / 4))) pthread_cond_t;
```
avdgrinten: Maybe document (for this + the other structs) that we're aligning to double word alignment since the / 4 looks somewhat odd.

options/posix/generic/pthread.cpp
```diff
@@ -52,140 +53,162 @@ struct ScopeTrace {
...
+int pthread_attr_destroy(pthread_attr_t *attr) {
+	if (attr && attr->__heap_ptr)
+		frg::destruct<__mlibc_threadattr>(getAllocator(), static_cast<__mlibc_threadattr *>(attr->__heap_ptr));
+
 	return 0;
 }
 
 int pthread_attr_getdetachstate(const pthread_attr_t *attr, int *detachstate) {
-	*detachstate = attr->__mlibc_detachstate;
+	auto a = static_cast<__mlibc_threadattr *>(attr->__heap_ptr);
```
avdgrinten: Why doesn't this use the `from()` helper (same for the static_casts below)?

options/internal/include/bits/pthread_types.h
```diff
@@ -0,0 +1,63 @@
+#ifndef _MLIBC_INTERNAL_PTHREAD_TYPTES_H
+#define _MLIBC_INTERNAL_PTHREAD_TYPTES_H
```
avdgrinten: No correctness issue but this has a typo.

options/posix/generic/pthread.cpp
```diff
@@ -292,35 +318,38 @@ int pthread_attr_setsigmask_np(pthread_attr_t *__restrict attr,
...
+	sigdelset(&a->__sigmask, SIGCANCEL);
 
 	return 0;
 }
 
 int pthread_getattr_np(pthread_t thread, pthread_attr_t *attr) {
 	auto tcb = reinterpret_cast<Tcb*>(thread);
-	*attr = pthread_attr_t{};
+
+	pthread_attr_init(attr);
```
avdgrinten, line 348: Why does `pthread_getattr_np` initialize the attribute?

PR #1729, opened 2026-03-27
POSIX AIO for linux

no92:


options/internal/include/mlibc/sysdep-tags.hpp
```diff
@@ -538,6 +539,18 @@ struct MqReceive {};
...
+// [[gnu::weak]] int sys_aio_write(struct aiocb *cb);
+struct AioWrite {};
+// [[gnu::weak]] int sys_aio_fsync(int op, struct aiocb *cb);
+struct AioFsync {};
+// [[gnu::weak]] int sys_aio_listio(int mode, struct aiocb *__restrict const list[], int n, struct sigevent *__restrict sevp);
+struct AioListio {};
+// [[gnu::weak]] int sys_aio_suspend(const struct aiocb *const list[], int n, const struct timespec *ts);
+struct AioSuspend {};
+// [[gnu::weak]] int sys_aio_cancel(int fildes, struct aiocb *cb, int *result);
+struct AioCancel {};
```
avdgrinten: Can we drop these obsolete `[[gnu::weak]]` comments from the entire file (in a separate commit but in the same PR)?

options/posix/generic/aio.cpp
```diff
@@ -0,0 +1,101 @@
...
+#include <bits/ensure.h>
+#include <errno.h>
+#include <limits.h>
+
+#include <mlibc/aio.hpp>
+#include <mlibc/all-sysdeps.hpp>
+
+int aio_read(struct aiocb *cb) {
+	if (cb->aio_nbytes > SSIZE_MAX)
+		return EINVAL;
```
avdgrinten: mlibc should not check kernel preconditions.

options/posix/generic/aio.cpp
```diff
@@ -0,0 +1,101 @@
...
+	if (int e = mlibc::sysdep_or_enosys<AioRead>(cb); e) {
+		errno = e;
+		return -1;
+	}
+	return 0;
+}
+
+int aio_write(struct aiocb *cb) {
+	if (cb->aio_nbytes > SSIZE_MAX)
+		return EINVAL;
```
avdgrinten: mlibc should not check kernel preconditions.

options/posix/generic/aio.cpp
```diff
@@ -0,0 +1,101 @@
...
+	if (state != mlibc::AioOpState::success && state != mlibc::AioOpState::cancelled) {
+		errno = EINVAL;
+		return -1;
+	}
+
+	// invalidate the state so that aio_return can only be called once
+	cbv.setState(mlibc::AioOpState::invalid);
+	ssize_t res = cb->__res;
+	cbv.resetOperationResult();
+	return res;
```
avdgrinten: Shouldn't this be a sysdep?

options/posix/generic/aio.cpp
```diff
@@ -0,0 +1,101 @@
...
+			return cb->__err;
+		case mlibc::AioOpState::inProgress:
+			return EINPROGRESS;
+		case mlibc::AioOpState::cancelled:
+			return ECANCELED;
+		case mlibc::AioOpState::invalid:
+			errno = EINVAL;
+			return -1;
+	}
+	__builtin_unreachable();
```
avdgrinten: Shouldn't this be a sysdep?

options/posix/include/mlibc/aio.hpp
```diff
@@ -0,0 +1,76 @@
...
+#include <utility>
+
+namespace mlibc {
+
+enum class AioOpState : int {
+	invalid,
+	success,
+	inProgress,
+	cancelled,
+};
```
avdgrinten: mlibc uses `snake_case` both for type and for enum variants.

options/posix/include/mlibc/aio.hpp
```diff
@@ -0,0 +1,76 @@
...
+
+template <typename T>
+struct BasicAioCbView {
+	T *cb_;
+
+	explicit constexpr BasicAioCbView(T *cb) noexcept : cb_{cb} {}
```
avdgrinten: mlibc also uses `snake_case` for functions.

options/posix/include/mlibc/aio.hpp
```diff
@@ -0,0 +1,76 @@
...
+		__atomic_store_n(&cb_->__state, std::to_underlying(state), __ATOMIC_RELEASE);
+		if (wake)
+			sysdep<FutexWake>(&cb_->__state, true);
+	}
+
+	inline bool
+	exchangeState(AioOpState state, AioOpState expected, bool wake = true) const noexcept {
+		if (__atomic_compare_exchange_n(
+		        &cb_->__state,
+		        (int *) &expected,
```
avdgrinten: Casting the pointer and then accessing it is UB.

avdgrinten (commented):
This is a review of the `options/` part.

I still need to review the io_uring-based implementation.

options/posix/generic/aio.cpp
no92: This is not a kernel precondition, but explicitly mandatory in POSIX.

options/posix/generic/aio.cpp
no92: I don't see any point in this (or `aio_error`) being a sysdep - these should exclusively operate on the `struct aiocb` members that track the state (`__state`, which is atomically updated by the worker thread to reflect the AIO request's live state) and result/error (`__res`/`__err`), which are only valid when the state signals completion.

Dennisbonke (approved):
LGTM

options/posix/generic/aio.cpp
```diff
@@ -0,0 +1,104 @@
...
+		return -1;
+	}
+	return 0;
+}
+
+int lio_listio(
+    int mode, struct aiocb *__restrict const list[], int n, struct sigevent *__restrict sevp
+) {
+	for (int i = 0; i < n; i++) {
+		if (!list[i])
```
Geertiebear: Why do we do the checking here and not in other places?

sysdeps/linux/generic/posix/aio.cpp
```diff
@@ -0,0 +1,951 @@
...
+	close(aioWorker.ringFd);
+	close(aioWorker.epollFd);
+	close(aioWorker.ring.fd);
+	__atomic_store_n(&aioWorker.initialized, false, __ATOMIC_RELEASE);
+}
+
+} // namespace
+
+void setupObserverThread() {
+	if (__atomic_load_n(&aioWorker.initialized, __ATOMIC_ACQUIRE) == false) {
```
Geertiebear, line 562: Can't this be a `pthread_once`?

no92: We can't use pthread_once here, as it cannot be reset back. The reset is necessary for correct atfork semantics, which are implemented in the atfork handler.

PR #1758, opened 2026-04-10
Sink signature rework

no92:
This PR adds a `append(Char *str, size_t input_max, size_t ouput_max)` function to all sinks. This aims to allow to limit both the number of characters to input (e.g. for supplying non-zero terminated strings) and to limit the number of characters to output (e.g. for implementing precision for `%ls` in printf or %s in wprintf). This also neatly translates to the `mbsnrtowcs`/`wcsnrtombs` functions, which we use to do wide <=> mb conversion.

### DIFF
--- a/options/ansi/generic/stdio.cpp
+++ b/options/ansi/generic/stdio.cpp
@@ -359,6 +359,15 @@ int fputws_unlocked(const wchar_t *__restrict ws, mlibc::abstract_file *f) {
 	return 1;
 }
 
+template <typename From, typename To = std::conditional_t<std::is_same_v<From, char>, wchar_t, char>>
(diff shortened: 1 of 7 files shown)


options/ansi/generic/stdio.cpp
```diff
@@ -725,6 +743,9 @@ struct ResizePrinter {
 	size_t count;
 };
 
+static_assert(frg::SinkFor<ResizePrinter>);
+static_assert(frg::SinkFor<ResizePrinter>);
```
Mathewnd, line 754: Duplicate static_assert?

Mathewnd (approved):

PR #1766, opened 2026-04-21
options/internal: initialize `GlobalConfig` while avoiding cxa_guard

no92:
LLVM's `libc++` uses pthread mutexes for the `__cxa_guard_{acquire,release}` functions. As `globalConfig` was a static variable, this used the guard, which in turn was used by our pthread mutex implementation. This resulted in unbounded recursion.

### DIFF
--- a/options/internal/generic/global-config.cpp
+++ b/options/internal/generic/global-config.cpp
@@ -1,8 +1,14 @@
 #include <stdlib.h>
-#include <string.h>
+#include <mlibc/allocator.hpp>
 #include <mlibc/global-config.hpp>
 
 namespace mlibc {
+
+namespace {
+
+constinit mlibc::lazy_eternal<GlobalConfig> globalConfigInstance;
+
+} // namespace
 
 struct GlobalConfigGuard {
 	GlobalConfigGuard();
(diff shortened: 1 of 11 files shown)


options/internal/generic/global-config.cpp
```diff
@@ -1,9 +1,15 @@
 #include <stdlib.h>
-#include <string.h>
+#include <mlibc/allocator.hpp>
 #include <mlibc/global-config.hpp>
 
 namespace mlibc {
 
+namespace {
+
+mlibc::lazy_eternal<GlobalConfig> globalConfigInstance;
```
avdgrinten: `constinit`.

options/internal/include/mlibc/allocator.hpp
```diff
@@ -35,4 +36,32 @@ MemoryAllocator &getAllocator();
...
+			}
+		}
+
+		return *self.box_.get();
+	}
+
+private:
+	mutable uint32_t initialized_ = 0;
+	mutable FutexLock lock_;
+	mutable frg::manual_box<T> box_;
```
avdgrinten, line 64: Why `mutable`?

avdgrinten (commented):
We should also use `lazy_eternal` for other singletons.

options/internal/include/mlibc/allocator.hpp
no92: This is useful for const-correctness of the struct when using it with `deducing this`.

Mathewnd (approved):

PR #1768, opened 2026-04-23
String collation

no92:
This implements the `strcoll_l` family of collation functions. This implementation uses the same collation tables as glibc, and aims to output the same values as glibc.

Depends on managarm/frigg#128.


no92:
The ABI break is a false positive.

ElectrodeYT (approved):
LGTM

PR #1775, opened 2026-05-05
GCC 16 support

no92:


options/rtld/aarch64/runtime.S
```diff
@@ -13,15 +13,106 @@ __mlibcTlsdescStatic:
...
+	ret
+
+.Lslowpath:
+	ldp x1, x2, [sp], #16
+
+	// Preserve all registers except x0.
+	stp x29, x30, [sp, #-16]!
+	mov x29, sp
+
+	// Save all integer registers x1-x28 (x29, x30 already saved)
```
avdgrinten: This is overly conservative: we only need to save the caller saved registers.

options/rtld/x86_64/runtime.S
```diff
@@ -9,23 +9,92 @@ __mlibcTlsdescStatic:
...
+
+	# Preserve all registers except rax and flags
+	push %rcx
+	push %rdx
+	push %rsi
+	push %rdi
+	push %r8
+	push %r9
+	push %r10
+	push %r11
```
avdgrinten: This is overly conservative (see Aarch64).

options/rtld/x86_64/runtime.S
```diff
@@ -9,23 +9,92 @@ __mlibcTlsdescStatic:
...
+	sub $256, %rsp
+	movdqu %xmm0, 0(%rsp)
```
avdgrinten: Technically we'd also need to save the extended state (AVX etc.), right? That seems like a major annoyance since we'd need to determine if AVX is available and enabled first.

avdgrinten (commented):
FWIW musl solves TLSDESC by updating the DTVs of all threads when they need to be extended.

options/rtld/aarch64/runtime.S
no92: The TLSDESC ABI documents explicitly specify that any register modified by these resolver functions must be preserved.

no92: The new commits avoid this by forcing the slow path resolver function to only use general registers.

options/internal/include/mlibc/tcb.hpp
```diff
@@ -161,6 +161,8 @@ struct Tcb {
 static_assert(offsetof(Tcb, stackCanary) == 0x28);
 // sysdeps/linux/x86_64/cp_syscall.S uses the offset of cancelBits.
 static_assert(offsetof(Tcb, cancelBits) == 0x30);
+// options/linker/x86_64/runtime.S uses the offset of dtvSize.
```
qookei: That's not true anymore, right?

options/internal/include/mlibc/tcb.hpp
```diff
@@ -173,6 +175,8 @@ static_assert(offsetof(Tcb, stackCanary) == 0x14);
 static_assert(offsetof(Tcb, cancelBits) == 0x18);
 #elif defined(__aarch64__)
 // The thread pointer on AArch64 points to 16 bytes before the end of the TCB.
+// options/linker/aarch64/runtime.S uses the offset of dtvSize.
```
qookei: Likewise.

options/rtld/generic/linker.cpp
```diff
@@ -1437,64 +1437,48 @@ Tcb *allocateTcb() {
...
-			mlibc::infoLogger() << "rtld: accessDtv wrote tls image at " << buffer
-					<< ", size = 0x" << frg::hex_fmt{object->tlsSegmentSize} << frg::endlog;
-		}
-	}
-
-	return (void *)((char *)tcb_ptr->dtvPointers[object->tlsIndex] + TLS_DTV_OFFSET);
+	size_t size = __atomic_load_n(&tcb_ptr->dtvSize, __ATOMIC_ACQUIRE);
+	__ensure(object->tlsIndex < size);
+	void **pointers = __atomic_load_n(&tcb_ptr->dtvPointers, __ATOMIC_ACQUIRE);
+	void *ptr = __atomic_load_n(&pointers[object->tlsIndex], __ATOMIC_ACQUIRE);
```
qookei: Why an atomic load here? The corresponding stores are not atomic, and I'm pretty sure that since dtvPointers is swapped in after inserting a new element, this can just be a regular load anyway.

options/rtld/generic/linker.cpp
```diff
@@ -1437,64 +1437,48 @@ Tcb *allocateTcb() {
...
 void *tryAccessDtv(SharedObject *object) {
 	Tcb *tcb_ptr = mlibc::get_current_tcb();
 
-	if (object->tlsIndex >= tcb_ptr->dtvSize)
+	size_t size = __atomic_load_n(&tcb_ptr->dtvSize, __ATOMIC_ACQUIRE);
+	if (object->tlsIndex >= size)
 		return nullptr;
-	if (!tcb_ptr->dtvPointers[object->tlsIndex])
+	void **pointers = __atomic_load_n(&tcb_ptr->dtvPointers, __ATOMIC_ACQUIRE);
+	void *ptr = __atomic_load_n(&pointers[object->tlsIndex], __ATOMIC_ACQUIRE);
```
qookei: Likewise.

options/rtld/generic/linker.cpp
```diff
@@ -2092,6 +2129,57 @@ void Loader::_processRelocations(Relocation &rel) {
...
+			if (!target) {
+				((uint64_t *)rel.destination())[0] = reinterpret_cast<uintptr_t>(&__mlibcTlsdescStatic);
+				((uint64_t *)rel.destination())[1] = 0;
+			} else if (target->tlsModel == TlsModel::initial) {
+				((uint64_t *)rel.destination())[0] = reinterpret_cast<uintptr_t>(&__mlibcTlsdescStatic);
+				uint64_t value = symValue + target->tlsOffset + tlsOffsetFromTp + rel.addend_norel();
+				((uint64_t *)rel.destination())[1] = value;
+			} else {
+				// Access DTV for object to force the entry to be allocated and initialized
+				accessDtv(target);
```
qookei: This is unnecessary since the DTV is resized eagerly. (also likewise in the other function handling TLSDESC relocs)

options/rtld/generic/linker.cpp
```diff
@@ -1986,6 +2011,18 @@ void Loader::_scheduleInit(SharedObject *object) {
...
 
+// TODO: TLSDESC relocations aren't aarch64/x86_64 specific
+#if defined(__aarch64__) || defined(__x86_64__)
+extern "C" void *__mlibcTlsdescStatic(void *);
+extern "C" void *__mlibcTlsdescDynamic(void *);
+
+struct TlsdescData {
+	uintptr_t tlsIndex;
+	uintptr_t addend;
+	SharedObject *object;
```
qookei: The object is not used anymore, right?

options/rtld/generic/linker.cpp
```diff
@@ -1937,6 +1921,47 @@ void Loader::_buildTlsMaps() {
...
+					}
+					ndtv[i] = tls_ptr;
+
+					// Initialize the TLS segment for this thread as well.
+					memset(tls_ptr, 0, object->tlsSegmentSize);
+					memcpy(tls_ptr, object->tlsImagePtr, object->tlsImageSize);
+				}
+			}
+
+			__atomic_store_n(&tcb->dtvPointers, ndtv, __ATOMIC_RELEASE);
```
qookei, line 1964: 1. Is a race between concurrent `_buildTlsMaps` calls possible, or is there locking in the caller?
2. This leaks the old DTV. We can't just free it here though... hm, maybe add a comment that this is intentional?

qookei (commented):
In general LGTM, but I'm feeling very nitpicky today :^)

options/rtld/generic/linker.cpp
no92: `_buildTlsMaps` takes the `runtimeTlsMapLock` at the top.

qookei: Ah my bad.

qookei (approved):

marv7000 (approved):

PR #1776, opened 2026-05-05
options/glibc: add support for `strto*_l` functions

no92:
Supersedes #1378, as these were the only functions from that PR not yet implemented. This should mean that building libc++ with locale support should work now (untested).


ElectrodeYT (approved):
LGTM

PR #1777, opened 2026-05-08
Various fixes, part 4

no92:
Most of these issues were pointed out by mint's Claude audit of mlibc.

### DIFF
--- a/options/ansi/include/inttypes.h
+++ b/options/ansi/include/inttypes.h
@@ -1,5 +1,5 @@
-#ifndef _STDINT_H
-#define _STDINT_H
+#ifndef _INTTYPES_H
+#define _INTTYPES_H
 
 #include <stdint.h>
 #include <bits/wchar_t.h>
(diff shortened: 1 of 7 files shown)


options/rtld/generic/main.cpp
```diff
@@ -480,9 +480,11 @@ extern "C" void *interpreterMain(uintptr_t *entry_stack) {
 		auto value = const_cast<char *>(view.data() + s + 1);
 
 		if(name == "LD_LIBRARY_PATH" && *value) {
-			env_ld_library_path = value;
+			if(!rtldConfig.secureRequired)
```
Mathewnd: Is this really nescessary? This gets checked already when actually using the env variables. Not only that, ``rtldConfig.secureRequired`` isn't even initialized yet.

Mathewnd (approved):
LGTM other than the ``rtldConfig.secureRequired`` stuff

PR #1787, opened 2026-05-25
options/ansi: fix %x sscanf with 0

Mathewnd:


marv7000 (approved):

tests/ansi/sscanf.c
```diff
@@ -394,6 +394,8 @@ int main() {
 	assert(int_value == 0x1234);
 	assert(sscanf(" -0x1234", "%x", &int_value) == 1);
 	assert(int_value == -0x1234);
+	assert(sscanf("0", "%x", &int_value) == 1);
```
no92, line 397: Can we also add checks for numeric strings that start with zero? E.g. `01337` or some invalid (?) ones like `00xc0ffee`?

PR #1788, opened 2026-05-26
Rtld promotion fix

Mathewnd:
Fixes a regression on shared object promotion, which breaks it on absolute paths. Also changes the tests to use an absolute path for libtwo.

### DIFF
--- a/options/rtld/generic/main.cpp
+++ b/options/rtld/generic/main.cpp
@@ -729,7 +729,12 @@ void *__dlapi_open(const char *file, int flags, void *returnAddress) {
 	// TODO: Thread-safety!
 	auto rts = rtsCounter++;
 
-	SharedObject *object = initialRepository->findLoadedObject(file);
+	auto objectName = frg::string_view{file};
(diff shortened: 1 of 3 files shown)


marv7000 (approved):

options/rtld/generic/main.cpp
```diff
@@ -729,7 +729,12 @@ void *__dlapi_open(const char *file, int flags, void *returnAddress) {
 	// TODO: Thread-safety!
 	auto rts = rtsCounter++;
 
-	SharedObject *object = initialRepository->findLoadedObject(file);
+	auto objectName = frg::string_view{file};
+	auto lastSlash = objectName.find_last('/');
+	if (lastSlash != static_cast<size_t>(-1))
+		objectName = objectName.sub_string(lastSlash + 1, objectName.size() - (lastSlash + 1));
```
avdgrinten, line 735: This will not work if multiple symlinks point to the same object. Maybe a better way to handle this is to store the SONAME instead of the file name? Since that's probably a larger change, we could open an issue and solve it separately.

PR #1793, opened 2026-06-02
options/{ansi,posix}: Assortment of fixes for XFCE

marv7000:


options/posix/include/dirent.h
```diff
@@ -46,6 +46,7 @@ struct posix_dent {
 
 #if defined(_DEFAULT_SOURCE)
 #define IFTODT(mode) (((mode) & 0170000) >> 12)
+#define DTTOIF(dirtype) ((dirtype) << 12)
```
no92, line 49: `DTTOIF` needs to be added to `scripts/rust-libc-config.yml`

options/posix/include/sys/types.h
```diff
@@ -37,6 +37,11 @@ typedef char *__caddr_t;
 #include <endian.h>
 #include <stdint.h>
 #include <sys/select.h>
+#include <sys/sysmacros.h>
```
no92, line 40: Is the `sys/sysmacros.h` include necessary? Shouldn't we include `abi-bits/dev_t.h`?

options/ansi/generic/stdio.cpp
```diff
@@ -1209,24 +1209,32 @@ int do_scanf(H &handler, const Char *fmt, __builtin_va_list args) {
 				unsigned long long res = 0;
 				char c = handler.look_ahead();
 				int count = 0;
+				// Number of input characters consumed by this conversion;
```
no92, line 1212: We should have a test for this.

options/ansi/generic/file-io.cpp
```diff
@@ -327,6 +327,9 @@ int abstract_file::seek(off_t offset, int whence) {
 	// TODO: If the seek is "small", we can just modify our internal offset.
 	purge();
 
+	// A successful seek clears the end-of-file indicator (C99 7.21.9.2).
+	__status_bits &= ~__MLIBC_EOF_BIT;
```
no92, line 331: We should have a test for this.

no92 (commented):
Would be nice to split these changes up in 3-4 commits.

PR #1794, opened 2026-06-05
Fixes for Firefox

marv7000:
The changes to sys/socket.h are breaking

### DIFF
--- a/options/posix/generic/posix_stdlib.cpp
+++ b/options/posix/generic/posix_stdlib.cpp
@@ -575,8 +575,14 @@ int getsubopt(char **__restrict__ optionp, char *const *__restrict__ keylistp, c
 char *secure_getenv(const char *name) {
 	if (mlibc::rtldConfig().secureRequired)
 		return nullptr;
-	else
-		return getenv(name);
+	if (!environ || !name[0])
+		return nullptr;
+	size_t len = strlen(name);
+	for (char **ep = environ; *ep; ++ep) {
+		if (!strncmp(*ep, name, len) && (*ep)[len] == '=')
+			return *ep + len + 1;
+	}
+	return nullptr;
 }
 
 void *reallocarray(void *ptr, size_t m, size_t n) {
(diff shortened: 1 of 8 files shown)


options/posix/include/sys/socket.h
```diff
@@ -56,7 +56,7 @@ struct sockaddr {
 
 /* For parsing control messages only. */
 /* Returns a pointer to the first header or nullptr if there is none. */
-#define CMSG_FIRSTHDR(m) ((size_t)(m)->msg_controllen <= sizeof(struct cmsghdr) \
```
no92, line 59: This change seems to diverge from musl and glibc behavior; why is this needed and correct?

options/rtld/generic/linker.cpp
```diff
@@ -908,10 +908,8 @@ void ObjectRepository::_parseDynamic(SharedObject *object) {
 		case DT_TLSDESC_PLT: case DT_TLSDESC_GOT:
 			break;
 		default:
-			// Ignore unknown entries in the os-specific area as we don't use them.
```
no92, line 953: What sort of entries does Firefox have? Do we maybe need to handle them?

options/posix/generic/pthread.cpp
```diff
@@ -58,7 +58,7 @@ static constexpr unsigned int mutex_excl_bit = static_cast<uint32_t>(1) << 30;
 static constexpr unsigned int rc_count_mask = (static_cast<uint32_t>(1) << 31) - 1;
 static constexpr unsigned int rc_waiters_bit = static_cast<uint32_t>(1) << 31;
 
-static constexpr size_t default_stacksize = 0x200000;
```
no92, line 65: How does this interact with the managarm sysdeps change?

options/rtld/generic/linker.cpp
marv7000: We can't handle them. They're non-standard (0x80000000 + something) and this mirrors glibc behavior.

options/posix/generic/posix_stdlib.cpp
```diff
@@ -575,8 +575,14 @@ int getsubopt(char **__restrict__ optionp, char *const *__restrict__ keylistp, c
 char *secure_getenv(const char *name) {
 	if (mlibc::rtldConfig().secureRequired)
 		return nullptr;
-	else
```
no92, line 578: Why is this change needed? The implementation also looks fragile.

marv7000: Firefox does this https://github.com/mozilla-firefox/firefox/commit/18afb03c175f2d43cfe8f8e76a9f4f4f5afc536b, and without the change it seems to stack overflow. But yes, this is probably not the right fix.

PR #1795, opened 2026-06-09
options/posix: Fix some bugs in read_dns_name and a leak in getservby{name,port}

Qwinci:
- Change the dns pointer to be of the proper type and cast the arguments to prevent unwanted sign extension.
- Change the dns response buffer size to 512 bytes which is the recommended minimum to prevent truncated/split dns packets.
- Increment `it` properly when skipping a record.
- Free aliases array in `getservby{name,port}` to prevent leaking memory.

### DIFF
--- a/options/posix/generic/lookup.cpp
+++ b/options/posix/generic/lookup.cpp
@@ -54,7 +54,7 @@ static frg::string<MemoryAllocator> read_dns_name(char *buf, char *&it) {
 		char code = *it++;
 		if ((code & 0xC0) == 0xC0) {
 			// pointer
-			uint8_t offset = ((code & 0x3F) << 8) | *it++;
+			uint16_t offset = (static_cast<uint8_t>(code & 0x3F) << 8) | static_cast<uint8_t>(*it++);
 			auto offset_it = buf + offset;
 			return res + read_dns_name(buf, offset_it);
 		} else if (!(code & 0xC0)) {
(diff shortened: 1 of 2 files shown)


Dennisbonke (approved):
LGTM

options/posix/generic/lookup.cpp
```diff
@@ -54,7 +54,7 @@ static frg::string<MemoryAllocator> read_dns_name(char *buf, char *&it) {
 		char code = *it++;
 		if ((code & 0xC0) == 0xC0) {
 			// pointer
-			uint8_t offset = ((code & 0x3F) << 8) | *it++;
+			uint16_t offset = ((uint8_t)(code & 0x3F) << 8) | (uint8_t)(*it++);
```
no92: It would be nicer to use static_casts here.

options/posix/generic/lookup.cpp
no92: Doesn't this have sign-extension problems? It seems like this should get the same static_cast treatment as the first hunk.

no92, line 226: The default case also needs to advance `it` by `rr_length` AFAICT.

options/posix/generic/netdb.cpp
no92: As a related fix, shouldn't this `for` loop be guarded by a `if (ret.s_aliases)` to prevent dereferencing a pointer without checking?

no92: As a related fix, shouldn't this `for` loop be guarded by a `if (ret.s_aliases)` to prevent dereferencing a pointer without checking?

no92 (approved):

PR #1809, opened 2026-07-01
Small bugfixes found while running KDE

Dennisbonke:
And a drive by stub function that was requested by sudo.

### DIFF
--- a/options/glibc/generic/shadow.cpp
+++ b/options/glibc/generic/shadow.cpp
@@ -244,3 +244,5 @@ struct spwd *sgetspent(const char *s) {
 		return &sp;
 	return nullptr;
 }
+
+void setspent(void) {}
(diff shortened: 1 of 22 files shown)


options/glibc/generic/shadow.cpp
```diff
@@ -244,3 +244,7 @@ struct spwd *sgetspent(const char *s) {
 		return &sp;
 	return nullptr;
 }
+
+void setspent(void) {
+	mlibc::infoLogger() << "mlibc: setspent is a stub" << frg::endlog;
```
no92: We shouldn't log here.

options/internal/generic/threads.cpp
```diff
@@ -183,6 +205,8 @@ __attribute__ ((__noreturn__)) void thread_exit(thread_exit_return ret_val) {
 		frg::destruct(getAllocator(), hand);
 	}
 
+	run_thread_local_destructors();
```
no92: This should probably run before the cleanup handlers.

no92 (approved):

PR #1810, opened 2026-07-02
sysdeps/managarm: use dedicated messages for epoll requests

no92:
Cyclic dependency on managarm side.

### DIFF
--- a/sysdeps/managarm/generic/file.cpp
+++ b/sysdeps/managarm/generic/file.cpp
@@ -1342,8 +1342,7 @@ int Sysdeps<EpollPwait>::operator()(
 
 	SignalGuard sguard;
 
-	managarm::posix::CntRequest<SysdepsAllocator> req(getSysdepsAllocator());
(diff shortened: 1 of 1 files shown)


avdgrinten (approved):

sysdeps/managarm/generic/file.cpp
```diff
@@ -1105,13 +1105,10 @@ int Sysdeps<Pselect>::operator()(
 }
 
 int Sysdeps<Poll>::operator()(struct pollfd *fds, nfds_t count, int timeout, int *num_events) {
-	__ensure(timeout >= 0 || timeout == -1); // TODO: Report errors correctly.
```
avdgrinten, line 1108: I assume that this is already handled correctly on the POSIX side?

no92: Yes, we have the same behavior as linux, where negative values are treated as infinite timeouts.

PR #1811, opened 2026-07-03
sysdeps/managarm: Fix races between submission and cancellation

avdgrinten:
Let mlibc send cancellation requests over the fs protocol (instead of doing it in posix-subsystem). This fixes a race: posix-subsystem does not know whether mlibc already got to the point where the request is submitted, so it can send cancellations out before the actual request is submitted. Sending the cancellation request from within mlibc avoids this race.


Dennisbonke (approved):
LGTM

PR #1812, opened 2026-07-03
Switch from frg::slab_pool to frg::sharded_slab_pool

avdgrinten:
`frg::sharded_slab` scales to multiple threads and it also supports arbitrary alignment.

### DIFF
--- a/options/internal/m68k/atomics.cpp
+++ b/options/internal/m68k/atomics.cpp
@@ -1,0 +1,84 @@
+#include <frg/mutex.hpp>
+#include <frg/spinlock.hpp>
+#include <stdbool.h>
+#include <stdint.h>
+
+namespace {
+
+frg::simple_spinlock atomic_locks[32] = {};
+
+inline frg::simple_spinlock &get_lock(const volatile void *ptr) {
+	uintptr_t hash = ((uintptr_t)ptr >> 3) & 31;
+	return atomic_locks[hash];
+}
+
+} // namespace
+
+extern "C" {
+
+// We need this pragma to avoid compiler warnings about differing declarations;
+// the compiler has these functions declared with default visibility, which we
+// explicitly want to avoid.
+#pragma GCC visibility push(hidden)
+
+uint64_t __atomic_load_8(const volatile void *ptr, int model [[maybe_unused]]) {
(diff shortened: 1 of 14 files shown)


options/internal/m68k/atomics.cpp
```diff
@@ -0,0 +1,84 @@
+#include <frg/mutex.hpp>
+#include <frg/spinlock.hpp>
+#include <stdbool.h>
+#include <stdint.h>
+
+namespace {
+
+frg::simple_spinlock atomic_locks[32] = {};
```
avdgrinten: We could just use a proper mlibc lock here (as `options/internal` has a futex-based lock).

options/internal/m68k/atomics.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+inline frg::simple_spinlock &get_lock(const volatile void *ptr) {
+	uintptr_t hash = ((uintptr_t)ptr >> 3) & 31;
```
avdgrinten: This is a poor hash function, for example, it will map all page aligned atomics to the same spot.

options/internal/m68k/atomics.cpp
```diff
@@ -0,0 +1,84 @@
...
+}
+
+} // namespace
+
+extern "C" {
+
+// We need this pragma to avoid compiler warnings about differing declarations;
+// the compiler has these functions declared with default visibility, which we
+// explicitly want to avoid.
+#pragma GCC visibility push(hidden)
```
avdgrinten: Do the functions actually end up with hidden visibility?

options/internal/m68k/atomics.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+uint64_t __atomic_load_8(const volatile void *ptr, int model [[maybe_unused]]) {
+	auto &l = get_lock(ptr);
+	l.lock();
+	uint64_t ret;
+	__builtin_memcpy(&ret, (const void *)ptr, 8);
```
avdgrinten: `__builtin_memcpy` seems overly pessimistic (e.g., it assumes no alignment at all etc). What about a pointer with may_alias attribute?

PR #1821, opened 2026-07-07
options/posix: do not log on `_SC_GETGR_R_SIZE_MAX`

no92:
This is a lib-level buffer size recommendation, and probably shouldn't be handled by sysdeps; the fallback is the expected use-case. As such, we shoudl avoid polluting output.


Mathewnd (approved):

Dennisbonke (approved):
LGTM

PR #1823, opened 2026-07-13
[ABI BREAK] options/ansi: always have time_t be 64-bit wide

### LINKED ISSUE #1600: Time types could overflow on 32-bit systems
While writing the BadgerOS sysdeps (which I will eventually seek to upstream), I noticed that certain types, most notably including `time_t` are defined as `long`, which would make them 32-bit on 32-bit systems. For time, however, this has the 2038 bug as `struct timespec` would contain a signed 32-bit integer for `tv_sec`. I would like to request an option to make these `__mlibc_int64` instead so users can enable a 2038-safe type on 32-bit platforms as well.

no92:
This is required by POSIX.


sysdeps/linux/generic/sysdeps.cpp
```diff
@@ -2533,14 +2891,22 @@ int Sysdeps<Fremovexattr>::operator()(int fd, const char *name) {
 }
 
 int Sysdeps<Statfs>::operator()(const char *path, struct statfs *buf) {
+#if defined(SYS_statfs64)
+	auto ret = do_cp_syscall(SYS_statfs64, path, buf);
+#else
 	auto ret = do_cp_syscall(SYS_statfs, path, buf);
+#endif
```
avdgrinten, line 3077: The ABI is wrong for statfs64.

sysdeps/linux/generic/sysdeps.cpp
```diff
@@ -2533,14 +2891,22 @@ int Sysdeps<Fremovexattr>::operator()(int fd, const char *name) {
...
 		return e;
 	return 0;
 }
 
 int Sysdeps<Fstatfs>::operator()(int fd, struct statfs *buf) {
+#if defined(SYS_fstatfs64)
+	auto ret = do_cp_syscall(SYS_fstatfs64, fd, buf);
+#else
 	auto ret = do_cp_syscall(SYS_fstatfs, fd, buf);
+#endif
```
avdgrinten, line 3088: Likewise.

sysdeps/linux/generic/sysdeps.cpp
```diff
@@ -2238,13 +2514,91 @@ int Sysdeps<Nice>::operator()(int increment, int *new_nice) {
...
+	  msg_rtime{static_cast<unsigned long>(m.msg_rtime)},
+	  msg_rtime_high{static_cast<unsigned long>(m.msg_rtime >> 32)},
+	  msg_ctime{static_cast<unsigned long>(m.msg_ctime)},
+	  msg_ctime_high{static_cast<unsigned long>(m.msg_ctime >> 32)},
+	  msg_cbytes{m.msg_cbytes},
+	  msg_qnum{m.msg_qnum},
+	  msg_qbytes{m.msg_qbytes},
+	  msg_lspid{m.msg_lspid},
+	  msg_lrpid{m.msg_lrpid} {}
+
```
avdgrinten, line 2689: Do we need the same mode shift here as for `kshmid64_ds`?

sysdeps/linux/generic/sysdeps.cpp
```diff
@@ -2238,13 +2514,91 @@ int Sysdeps<Nice>::operator()(int increment, int *new_nice) {
...
+		case IPC_STAT:
+		case IPC_RMID:
+			break;
+		case IPC_SET:
+			kbuf = kmsqid64_ds{*buf};
+			break;
+		case IPC_INFO:
+		default:
+			mlibc::infoLogger() << "mlibc: unexpected msgctl op " << cmd << frg::endlog;
+			break;
```
avdgrinten, line 2741: Likewise: there are other ops defined on the man page. Is this an intentional gap?

sysdeps/linux/generic/sysdeps.cpp
```diff
@@ -2145,10 +2380,51 @@ int Sysdeps<Shmat>::operator()(void **seg_start, int shmid, const void *shmaddr,
...
+		case IPC_STAT:
+		case IPC_RMID:
+			break;
+		case IPC_SET:
+			kbuf = kshmid64_ds{*buf};
+			break;
+		case IPC_INFO:
+		default:
+			mlibc::infoLogger() << "mlibc: unexpected shmctl op " << cmd << frg::endlog;
+			break;
```
avdgrinten, line 2547: There are other ops defined on the man page. Is this an intentional gap?

sysdeps/linux/generic/sysdeps.cpp
no92: Linux defines more ops, but we only declare these for now; other ops have weirder semantics, some of which require casting.

sysdeps/linux/generic/sysdeps.cpp
no92: Likewise.

sysdeps/linux/generic/sysdeps.cpp
no92: What mode?

no92: Nevermind, I'm blind and forgetful.

PR #1824, opened 2026-07-14
Fixes for KDE on BSD-likes + Zinnia

marv7000:


options/bsd/include/pthread_np.h
```diff
@@ -0,0 +1,16 @@
+#ifndef _PTHREAD_NP_H_
+#define _PTHREAD_NP_H_
+
+#include <pthread.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+int pthread_attr_get_np(pthread_t __thrd, pthread_attr_t *__attr);
```
no92, line 12: `#ifndef __MLIBC_ABI_ONLY`

options/rtld/generic/linker.cpp
```diff
@@ -421,6 +425,19 @@ frg::expected<LinkerError, SharedObject *> ObjectRepository::requestObjectAtPath
 		frg::destruct(getAllocator(), object);
 		return LinkerError::notFound;
 	}
+
+	// A different path can still resolve to an already-loaded object. Deduplicate on the file's device/inode.
```
no92, line 439: Is this even correct? Shouldn't deduplication logic work based on SONAME?

PR #1826, opened 2026-07-16
sysdeps/managarm: Various set/get id fixes for sudo

Dennisbonke:
Blocked on managarm/managarm#1492


no92 (changes requested):
Ideally, we should turn this into a single request such that this operation is atomic, and not split into up to 3 requests.

PR #1829, opened 2026-07-18
sysdeps/ethereal: Upstream port

sasdallas:
This PR adds sysdeps for the [Ethereal Operating System](https://github.com/sasdallas/Ethereal).\
They were ported from an mlibc version I forked almost 8 months ago, tested and everything seems to work okay.

Left `MLIBC_MMAP_ALLOCATE_DSO` off because it makes debugging easier in my own personal opinion


sysdeps/ethereal/generic/ptrace.cpp
```diff
@@ -0,0 +1,44 @@
...
+#include <stdarg.h>
+#include <errno.h>
+
+#include <bits/ensure.h>
+#include <mlibc/debug.hpp>
+#include <sys/types.h>
+#include <stdarg.h>
+#include <sys/syscall.h>
+
+static int __ethereal_ptrace(long req, pid_t pid, void *addr, void *data, long *out) {
```
no92: Prefer anon namespaces over static functions.

sysdeps/ethereal/generic/thread.cpp
```diff
@@ -0,0 +1,57 @@
...
+#include <mlibc/arch-defs.hpp>
+#include <sys/syscall.h>
+#include <sys/syscall_nums.h>
+
+extern "C" void __mlibc_thread_trampoline(void *(*fn)(void *), Tcb *tcb, void *arg) {
+	while (__atomic_load_n(&tcb->tid, __ATOMIC_RELAXED) == 0) {
+		mlibc::sysdep<FutexWait>(&tcb->tid, 0, nullptr);
+	}
+
+	tcb->invokeThreadFunc(reinterpret_cast<void *>(fn), arg);
```
no92, line 19: You should enable cancellation here, otherwise you might get unexpected hangs. You can use other sysdeps (e.g. astral) for reference.

no92 (commented):
It would be nice to have the new files be clang-formatted right away. This should also fix the mixed indentation.

One thing of note are the abi-bits: we generally only place symlinks in `sysdeps/os/include/abi-bits`, and definitions should go under `abis/os/` and be symlinked. Also, you are symlinking astral ABI in places; are you sure you want that? Astral ABI could change, and you'd end up subscribing to possible ABI breaks that way.

no92:
Also, given that this is an atomic change, squashing the commits into one would be nice.

sasdallas:
No that is extremely weird? I remember symlinking Linux ABIs and rewriting needed. I'm away from home for a few days, will try to do the squash later

sasdallas:
As far as I can see that fixes most of the complaints? Let me know though

PR #1830, opened 2026-07-20
Add roxy sysdeps

NikoPit:


sysdeps/roxy/sysdeps.cpp
```diff
@@ -0,0 +1,253 @@
+#include <mlibc/all-sysdeps.hpp>
+#include <roxy/syscall.h>
+
+#include <stdint.h>
+
+static int syscall_error(long result) {
```
no92, line 6: It would be nicer to put these in an anon namespace instead of making them static; I think this can also be constexpr.

sysdeps/roxy/sysdeps.cpp
```diff
@@ -0,0 +1,253 @@
...
+	return result < 0 ? static_cast<int>(-result) : 0;
+}
+
+static int syscall_result(long result, ssize_t *transferred) {
```
no92, line 10: Same here.

no92 (commented):
I gave this a preliminary look, even if this is drafted - I only have some nits:
- see the two comments
- it would be nice to squash all commits into one before merging

I see that you symlinked the Linux ABI bits, which is perfectly valid - I just want to spell out that this means that you are effectively subscribed to that; if it turns out that some definition in there was wrong (i.e. different from what it should be in Linux), that might be an ABI break.

Other than that, the PR looks solid. Once you deem this ready (i.e. by undrafting), I will set up the CI run for this, so no need to worry about that.

no92:
Is this supposed to ever be upstreamed? If you just want to keep your own commits on a branch on your fork, keeping open a draft PR is unnecessary.

PR #1839, opened 2026-07-25
sysdeps/ethereal: Implement get_nprocs along with misc. system calls

sasdallas:
Implements stubs along with the copied Linux interface `get_nprocs()`. Not quite sure why but a lot of things get annoyed at me if it's not there.

All newly added system calls except for `InetConfigured` are stubbed, likely will be implemented later.

Tested and working build on my host.


Dennisbonke (approved):
LGTM

Issue #1837, opened 2026-07-24
pthread_mutex_lock faults during constructors


sasdallas:
Tested via the Bochs application.

In this code:
```cpp
struct ScopeTrace {
	ScopeTrace(const char *file, int line, const char *function)
	: _file(file), _line(line), _function(function) {
		if(!mlibc::globalConfig().debugPthreadTrace)
			return;
		mlibc::infoLogger() << "trace: Enter scope "
				<< _file << ":" << _line << " (in function "
				<< _function << ")" << frg::endlog;
	}

	~ScopeTrace() {
		if(!mlibc::globalConfig().debugPthreadTrace)
			return;
		mlibc::infoLogger() << "trace: Exit scope" << frg::endlog;
	}

private:
	const char *_file;
	int _line;
	const char *_function;
};
```

The line `mlibc::globalConfig().debugPthreadTrace` will cause a fault if done from the constructors (in this case libgcc). By forcing it to 0 it fixes issue. 

My personal error is in `__register_frame_info_bases` right after `__gthread_mutex_lock`

PR #1840, opened 2026-07-26
Fix init/fini ordering in static builds

riblanc:
## The problem

In a static build mlibc's `.init_array` and `.fini_array` entries share the arrays with the executable's and are ordered by link order, which puts `libc.a` last.
mlibc therefore initialises after the program's constructors and, since fini arrays run in reverse, finalises before its destructors.

From a `[[gnu::constructor]]`: `printf()` faulted on an unconstructed `stdout`, `getenv()` returned `NULL` because `init_libc()` had not populated `environ` yet, and `at_quick_exit()` faulted on `quickExitQueue`'s null allocator.
`GlobalConfig` is built on first use and reads the environment, so anything reaching it that early also latched every `MLIBC_DEBUG_*` option to false for the rest of the process.

From a `[[gnu::destructor]]`: `printf()` used a `stdout` that `__cxa_finalize()` had already destroyed, ending in a pure virtual call.

```c
#include <stdio.h>

__attribute__((constructor)) void c(void) { printf("constructor\n"); }
__attribute__((destructor))  void d(void) { printf("destructor\n"); }

int main(void) { printf("hello\n"); return 0; }
```

Linked statically against mlibc this faults in the constructor; replacing the body with a raw `write(2)` gets past it and it then faults in the destructor.
The same program is fine against glibc, and fine against a dynamically linked mlibc, where `ld.so` initialises `libc.so` as a separate object before the executable.

Note that `constructor(101)` does not help: prioritised entries are placed before unprioritised ones, so the program's constructor moves even further ahead of mlibc.

## Why this is a bug rather than a limitation

Two comments in the tree already state the intent, and link order defeats both.

`__mlibc_do_finalize()` in `options/lsb/generic/dso_exit.cpp` deliberately avoids destroying mlibc's globals so that they stay "available to `[[gnu::destructor]]` functions which we invoke below".
But `__mlibc_do_destructors` is itself an entry in the fini array, and in a static build it is the last one, so it runs first.

`GlobalConfigGuard` in `options/internal/generic/global-config.cpp` forces the config to be built "during initialization of libc.so".
It is an unprioritised `.init_array` entry, so in a static build that happens after the program's constructors instead.

`mlibc-static` is also a first-class CI configuration rather than a best-effort one, and both glibc and musl initialise stdio and the environment before walking `.init_array`.

## The fix

`init_libc()` and the construction of the standard streams get priorities from the range reserved for the implementation, so they run before any constructor a program can register; 101 is the lowest a program may use.

For shutdown, rather than ordering mlibc against `__cxa_finalize()`, the standard streams live in `frg::manual_box` and are never destroyed, with `mlibc::flush_all_files()` flushing them at the end of `__mlibc_do_finalize()`.
The fini array itself is unchanged, so a program's C++ globals are still destroyed before its `[[gnu::destructor]]` functions, matching glibc.

`quickExitQueue` moves behind `lazy_eternal`, as `dso_exit.cpp` already does for the atexit queue.

## Testing

Ran the CI workflow locally with `act` on `debian:trixie`: `mlibc-static` and `mlibc-ansi-only` with gcc 14.2, and `mlibc` with clang 19.1.7, all with `-Dwerror=true` and `-Db_sanitize=undefined`.
The only failures are `posix/getaddrinfo`, which fails identically in its `host-libc` variant because the container has no DNS.
Every commit also builds on its own, and `pre-commit run --all-files` passes.

Behaviour was compared against glibc throughout: static and dynamic mlibc now produce identical output, in C and in C++.

## Follow-ups, not addressed here

Other translation units still initialise after the program's constructors, so anything a constructor reaches inside them is exposed to the same problem:

- `pthread.cpp` builds `pthread_signal_installer`, which installs the cancellation signal handler.
- `locale.cpp` has `locpaths`, `localeFiles` and `localeArchive`. The `"C"` locale path does not reach them, but loading a locale from an archive does.
- `utmpx.cpp`, and the POSIX timer context in `sysdeps/linux/generic/sysdeps.cpp`.

Each needs its own reproduction scenario, so I left them alone.

#1837 may well be the same family, but I could not reproduce it: `pthread_mutex_lock()` from a constructor works on the linux sysdep both before and after this series, so I am not claiming this fixes it.

### DIFF
--- a/options/ansi/generic/file-io.cpp
+++ b/options/ansi/generic/file-io.cpp
@@ -14,11 +14,13 @@
 
 #include <abi-bits/fcntl.h>
 #include <frg/allocation.hpp>
+#include <frg/manual_box.hpp>
 #include <frg/mutex.hpp>
 #include <frg/scope_exit.hpp>
 #include <mlibc/all-sysdeps.hpp>
 #include <mlibc/allocator.hpp>
 #include <mlibc/file-io.hpp>
+#include <mlibc/init-priority.hpp>
 #include <mlibc/lock.hpp>
 #include <mlibc/exit.hpp>
 
@@ -626,27 +628,41 @@ int fd_file::parse_modestring(const char *mode) {
 } // namespace mlibc
 
 namespace {
-	mlibc::fd_file stdin_file{0};
-	mlibc::fd_file stdout_file{1};
-	mlibc::fd_file stderr_file{2, nullptr, true};
+	// Never destroyed: manual_box is trivially destructible, so __cxa_finalize()
+	// cannot tear the streams down while [[gnu::destructor]] functions may still
+	// use them. mlibc::flush_all_files() flushes them at exit.
+	constinit frg::manual_box<mlibc::fd_file> stdin_box;
+	constinit frg::manual_box<mlibc::fd_file> stdout_box;
+	constinit frg::manual_box<mlibc::fd_file> stderr_box;
 
-	struct stdio_guard {
-		stdio_guard() { }
+	[[gnu::constructor(MLIBC_INIT_PRIORITY_STDIO)]]
+	void init_stdio() {
+		stdin_box.initialize(0);
+		stdout_box.initialize(1);
+		stderr_box.initialize(2, nullptr, true);
 
-		~stdio_guard() {
-			// Only flush the files but do not close them.
-			for(auto it : mlibc::global_file_list()) {
-				if(int e = it->flush(); e && !mlibc::processIsExiting.load(std::memory_order_relaxed))
-					mlibc::infoLogger() << "mlibc warning: Failed to flush file before exit()"
-							<< frg::endlog;
-			}
-		}
-	} global_stdio_guard;
+		stdin = stdin_box.get();
+		stdout = stdout_box.get();
+		stderr = stderr_box.get();
+	}
 } // namespace
 
-FILE *stderr = &stderr_file;
-FILE *stdin = &stdin_file;
-FILE *stdout = &stdout_file;
+FILE *stderr;
+FILE *stdin;
+FILE *stdout;
+
+namespace mlibc {
+
+void flush_all_files() {
+	// Only flush the files but do not close them.
+	for(auto it : global_file_list()) {
+		if(int e = it->flush(); e && !processIsExiting.load(std::memory_order_relaxed))
+			infoLogger() << "mlibc warning: Failed to flush file before exit()"
+					<< frg::endlog;
+	}
+}
+
+} // namespace mlibc
 
 int fileno_unlocked(FILE *file_base) {
 	auto file = static_cast<mlibc::fd_file *>(file_base);
--- a/meson.build
+++ b/meson.build
@@ -126,6 +126,10 @@ if not headers_only
 	add_project_arguments('-Werror=misleading-indentation', language: ['c', 'cpp'])
 	add_project_arguments('-fno-rtti', '-fno-exceptions', language: 'cpp')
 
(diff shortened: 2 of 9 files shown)


options/ansi/generic/file-io.cpp
```diff
@@ -626,27 +627,74 @@ int fd_file::parse_modestring(const char *mode) {
...
+	// init_stdio() then constructs the objects in place, and
+	// mlibc::flush_all_files() flushes them at exit.
+	union eternal_fd_file {
+		constexpr eternal_fd_file() : dummy{} { }
+		// Deliberately does not destroy `file`. See above.
+		~eternal_fd_file() { }
+
+		char dummy;
+		mlibc::fd_file file;
+	};
```
avdgrinten: This can use `frg::manual_box` instead of re-implementing the same pattern.

avdgrinten (commented):
The changes look technically correct in general.

Two style issues:
- Comments and commit messages are unnecessarily verbose. Please condense them / trim them down to be in line with mlibc's existing code style.
- I assume that this was authored with the help of coding agents, so please add an `Assisted-by` trailer to the commits (like some other recent commits do, e.g., caad55f61d7f9baec519bebcbb289c80264eccfd).

riblanc:
Thanks for the review.

Yes, this was assisted by a coding agent, and I've added the `Assisted-by` trailer to the commits where it applies. English isn't my first language and I leaned on it a fair bit for the prose, which is where most of the bloat came from. I've trimmed the comments down to the 1-3 line range used around them and cut the commit bodies down.

I also rewrote the PR description as it suffers from the same problem, and a good part of it argued for the hand-rolled union that `manual_box` replaced, so it was out of date too.

options/ansi/generic/file-io.cpp
riblanc: Switched the streams over to `frg::manual_box`.

I was worried it would break `stdout`, since it is no longer statically initialised and a non-PIE executable could take a copy relocation on it. Neither GNU ld nor lld does here, and static and dynamic output still matches glibc.

Thank you for the suggestion, it is a clear improvement.

no92:
I'm not convinced that the original problem of printing in constructors is even valid - it rather seems like that would be UB anyway.

lzcunt:
Can you test whether a statically built executable piped to a file segfaults on exit with these changes? That's a bug I encountered and this might also fix that.

To reproduce, mlibc has to be statically linked and stdout has to be a non-tty file.

riblanc:
> I'm not convinced that the original problem of printing in constructors is even valid - it rather seems like that would be UB anyway.

Since constructor/destructor attributes are a GNU extension rather than ISO C, I am not sure UB is the right framing here: what libc state exists when a constructor runs looks implementation-defined to me, and both glibc and musl make stdio and environ work there.

The destructor side seems clearer cut, since [this comment](https://github.com/managarm/mlibc/blob/bd31a876074edd89a3abc78622b5f52eb10a6172/options/lsb/generic/dso_exit.cpp#L72-L76) already expects stdout to survive into `[[gnu::destructor]]` functions. It does not, because `__mlibc_do_destructors` is itself a fini array entry and runs before the program's, so `__cxa_finalize()` has already destroyed the streams. Statically linked, this prints only `hello`:
```c
#include <stdio.h>
__attribute__((destructor)) void d(void) { printf("destructor\n"); }
void main(void) { printf("hello\n"); }
```
The write lands in a freed buffer and is lost silently

@lzcunt I could not reproduce the bug you described with a statically linked binary, either before or after these changes. Do you have a more detailed case ?

avdgrinten (approved):

PR #1841, opened 2026-07-30
[ABI BREAK] Various fixes from review

no92:
Tested with managarm, KDE and the browsers still work.

### DIFF
--- a/options/internal/include/bits/pthread_types.h
+++ b/options/internal/include/bits/pthread_types.h
@@ -45,11 +45,11 @@ typedef struct {
 
 typedef struct {
 	int __opaque[__MLIBC_THREAD_ONCE_SIZE / sizeof(int)];
-} pthread_once_t;
+} __attribute__((aligned(4))) pthread_once_t;
 
 typedef struct {
 	int __opaque[__MLIBC_THREAD_SPINLOCK_SIZE / sizeof(int)];
-} pthread_spinlock_t;
+} __attribute__((aligned(4))) pthread_spinlock_t;
 
 #ifdef __cplusplus
 }
--- a/options/posix/generic/spawn.cpp
+++ b/options/posix/generic/spawn.cpp
@@ -10,6 +10,7 @@
 #include <sched.h>
 #include <signal.h>
 #include <spawn.h>
+#include <stdarg.h>
 #include <sys/wait.h>
 #include <unistd.h>
 
@@ -46,8 +47,20 @@ struct args {
 	int (*exec_fn)(const char *path, char *const argv[], char *const env[]);
 };
 
+int sysdep_fcntl(int fd, int request, ...) {
+	va_list args;
+	int result = 0;
+	va_start(args, request);
+	int e = mlibc::sysdep_or_enosys<Fcntl>(fd, request, args, &result);
+	va_end(args);
+	if (e)
+		return -e;
+	return result;
+}
+
 int child(void *args_vp) {
-	int i, ret;
+	int i;
+	int ret = 0;
 	struct sigaction sa = {};
 	struct args *args = (struct args *)args_vp;
 	int p = args->p[1];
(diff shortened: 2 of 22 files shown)


options/posix/generic/spawn.cpp
```diff
@@ -190,8 +227,13 @@ int child(void *args_vp) {
 fail:
 	/* Since sizeof errno < PIPE_BUF, the write is atomic. */
 	ret = -ret;
```
Mathewnd: This is kind of confusing. Every write to ret is a negated error number, only for it to be negated back here? Why can't we just write positive error numbers from the beggining?

Mathewnd (approved):
LGTM other than a small detail

options/internal/include/bits/pthread_types.h
```diff
@@ -45,11 +45,11 @@ typedef struct {
 
 typedef struct {
 	int __opaque[__MLIBC_THREAD_ONCE_SIZE / sizeof(int)];
-} pthread_once_t;
+} __attribute__((aligned(__INTPTR_WIDTH__ / 8))) pthread_once_t;
```
avdgrinten: I think alignment 4 (= `sizeof(int)`) is fine here.

options/internal/include/bits/pthread_types.h
```diff
@@ -45,11 +45,11 @@ typedef struct {
...
 
 typedef struct {
 	int __opaque[__MLIBC_THREAD_SPINLOCK_SIZE / sizeof(int)];
-} pthread_spinlock_t;
+} __attribute__((aligned(__INTPTR_WIDTH__ / 8))) pthread_spinlock_t;
```
avdgrinten: Likewise.

options/internal/include/bits/pthread_sizes.h
```diff
@@ -5,7 +5,7 @@
 #define __MLIBC_THREAD_MUTEX_SIZE 64
 #define __MLIBC_THREAD_COND_SIZE 64
 #define __MLIBC_THREAD_RWLOCK_SIZE 64
-#define __MLIBC_THREAD_ONCE_SIZE 4
-#define __MLIBC_THREAD_SPINLOCK_SIZE 4
+#define __MLIBC_THREAD_ONCE_SIZE (__INTPTR_WIDTH__ / 8)
+#define __MLIBC_THREAD_SPINLOCK_SIZE (__INTPTR_WIDTH__ / 8)
```
avdgrinten: With the alignment changed to 4, we can revert this hunk.

sysdeps/linux/generic/sysdeps.cpp
```diff
@@ -555,8 +555,9 @@ int Sysdeps<Sigaction>::operator()(int signum, const struct sigaction *act,
 	struct ksigaction kernel_act, kernel_oldact;
 	if (act) {
 		kernel_act.handler = act->sa_handler;
-		kernel_act.flags = act->sa_flags | SA_RESTORER;
+		kernel_act.flags = static_cast<unsigned long>(act->sa_flags);
```
avdgrinten: This sign extends to 64 bit, we probably want zero extension.

options/posix/generic/spawn.cpp
```diff
@@ -229,7 +271,9 @@ int posix_spawn_impl(pid_t *__restrict res, const char *__restrict path,
 	/* Mlibc change: We use fork + execve, as clone is not implemented.
 	 * This yields the same result in the end. */
 	//pid = clone(child, stack + sizeof stack, CLONE_VM | CLONE_VFORK | SIGCHLD, &args);
-	pid = fork();
+	if(ec = mlibc::sysdep_or_enosys<Fork>(&pid); ec)
+		goto fail;
```
avdgrinten, line 278: Fork failures leak the fds allocated above. We could either add `frg::scope_exit` or defer to an issue for this.

avdgrinten: Also, `fork()` does some extra stuff such as refreshing the cached TID. It's probably fine not to do that here though since we `execve()` anyway.

avdgrinten:
We also need to fix `SIOCGSTAMP` (potentially by deleting it?).

PR #1843, opened 2026-07-31
Smaller fixes from review

no92:


options/internal/aarch64/assembly-asserts.cpp
```diff
@@ -0,0 +1,4 @@
+#include <stddef.h>
+#include <setjmp.h>
+
+static_assert(offsetof(__jmp_buf, __savesigs) == 176);
```
avdgrinten: In Thor, we handle this more robustly by having a `#define` for the offset, including that `#define` both from C++ and assembly, and static asserting in C++.

avdgrinten (approved):

PR #1845, opened 2026-07-31
options/ansi: lock the global file list to avoid concurrent modifications

lasiotus:


no92:
The code itself looks correct to me, however the commit message should follow the preexisting style: `options/ansi: lock the global file list to avoid concurrent modifications`

lasiotus:
> The code itself looks correct to me, however the commit message should follow the preexisting style: `options/ansi: lock the global file list to avoid concurrent modifications`

Done.

PR #1846, opened 2026-07-31
posix_spawn improvements

no92:


options/posix/generic/spawn.cpp
```diff
@@ -411,6 +411,10 @@ int posix_spawn_file_actions_destroy(posix_spawn_file_actions_t *file_actions) {
 
 int posix_spawn_file_actions_adddup2(posix_spawn_file_actions_t *file_actions,
 		int fildes, int newfildes) {
+	// POSIX-mandated fd checks
+	if (fildes < 0 || fildes >= OPEN_MAX || newfildes < 0 || newfildes >= OPEN_MAX)
+		return EBADF;
```
avdgrinten, line 440: A check against `OPEN_MAX` is wrong because `OPEN_MAX` is not the true limit. We either need to use sysconf or just drop the check.

options/posix/generic/spawn.cpp
```diff
@@ -427,6 +435,10 @@ int posix_spawn_file_actions_addclose(posix_spawn_file_actions_t *file_actions,
 
 int posix_spawn_file_actions_addopen(posix_spawn_file_actions_t *__restrict file_actions,
 		int fildes, const char *__restrict path, int oflag, mode_t mode) {
+	// POSIX-mandated fd checks
+	if (fildes < 0 || fildes >= OPEN_MAX)
+		return EBADF;
```
avdgrinten: Likewise.

options/posix/generic/spawn.cpp
```diff
@@ -343,13 +343,25 @@ int posix_spawnattr_setflags(posix_spawnattr_t *attr, short flags) {
 	return 0;
 }
 
+int posix_spawnattr_getflags(const posix_spawnattr_t *__restrict attr, short *flags) {
```
avdgrinten: POSIX has `restrict` on both arguments (same for the other functions).

avdgrinten (commented):
Also, the new attributes are not covered by tests.

options/posix/generic/spawn.cpp
no92: Checking against `OPEN_MAX` is required by POSIX, though.

avdgrinten (approved):

PR #1850, opened 2026-08-02
options/ansi: use tgkill for raise if supported by sysdep

Mathewnd:
This fixes raise() to match the behavior expected by POSIX:

```
The raise() function shall send the signal sig to the executing thread or process. 
If a signal handler is called, the raise() function shall not return until after the signal handler does.

If the implementation supports the Threads option, the effect of the raise() function shall be equivalent to calling:

pthread_kill(pthread_self(), sig);

Otherwise, the effect of the raise() function shall be equivalent to calling:

kill(getpid(), sig);
```

I have hit issues related to this while debugging webkit, where it would hit a ``ud2`` on abort because the signal went to another thread.

### DIFF
--- a/options/ansi/generic/signal.cpp
+++ b/options/ansi/generic/signal.cpp
@@ -5,6 +5,7 @@
 
 #include <mlibc/all-sysdeps.hpp>
 #include <mlibc/debug.hpp>
+#include <mlibc/tid.hpp>
 
 __sighandler signal(int sn, __sighandler handler) {
 	struct sigaction sa;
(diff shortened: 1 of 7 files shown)


options/ansi/generic/signal.cpp
```diff
@@ -20,10 +20,21 @@ __sighandler signal(int sn, __sighandler handler) {
...
 
 int raise(int sig) {
-	MLIBC_CHECK_OR_ENOSYS(mlibc::IsImplemented<GetPid> && mlibc::IsImplemented<Kill>, -1);
-	pid_t pid = mlibc::sysdep_or_panic<GetPid>();
+	int e;
+
+#if __MLIBC_POSIX_OPTION
+	if constexpr (mlibc::IsImplemented<Tgkill>) {
+		pid_t pid = mlibc::sysdep_or_panic<GetPid>();
+		pid_t tid = mlibc::sysdep_or_panic<GetTid>();
```
no92: To avoid the sysdep, shouldn't we use `mlibc::this_tid()`?

PR #1851, opened 2026-08-02
options/ansi: properly return EOF on failed fflush(NULL)

Mathewnd:


options/ansi/generic/file-io.cpp
```diff
@@ -737,13 +737,16 @@ long ftell(FILE *file_base) {
 int fflush_unlocked(FILE *file_base) {
 	if(file_base == nullptr) {
 		// Only flush the files but do not close them.
+		int result = 0;
 		frg::unique_lock list_lock(mlibc::global_file_list_mutex);
 		for(auto it : mlibc::global_file_list()) {
-			if(int e = it->flush(); e)
+			if(int e = it->flush(); e) {
+				result = EOF;
```
no92, line 744: We also need to set the error indicator on the stream, and also set errno.

PR #1858, opened 2026-08-04
sysdeps/managarm: Implement renameat2

Dennisbonke:
Part of the Compiling GCC on Managarm project.

Blocked on managarm/managarm#1514


avdgrinten:
Instead of hacking around this, we should just add the `flags` argument to the renameat IPC message and implement both renameat and renameat2 through the same IPC message.

avdgrinten (approved):

PR #1860, opened 2026-08-06
options/rtld+sysdeps/managarm: early exit on foreign systems

lzcunt:
Autoconf checks love to build and run test binaries, even when cross compiling. This can get e.g. Managarm binaries executing on a foreign kernel (e.g. Linux). To prevent damage that could be caused by executing random syscalls, and in the Managarm case preventing HEL_CHECK(helPanic()) from spinning forever, let's ensure we are running on the correct system before doing anything crazy.

The bug was found with the help of Claude Fable 5, all code in this PR is either written by me or very thoroughly checked.


lzcunt:
Replaced by managarm/managarm#1520.

PR #1861, opened 2026-08-08
rtld: Use DTV indexes, not pointers as __tls_get_addr argument

avdgrinten:
Prior to this change, mlibc resolved relocations of the first `__abi_tls_entry` word to a pointer to our internal `SharedObject` struct and `__tls_get_addr` used that struct to find the TLS offset.

However, it turns out that linkers may not always generate relocations for `__abi_tls_entry`. In particular, if a dynamic TLS object is statically linked into an executable, the linker will hard code `1` as the first `__abi_tls_entry` word. This is not visible on x86_64 and aarch64 where the whole `__tls_get_addr` call is eliminated in that case, but it faults on RISC-V.

### DIFF
--- a/options/lsb/generic/tls.cpp
+++ b/options/lsb/generic/tls.cpp
@@ -1,17 +1,7 @@
 #include <internal-config.h>
-#include <mlibc/thread.hpp>
 #include <mlibc/rtld-abi.hpp>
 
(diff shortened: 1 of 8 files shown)


avdgrinten:
The last commit fixes a pre-existing bug that was just not visible in CI because no test exercised it.

marv7000 (approved):

PR #1865, opened 2026-08-12
Building mlibc with a recent clang (22.x here, LLVM/Homebrew, libc++, -std=c++23) fails in netdb.cpp

bayar17:
error: alias template 'string' requires template arguments; argument deduction only allowed for class templates
frg::string is an alias template (using string = basic_string<char, Allocator>; in subprojects/frigg/include/frg/string.hpp), not a class template. Alias templates were never really supposed to support class template argument deduction, but older/looser compilers let frg::string{getAllocator()} slide anyway by deducing through to the underlying basic_string. C++23 tightened that up, and clang now enforces it. Any construction that relied on CTAD to fill in the allocator type stopped compiling.

Fix is mechanical: spell out the allocator at each construction site (frg::string<MemoryAllocator>{...}) instead of leaning on deduction. No behavior change.

Touches options/internal/generic/locale.cpp (locale name construction across the LC_* categories) and options/posix/generic/netdb.cpp (the one that was actually breaking my build). Splitting this out on its own since it's a general build fix, not tied to any particular sysdeps backend. It just happened to be what I hit first while getting a new backend building.

### DIFF
--- a/options/internal/generic/locale.cpp
+++ b/options/internal/generic/locale.cpp
@@ -145,7 +145,7 @@ frg::vector<frg::string<MemoryAllocator>, MemoryAllocator> buildLocaleList(Decom
 		if (codeset)
 			return normalizeCodeset(*codeset);
 		else
-			return frg::string{getAllocator()};
+			return frg::string<MemoryAllocator>{getAllocator()};
 	}();
 	frg::vector<frg::string<MemoryAllocator>, MemoryAllocator> ret{getAllocator()};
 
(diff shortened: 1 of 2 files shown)


no92:
Wouldn't a deduction guide in frigg solve this better?

```c++
namespace frg {

template<typename Allocator>
using string = basic_string<char, Allocator>;

template<typename Allocator>
string(Allocator) -> string<Allocator>;

} // namespace frg
```

bayar17:
> Wouldn't a deduction guide in frigg solve this better?
> 
> ```c++
> namespace frg {
> 
> template<typename Allocator>
> using string = basic_string<char, Allocator>;
> 
> template<typename Allocator>
> string(Allocator) -> string<Allocator>;
> 
> } // namespace frg
> ```

I didn't think of it. PS(I was too busy working with my kernel, and x11.)

PR #1866, opened 2026-08-12
sysdeps/robu: add robu support #2

bayar17:
This adds a sysdeps backend for Robu, a hobby x86_64 microkernel I've been building. Repo's at https://github.com/bayar17/robu_microkernel. Not Linux-ABI compatible, it's its own capability/IPC-based design (closer in spirit to sysdeps/managarm than to the Linux backends), with a devfs/procfs/sysfs/ramfs split exposed over a small generic VFS protocol rather than a single syscall surface.

On what this is actually running: real GNU bash 5.3, interactively, with working job-control-aware startup, readline (arrow keys, history, line editing), and true fork()/pipe()/execve(), not a fork emulation layer. There's also a coreutils-equivalent (minibox) and a handful of small services (mount daemons, a tty service, init) all built against this backend. No gcc port yet, Robu doesn't self-host a compiler at this point, but bash felt like the more meaningful bar to clear for an interactive POSIX-ish environment, and that's solidly working.

The commits build up roughly in the order I actually built it: initial backend and crt/entry setup, real file I/O and spawn/wait, fork/pipe/execve, then a migration of the fs clients onto one generic VFS protocol instead of four bespoke ones, followed by incremental sysdep additions (signals, termios/ioctl/pgid, pselect, uid/gid, the mkdir family, and a batch of smaller ones bash's startup path needed) and a couple of real bug fixes found along the way, a stdio lazy-init issue and a readdir cursor that wasn't being reset on fd reuse.

One dependency worth flagging: this currently needs #1865 (the frg::string C++23 fix) to build on a strict C++23 compiler. Hit that while building this backend, but it's a general issue unrelated to Robu specifically, so I split it out rather than bundling it in here.


no92:
Please rebase instead of having merge commits.

bayar17:
> Please rebase instead of having merge commits.

Rebase is complete

PR #1868, opened 2026-08-16
options/ansi: track the access mode of files

### LINKED ISSUE #1251: Failing MPFR tests
Two tests in MPFR 4.2.1 fails:
```
FAIL: tfpif
===========

mpfr_fpif_export did not fail on a read only stream
mlibc warning: Failed to flush file before exit()
FAIL tfpif (exit status: 1)

FAIL: tsprintf
==============

mlibc: Unknown printf terminator '97'
In function operator(), file ../options/ansi/generic/stdio.cpp:81
__ensure(!"Illegal printf terminator") failed
FAIL tsprintf (exit status: 132)
```

The second issue is simply #1102, the first is `fwrite()` succeeding on a read-only stream.

lzcunt:
abstract_file never recorded whether a stream was opened for reading,
writing or both, so a write to a read-only stream silently landed in the
buffer and only failed later at flush time. If the buffer was never
filled, it did not fail at all. Reads from a write-only stream had the
mirrored problem.

Record the access mode from the O_* flags the stream was opened with and
reject the operation with EBADF in abstract_file::{read,write}(), setting
the error indicator but not the EOF indicator.

Closes #1251, however there appears to be another printf bug that MPFR tests have caught, so we are not yet done with MPFR tests.

### DIFF
--- a/options/ansi/generic/file-io.cpp
+++ b/options/ansi/generic/file-io.cpp
@@ -104,12 +104,24 @@ void abstract_file::dispose() {
 	_do_dispose(this);
 }
 
+void abstract_file::set_access_mode(int flags) {
+	_readable = (flags & O_ACCMODE) != O_WRONLY;
+	_writable = (flags & O_ACCMODE) != O_RDONLY;
+}
+
 // Note that read() and write() are asymmetric:
 // While read() can trigger a write-back, write() can never trigger a read-ahead().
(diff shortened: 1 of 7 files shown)


Dennisbonke:
Partial duplicate / overlap with #1857 (the flattening errno commit)?

lzcunt:
Looks like it

PR #1869, opened 2026-08-16
options/posix+abis/linux: expose gated `sys/resource.h` XSI declarations

lzcunt:
These declarations were gated where neither glibc nor musl gates them.

Caught when compiling zstd for linux against mlibc. Introduces namespace pollution that was previously fixed but it's namespace pollution that zstd expects under linux. Possibly the gating could be relaxed but only for Linux in order to not introduce namespace pollution under other other targets.

### DIFF
--- a/abis/linux/resource.h
+++ b/abis/linux/resource.h
@@ -13,7 +13,6 @@ extern "C" {
 #define PRIO_MAX 20
 #endif /* defined(_GNU_SOURCE) */
 
-#if defined(_DEFAULT_SOURCE) || __MLIBC_XOPEN
 #define PRIO_PROCESS 0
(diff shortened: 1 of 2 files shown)


no92 (changes requested):
NAK, changing correct gating based on broken programs is pointless. Instead, figure out why the program does not work; often enough, their build system only define some FTMs or _GNU_SOURCE for select OSes, and breaking all others even if they are compliant and correct.

lzcunt:
The program in question (zstd) defines the correct feature test macros on every system other than Linux, where it sets `_POSIX_C_SOURCE 200809L` without `_XOPEN_SOURCE`. glibc and musl both expose these without any gating even if that's not entirely correct.

PR #1871, opened 2026-08-17
options/internal: let allocations fail instead of aborting

lzcunt:
frg::slab_pool already treats a zero return from the policy's map() as an
allocation failure and passes it on as a null pointer, so reporting the
error is all that is needed.

Found via OpenSSL's mem_alloc_test.


avdgrinten:
There is probably overlap with https://github.com/managarm/mlibc/pull/1812

PR #1872, opened 2026-08-18
options/posix: fix inet_pton() for embedded IPv4 with a zero octet

lzcunt:
This patch rejects "01" and "007" as before, accepts a plain "0", and
additionally rejects "00", which the old check accepted. This amtches
glibc for every form tested.

Note this only concerns IPv4 addresses embedded in an IPv6 address. IPv4
addresses are not affected.

Found via OpenSSL's tests

### DIFF
--- a/options/posix/generic/arpa-inet.cpp
+++ b/options/posix/generic/arpa-inet.cpp
@@ -301,7 +301,7 @@ int inet_pton(int af, const char *__restrict src, void *__restrict dst) {
 					if(ipv4end[0] != '.' || value0 > UINT8_MAX || src == ipv4end)
 						return 0;
 					// reject overlong segments or leading zeroes
-					if ((ipv4end - src) > 3 || (src[0] == '0' && value))
+					if ((ipv4end - src) > 3 || (src[0] == '0' && ipv4end - src > 1))
 						return 0;
 					src = ipv4end + 1;
 					if (!isdigit(*src))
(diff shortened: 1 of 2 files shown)


Dennisbonke:
I'm not sure if this is correct. [POSIX 2024](https://pubs.opengroup.org/onlinepubs/9799919799/functions/inet_pton.html) states the following (emphasis mine):

If the af argument of inet_pton() is AF_INET, the src string shall be in the standard IPv4 dotted-decimal form:

ddd.ddd.ddd.ddd

where "ddd" is a one to three digit decimal number between 0 and 255. **Leading zeros shall be allowed.** The inet_pton() function does not accept other formats (such as the octal numbers, hexadecimal numbers, and fewer than four numbers that [inet_addr()](https://pubs.opengroup.org/onlinepubs/9799919799/functions/inet_addr.html) accepts).

It looks like glibc diverges here, and this is an addition to the standard made in issue 8 of POSIX, but that doesn't mean we get to break it per se. We should think if we want to follow glibc here or if we can adhere to POSIX without breaking programs.

lzcunt:
> ddd.ddd.ddd.ddd
> 
> where "ddd" is a one to three digit decimal number between 0 and 255. **Leading zeros shall be allowed.**

That's IPv4 while this PR touches IPv4-embedded-in-IPv6. Not sure if the parsing for that is supposed to be any different. But I think we should follow whatever standards suggest.

I didn't allow leading zeros because that was the case before this patch.

PR #1873, opened 2026-08-18
options/ansi: gmtime stuff

lzcunt:
Unlike my other recent PRs that fixed stuff from OpenSSL and other packages' tests, this PR is AI-assisted. I don't feel confident in calendar/date stuff (my previous timezone patch was HELL) so I asked claude to fix these issues and reviewed it. Other PRs in this "series" have been written by me so far.

Caught by OpenSSL's tests again

### DIFF
--- a/tests/ansi/gmtime.c
+++ b/tests/ansi/gmtime.c
@@ -1,0 +1,58 @@
+#include <assert.h>
+#include <stdio.h>
+#include <string.h>
+#include <time.h>
+#include <stdint.h>
+
+// int64_t instead of time_t so this compiles for 32-bit time_t systems.
+static void check(int64_t time, int year, int mon, int mday, int hour, int min, int sec,
+		int yday, int wday) {
+	time_t t = (time_t) time;
+	struct tm tm;
+	memset(&tm, 0, sizeof(tm));
+	assert(gmtime_r(&t, &tm) == &tm);
+	printf("%lld -> %04d-%02d-%02d %02d:%02d:%02d yday=%d wday=%d\n", (long long)t,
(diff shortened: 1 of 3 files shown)


tests/ansi/gmtime.c
```diff
@@ -0,0 +1,58 @@
+#include <assert.h>
+#include <stdio.h>
+#include <string.h>
+#include <time.h>
+#include <stdint.h>
+
+// int64_t instead of time_t so this compiles for 32-bit time_t systems.
```
no92, line 7: ????
We removed 32-bit time_t support with the last ABI break.

tests/ansi/gmtime.c
```diff
@@ -0,0 +1,58 @@
...
+static void check(int64_t time, int year, int mon, int mday, int hour, int min, int sec,
+		int yday, int wday) {
+	time_t t = (time_t) time;
+	struct tm tm;
+	memset(&tm, 0, sizeof(tm));
+	assert(gmtime_r(&t, &tm) == &tm);
```
no92, line 13: `assert()` should be written to be side-effect free.

tests/ansi/gmtime.c
lzcunt: this is a test, it can be built against 32-bit time non-mlibc libc with cross-libc tests

no92: The problem is that POSIX requires time_t to be 64-bit :^)

lzcunt: the version of glibc on x86 on the github CI is 32-bit time.

lzcunt: I think CI passing matters more and this is an ANSI test, this can be ran on a system that is nowhere near POSIX.

PR #1874, opened 2026-08-19
options/{internal,linux,posix}: Fixes for fio

avdgrinten:
Small fixes to make `fio` work on Managarm.

### DIFF
--- a/options/internal/generic/getopt.cpp
+++ b/options/internal/generic/getopt.cpp
@@ -30,37 +30,42 @@ int getopt_common_internal(int argc, char * const argv[], const char *optstring,
 
 		frg::optional<size_t> i = frg::null_opt;
 
-		// first, attempt to find exactly one exact match
+		// First, attempt to find an exact match.
+		// Like glibc, take the first one even if a later entry has the same name;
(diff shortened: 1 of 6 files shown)


options/posix/generic/unistd.cpp
```diff
@@ -603,11 +603,28 @@ int lockf(int fd, int op, off_t size) {
...
+	if constexpr (mlibc::IsImplemented<Nice>) {
+		if(int e = mlibc::sysdep<Nice>(nice, &new_nice); e) {
+			errno = e;
+			return -1;
+		}
+		return new_nice;
+	}
+
+	// Like glibc, fall back to implementing nice() via the priority sysdeps.
+	int old_nice;
```
Mathewnd, line 615: Considering the Linux sysdeps implement Nice using this fallback, I think we could remove the Linux implementation, not only for deduplication but because it is also wrong (subtracts from 20 twice).

options/internal/generic/getopt.cpp
Mathewnd: Should the new behavior also be applied here?

Mathewnd (approved):
LGTM other than two minor things

Dennisbonke (approved):
LGTM

PR #1879, opened 2026-08-20
options/bsd: implement issetugid()

lzcunt:
musl, bionic, and the BSDs implement this too.

### DIFF
--- a/options/bsd/include/bits/bsd/bsd_unistd.h
+++ b/options/bsd/include/bits/bsd/bsd_unistd.h
@@ -1,0 +1,22 @@
+#ifndef _MLIBC_BSD_UNISTD_H
+#define _MLIBC_BSD_UNISTD_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <mlibc-config.h>
+
+#ifndef __MLIBC_ABI_ONLY
+
+#if defined(_DEFAULT_SOURCE) || defined(_BSD_SOURCE)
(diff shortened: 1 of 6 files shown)


options/bsd/include/bits/bsd/bsd_unistd.h
```diff
@@ -0,0 +1,22 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <mlibc-config.h>
+
+#ifndef __MLIBC_ABI_ONLY
+
+#if defined(_DEFAULT_SOURCE) || defined(_BSD_SOURCE)
```
no92, line 12: `_BSD_SOURCE` is deprecated, `_DEFAULT_SOURCE` should be enough.

lzcunt: It's deprecated in glibc. We don't ought to follow that especially if we intend on supporting NetBSD which while doesn't have _BSD_SOURCE, heavily makes use of _NETBSD_SOURCE. In my NetBSD port I'll wire that to imply _BSD_SOURCE for the purposes of DRY between BSDs. There's also _OPENBSD_SOURCE but it does not mean what you expect it to mean. Anyhow what I'm saying is that in glibc _BSD_SOURCE usage could be replaced with _DEFAULT_SOURCE but not in BSD libcs (which I want my NetBSD port to be source compatible with).

PR #1883, opened 2026-08-21
options/posix: fix namespace pollution in <sys/select.h>

lzcunt:
Found via libsodium tests.


no92 (commented):
> Inclusion of the <sys/select.h> header may make visible all symbols from the headers `<signal.h>` and `<time.h>`.

PR #1886, opened 2026-09-01
options/rtld: Skip objects w/o dynamic sections in resolveSymbol

ThatMishakov:
This fixes dlsym() and such panicking with statically linked executables

Other parts (e.g. linkObjects) do seemingly have this check (?)

Also, I couldn't test this on Linux


avdgrinten:
This fix is correct but `__dlapi_resolve()` can also call directly into `resolveInObject()` on different code paths. Also, `dladdr()` hits the same assert AFAICT.

PR #1888, opened 2026-09-04
sysdeps/linux: Fix stack alignment in _start function(s) for ia32

ThatMishakov:
Align the argmunet area/stack to 16 bytes, as required by the ABI in (S)crt1.S. thread_entry.S seems ok as is (?)


sysdeps/linux/x86/crt-src/crt1.S
```diff
@@ -10,6 +10,10 @@ _start:
 .cfi_undefined eip
 	xor %ebp, %ebp
 	mov %esp, %ecx
+
+	/* Align the stack to 16-byte boundary per i386 psABI */
+	lea -8(%esp), %esp
```
no92: Does the kernel guarantee that the stack is 16-byte aligned on entry here? AFAICT there is no such guarantee, and we should use e.g. `andl $~15, %esp` logic instead.

options/rtld/x86/entry.S
```diff
@@ -4,9 +4,12 @@ PROC_START(_start)
 .cfi_undefined eip
 	call relocateSelf
```
no92, line 10: Shouldn't we ensure that `relocateSelf` is called with proper alignment too?

sysdeps/linux/x86/crt-src/crt1.S
ThatMishakov: i386 and x86_64 psABIs require 16 byte alignment of %rsp/%esp on program entry (and also aarch64? I'm not sure). This could also make for a good __ensure() as well I guess

ThatMishakov: Yeah, actually an older spec didn't require it https://refspecs.linuxfoundation.org/elf/abi386-4.pdf . But linux seems to align the stack? https://github.com/torvalds/linux/blob/master/fs/binfmt_elf.c#L301

Issue #1825, opened 2026-07-15
Add error case to all exhaustive preprocessor architecture if-elses


no92:
In multiple places we have the pattern of `#if defined(__arch__)` switches where arch-specific code is handled. In order to help future ports, we should review all of these if-elses, determine which ones ought to be exhaustive (e.g. ifunc handling in `handleIfunc`), and add an error case:

```c
#else
#error Unknown architecture
#endif
```

PR #1890, opened 2026-09-07
Error out on unknown architectures in the switches that have to cover every port

swimmingbrain:
A port to a new architecture falls through a few `#if defined(__arch__)` chains silently: `handleIfunc` in the rtld returns nothing, `sys/procfs.h` leaves `ELF_NGREG` undefined, `struct stat` and `blksize_t` in the linux abi are missing, and the linux clone sysdep hands the TCB pointer through unadjusted. A first pass at #1825.

Those chains get an `#else` with an `#error` now, in the wording each file already uses. The clone sysdep gains an explicit x86_64 branch, since that is the one architecture where no adjustment is the right thing. Chains that fall back to a `#warning` or to a generic default (`sys/user.h`, `fcntl.h`, `poll.h`, `execinfo.cpp`) were left alone, that looks deliberate, and the other sysdeps were not touched either, those switches belong to their ports.

Built the linux sysdeps for x86_64 before and after, nothing changes.


marv7000:
LGTM but I would rephrase the commit messages to better represent what they're actually changing, not what the problem was.

swimmingbrain:
Fair point, thanks. The four messages now say what each commit adds rather than what went wrong before, same diff.

Dennisbonke (approved):
LGTM

PR #1895, opened 2026-09-12
sysdeps/ethereal: Implement sigreturn, sigaltstack, fdatasync, flock

sasdallas:
Implements a few new system calls


sysdeps/ethereal/x86_64/signal.S
```diff
@@ -0,0 +1,9 @@
+.section .text
+.global __mlibc_restorer
+__mlibc_restorer:
+	/* RSP = ucontext structure */
+	mov %rsp, %rdi
+	mov $121, %rax /* SYS_SIGRETURN */
```
marv7000: Why not just `#include <sys/syscall_nums.h>` and
```suggestion
	mov $SYS_SIGRETURN, %rax
```

marv7000 (approved):
LGTM, but one nit

sysdeps/ethereal/x86_64/signal.S
sasdallas: yeah hold on

PR #1897, opened 2026-09-13
options/rtld: Fix relocateSelf, avoid running constructors and doing relocations twice, and allow constructors in ld.so

ThatMishakov:
The first two fixes are necessary for ia32 (I guess the other arches didn't have any problems because they use RELA array), and the third one allows constructors in ld.so (it's necessary for my filesystem stuff sysdeps with a microkernel, for example), I've had a discussion about it on discord where the consensus was to allow constructors for everyone (?)

I'm not quite sure if adding those two new variables is the cleanest solution, but the thinking behind it is to also reuse it for eventual static PIE support, which would also need to call relocateSelf in rcrt1.o (if I'm understanding it correctly)


options/rtld/generic/linker.hpp
```diff
@@ -246,6 +246,8 @@ struct SharedObject {
 
 	uint64_t globalRts;
 	bool wasLinked;
+	bool skipRelocation = false;
+	bool skipInit = false;
```
avdgrinten, line 250: Instead of adding two variables, isn't it easier to check if a DSO is `ldso`?

ThatMishakov: I've tried to explain it above, I thought about doing it initially like that, but I thought the second variable would be userful for static PIE

avdgrinten (approved):

no92 (approved):

PR #1900, opened 2026-09-17
ansi: strerror_r gated like musl

xinitrcn1:
See https://github.com/ifduyue/musl/blob/5e9972eaef08ccf55dabe254ac829a30329793d3/include/string.h#L66


no92 (changes requested):
Simply removing the GNU-style variant is an ABI break, and will break some downstream users (e.g. OpenJDK 17).

PR #1902, opened 2026-09-21
abis/linux: define greg_t/gregset_t on all architectures

lzcunt:
LoongArch and m68k already defined these. musl and glibc define them for all architectures. Gated behind `_DEFAULT_SOURCE` to match musl gating. Needed for libunwind.


lzcunt:
Technically on aarch64 glibc defines them under a different name but I doubt anything in the wild uses that

marv7000 (approved):
LGTM

PR #1903, opened 2026-09-21
options/linux: implement memalign

lzcunt:
Needed by pixman's tests


marv7000:
This is already part of #1794. It's been stale for a bit, I am going to fix it as soon as I can.

PR #1906, opened 2026-09-23
abis/linux: provide `MAP_HUGE_*` hugetlb flags

lzcunt:
Required by ripgrep


abis/linux/vm-flags.h
```diff
@@ -30,6 +30,22 @@
 #define MAP_SYNC      0x80000
 #define MAP_FIXED_NOREPLACE 0x100000
 
+#define MAP_HUGE_SHIFT 26
+#define MAP_HUGE_MASK 0x3f
+#define MAP_HUGE_16KB (14 << 26)
```
no92, line 35: Why is `26` hardcoded instead of using `MAP_HUGE_SHIFT 26`?

lzcunt: No reason I can change that

